PR #14, opened 2017-02-20
Nested docs querying benchmark.

### LINKED ISSUE #8: Add a track to test nested / parent child performance
- Index + search
- We should also force a high update rate (to see the cost of updating nested docs)

markharwood:
Uses StackOverflow questions+answers nested docs with just title text, tags, authors and dates for fields.

Closes #8


nested/track.json
```diff
@@ -0,0 +1,58 @@
+{
+	"short-description": "Nested query benchmark using up to 11,203,029 questions from StackOverflow",
+	"description": "This test indexes and queries up to 11,203,0299 questions from StackOverflow (total 3.64 GB json) using 2 client threads and 500 docs per bulk request against Elasticsearch",
+	"data-url": "http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora/nested",
+	"#COMMENT": "DEPRECATED: The meta-properties below will stay for a while (for backwards compatibility). Please specify them on top-level for your own tracks",
+	"meta": {
```
danielmitterdorfer: I think we can remove the `meta` block and the commend above for a new track.

nested/track.json
```diff
@@ -0,0 +1,58 @@
...
+			"clients": 1
+		}, {
+			"operation": "randomized-nested-queries",
+			"clients": 2,
+			"warmup-iterations": 10,
+			"iterations": 5000
+		}
+		]
+	}]
+}
```
danielmitterdorfer: Nit: Missing new line

nested/track.json
```diff
@@ -0,0 +1,58 @@
+{
+	"short-description": "Nested query benchmark using up to 11,203,029 questions from StackOverflow",
+	"description": "This test indexes and queries up to 11,203,0299 questions from StackOverflow (total 3.64 GB json) using 2 client threads and 500 docs per bulk request against Elasticsearch",
```
danielmitterdorfer: it's 4 client threads. Maybe just skip this detail?

nested/track.py
```diff
@@ -0,0 +1,75 @@
...
+            },
+            "index": None,
+            "type": None,
+            "use_request_cache": False
+        }
+        return result
+
+
+def register(registry):
+    registry.register_param_source("nested-query-source", QueryParamSource)
```
danielmitterdorfer: Nit: Missing new line.

nested/track.py
```diff
@@ -0,0 +1,75 @@
...
+                self.dates.append(row[2])
+
+    def partition(self, partition_index, total_partitions):
+        return self
+
+    def size(self):
+        return 1
+
+    # Don't do anything time-consuming here. This is on a hot code path! Please run Rally with --enable-driver-profiling and check
+    # that this function does not show up high in the profiler output.
```
danielmitterdorfer: Leftover?

nested/track.json
```diff
@@ -0,0 +1,52 @@
...
+	"challenges": [{
+		"name": "nested-search-challenge",
+		"description": "Indexes document corpus for an hour using Elasticsearch default settings . After that randomized nested queries are run.",
+		"default": true,
+		"index-settings": {
+			"index.number_of_replicas": 0
+		},
+		"schedule": [{
+			"operation": "index-append",
+			"warmup-time-period": 20,
```
danielmitterdorfer: I would increase the warmup time to 120 or maybe 240 seconds. You see that 20 seconds is too small because of the high variation between minimum and median indexing rate in the summary report.

nested/track.json
```diff
@@ -0,0 +1,52 @@
...
+			"time-period":3600,
+			"clients": 4
+		}, {
+			"operation": "force-merge",
+			"clients": 1
+		}, {
+			"operation": "randomized-nested-queries",
+			"clients": 2,
+			"warmup-iterations": 10,
+			"iterations": 5000
```
danielmitterdorfer: You are (at worst) driving the system into 100% utilization measuring latency at an unstable point. I recommend that you define a fixed throughput (say 50 operations / second). See http://esrally.readthedocs.io/en/latest/faq.html#what-does-latency-and-service-time-mean-and-how-do-they-related-to-the-took-field-that-elasticsearch-returns for details (I also linked to a talk of mine where I explain this topic in more depth although I improved the slides in the mean time, see http://daniel.mitterdorfer.name/talks/2016/Benchmarking%20Elasticsearch.pdf - starting with slide 27)

nested/track.py
```diff
@@ -0,0 +1,73 @@
...
+               "query": {
+                  "bool": {
+                     "must": [
+                        {
+                           "match": {
+                              "tag": "%s" % random.choice(self.tags)
+                           }
+                        },
+                        {
+                           "nested": {
```
jpountz: maybe put it in a `filter` clause rather than `must`? it does not seem to produe useful scores due to the fact that the embeeded query under `nested` only has `filter` clauses.

nested/track.json
```diff
@@ -0,0 +1,52 @@
...
+			"documents": "documents.json.bz2",
+			"document-count": 11203029,
+			"compressed-bytes": 695293381,
+			"uncompressed-bytes": 3637747670
+		}]
+	}],
+	"operations": [{
+		"name": "index-append",
+		"operation-type": "index",
+		"bulk-size": 500
```
jpountz: maybe we should have a larger bulk size since docs are tiny? For instance geonames seems to be using 5k as a mulk size.

nested/track.py
markharwood: I was concerned that a `filter` clause may "thrash the cache". The date ranges I use in the query data vary a lot.

jpountz: Then maybe keep the `must` clause here and use `must` clauses instead of `filter` on the `bool` query below too?

PR #16, opened 2017-02-23
Improve search-time benchmarking with the nested track.

jpountz:
This commit adds benchmarks for nested sorting, nested aggregations as well
as queries that do not use the nested structure. The latter might sound
surprising but when nested mappings are used, Elasticsearch needs to add
filters in order to make sure that queries may only match root documents, so
it is important to verify we make these filters efficient.

It also changes the existing nested query so that it matches more documents
in order to exercise the join logic more. The previous queries were very
selective and returned in a couple milliseconds.


nested/track.json
```diff
@@ -20,17 +20,59 @@
...
 		"name": "randomized-nested-queries",
 		"operation-type": "search",
 		"param-source": "nested-query-source",
 		"cache": false
-	}],
+        }, {
+                "name": "randomized-term-queries",
+                "operation-type": "search",
+                "param-source": "term-query-source",
+                "cache": false
```
danielmitterdorfer: Nit: If you call this `use_request_cache` and pass this value with `self._params["use_request_cache"]` in the custom parameter source, it's better aligned with Rally core.

nested/track.py
```diff
@@ -18,15 +16,59 @@ def __init__(self, indices, params):
...
+                            "mode" :  "max",
+                            "order": "desc",
+                            "nested_path": "answers"
+                        }
+                    }
+                ]
+            },
+            "index": None,
+            "type": None,
+            "use_request_cache": False
```
danielmitterdorfer: Here you can do: `"use_request_cache": self._params["use_request_cache"]`. This makes it better aligned with Rally core.

nested/track.json
```diff
@@ -20,17 +20,59 @@
...
+        }, {
+                "name": "randomized-sorted-term-queries",
+                "operation-type": "search",
+                "param-source": "sorted-term-query-source",
+                "cache": false
```
danielmitterdorfer: same as above w.r.t. to `use_request_cache`.

nested/track.json
```diff
@@ -20,17 +20,59 @@
...
+	}, {
+                "name": "match-all",
+                "operation-type": "search",
+                "body": {
+                        "query": {
+                                "match_all": {}
+                        }
+                },
+                "cache": false
```
danielmitterdorfer: same as above w.r.t. to `use_request_cache`

nested/track.json
```diff
@@ -20,17 +20,59 @@
...
+                                                        "date_histogram": {
+                                                                "field": "answers.date",
+                                                                "interval": "month"
+                                                        }
+                                                }
+                                        }
+                                }
+                        }
+                },
+                "cache": false
```
danielmitterdorfer: same as above w.r.t. to `use_request_cache`

nested/track.py
```diff
@@ -18,15 +16,59 @@ def __init__(self, indices, params):
...
+            "body": {
+                "query": {
+                    "match": {
+                        "tag": "%s" % random.choice(self.tags)
+                    }
+                }
+            },
+            "index": None,
+            "type": None,
+            "use_request_cache": False
```
danielmitterdorfer: same as above w.r.t. to `use_request_cache`

nested/track.json
jpountz: maybe we should update `docs/adding_tracks.rst`?

jpountz:
@danielmitterdorfer I pushed a new commit, is it what you had in mind?

PR #17, opened 2017-04-07
Fixed a small typo

kategray:


danielmitterdorfer:
Good catch and thanks for the fix @kategray! I'll merge it soon.

PR #20, opened 2017-04-20
Add `terms` queries to the `geonames` track.

jpountz:
This adds `terms` queries with 30k terms used in various configurations to the
`geonames` dataset so that we are more likely to know about regressions in the
future.

### DIFF
--- a/geonames/track.json
+++ b/geonames/track.json
@@ -318,21 +318,21 @@
           "clients": 1,
           "warmup-iterations": 500,
           "iterations": 1000,
-          "target-throughput": 50
+          "target-throughput": 3
         },
         {
           "operation": "large_filtered_terms",
           "clients": 1,
           "warmup-iterations": 500,
           "iterations": 1000,
-          "target-throughput": 50
+          "target-throughput": 3
         },
         {
           "operation": "large_prohibited_terms",
           "clients": 1,
           "warmup-iterations": 500,
           "iterations": 1000,
-          "target-throughput": 50
+          "target-throughput": 3
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


geonames/track.json
```diff
@@ -297,6 +312,27 @@
...
+          "warmup-iterations": 500,
+          "iterations": 1000,
+          "target-throughput": 50
+        },
+        {
+          "operation": "large_prohibited_terms",
+          "clients": 1,
+          "warmup-iterations": 500,
+          "iterations": 1000,
+          "target-throughput": 50
```
danielmitterdorfer: Based on the relation of latency and service_time I've seen in my tests I'd suggest to lower that setting to 3 ops/s in order to not over-saturate the system on our current hardware.

geonames/track.json
```diff
@@ -297,6 +312,27 @@
...
+        },
+        {
+          "operation": "large_filtered_terms",
+          "clients": 1,
+          "warmup-iterations": 500,
+          "iterations": 1000,
+          "target-throughput": 50
```
danielmitterdorfer: Based on the relation of latency and service_time I've seen in my tests I'd suggest to lower that setting to 3 ops/s in order to not over-saturate the system on our current hardware.

geonames/track.json
```diff
@@ -297,6 +312,27 @@
           "warmup-iterations": 500,
           "iterations": 1000,
           "target-throughput": 20
+        },
+        {
+          "operation": "large_terms",
+          "clients": 1,
+          "warmup-iterations": 500,
+          "iterations": 1000,
+          "target-throughput": 50
```
danielmitterdorfer: Based on the relation of latency and service_time I've seen in my tests I'd suggest to lower that setting to 3 ops/s in order to not over-saturate the system on our current hardware.

PR #22, opened 2017-04-21
Adds challenges to benchmark index sorting

### LINKED ISSUE #21: Add challenges for index sorting
Index sorting is a new feature in elasticsearch 6.x. It might be nice to benchmark some of the default tracks to measure the cost in terms of indexing throughput. 
We're only interested in indexing throughput at the moment since we don't have any search/aggs feature that takes advantage of the index sorting.
The default tracks are maybe not the best fit for index sorting in terms of what we can achieve with it but they offer a variety of use cases that would allow us to measure the cost of this feature with different types of data.

jimczi:
This change adds some challenges to the existing tracks to measure the cost of index sorting.
The goal of these new challenges is just to provide benchmarks for index sorting in various cases.
The challenges measure only indexing throughput and merges since we don't have search/aggs features that take advantage of the index sorting yet.
These challenges are also only compatible with elasticsearch master because index sorting is a new feature in 6.x.

Closes #21

### DIFF
--- a/geonames/track.json
+++ b/geonames/track.json
@@ -359,7 +359,7 @@
       ]
     },
     {
-      "name": "append-fast-sorted no-conflicts",
+      "name": "append-fast-sorted-no-conflicts",
       "description": "Indexes the whole document corpus in an index sorted by country_code field in ascending order and using a setup that will lead to a larger indexing throughput than the default settings. Document ids are unique so all index operations are append only.",
       "index-settings": {
         "index.number_of_replicas": 0,
--- a/geonames/track.json
+++ b/geonames/track.json
@@ -359,6 +359,29 @@
       ]
     },
     {
+      "name": "append-fast-sorted no-conflicts",
+      "description": "Indexes the whole document corpus in an index sorted by country_code field in ascending order and using a setup that will lead to a larger indexing throughput than the default settings. Document ids are unique so all index operations are append only.",
+      "index-settings": {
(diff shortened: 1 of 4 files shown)


geonames/track.json
```diff
@@ -359,6 +359,29 @@
       ]
     },
     {
+      "name": "append-fast-sorted no-conflicts",
```
danielmitterdorfer: I think this challenge should be named "append-fast-sorted-no-conflicts" (i.e. without any spaces). This makes it easier to type the name on the command line (no quoting required).

jpountz:
@danielmitterdorfer This makes me wonder why the logging track is not available at https://benchmarks.elastic.co/?

jimczi:
Thanks @jpountz @danielmitterdorfer 

> This makes me wonder why the logging track is not available at https://benchmarks.elastic.co/?

https://github.com/elastic/night-rally/issues/25
Though no datapoint and it seems that some part are not live yet. I can access
https://elasticsearch-benchmarks.elastic.co/logging/index.html
but the logging track link is missing from the other pages.

jpountz:
Thanks @jimczi for the pointers!

danielmitterdorfer:
@jpountz I added the logging track a few days ago. My usual approach for new tracks is to gather some data first and only then add them to the menu structure. However, I think we need to tweak the challenges that we run.

PR #24, opened 2017-04-26
Add inner hits to nested track

martijnvg:
Something to compare against to see if optimizations that will be added via elastic/elasticsearch#23917 actually help.

### DIFF
--- a/nested/track.json
+++ b/nested/track.json
@@ -34,6 +34,22 @@
       "use_request_cache": false
     },
     {
+      "name": "randomized-nested-queries-with-inner-hits_default",
+      "operation-type": "search",
+      "param-source": "nested-query-source-with-inner-hits",
+      "use_request_cache": false,
+      "size" : 10,
+      "inner_hits_size" : 3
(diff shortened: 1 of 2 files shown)


martijnvg:
>  I'll take care of that in #25.

Ok, thanks @danielmitterdorfer!

I noticed that there are no version labels, so just to be clear, I will also cherry-pick this to the `5` branch.

danielmitterdorfer:
> I noticed that there are no version labels, so just to be clear, I will also cherry-pick this to the 5 branch.

We have four branches:

* `master`: used by Rally for the ES master branch
* `5`: used by Rally for the ES 5.x
* `2`: used by Rally for the ES 2.x
* `1`: used by Rally for the ES 1.x

I usually push a change to master, then cherry-pick it to 5, then cherry-pick that to 2 and finally cherry-pick to 1 (provided the change makes sense for all versions which applies to most of them). I think for your change this should be straightforward but if you are unsure I can help with that as well.

PR #30, opened 2017-07-03
Add NOAA benchmark

martijnvg:
This now benchmarks range fields specifically, but it can also be used to benchmark other numeric query/agg operations.

### DIFF
--- a/noaa/README.txt
+++ b/noaa/README.txt
@@ -1,0 +1,201 @@
+Dataset containing daily weather measurement from NOAA:
+ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/by_year/
+
+Instructions on how to recreate the json documents:
+1) Download the following files:
+* ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/by_year/2016.csv.gz
+* ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/ghcnd-stations.txt
+* ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/ghcnd-countries.txt
+* ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/ghcnd-states.txt
+2) Execute: gunzip 2016.csv.gz
+3) Sort the file by station: sort --field-separator=',' --key=1,2 -o 2016-sorted.csv 2016.csv
+4) Execute a script like is specified at the end of the file to create json doccuments.
+5) Compress the documents json file: 
+   bzip2 -9 -c documents.json > documents.json.bz2
+
+
+
+Python script that process the csv file into Elasticsearch json documents:
+
+----------------------------------------------------------------------------------------------------
+
+import os
+import csv
+import json
+from datetime import datetime
+
+stationsFile = 'ghcnd-stations.txt'
+countriesFile = 'ghcnd-countries.txt'
+statesFile = 'ghcnd-states.txt'
+
+weatherDataFile = '2016-sorted.csv'
+indexPrefix = 'weather-data'
+docType = 'summary'
+
+def loadStatesFile(statesFile):
+    statesMap = {}
+    with open(statesFile, 'r') as file:
+        csvreader = csv.reader(file, delimiter=' ', quotechar='"')
+        for row in csvreader:
+            statesMap[row[0].strip()] = row[1].strip()
+    return statesMap
+
+def loadCountriesFile(countriesFile):
+    countriesMap = {}
+    with open(countriesFile, 'r') as file:
+        csvreader = csv.reader(file, delimiter=' ', quotechar='"')
+        for row in csvreader:
+            countriesMap[row[0].strip()] = row[1].strip()
+    return countriesMap
+
+def loadStationsFile(stationsFile, statesFile, countriesFile):
+    statesMap = loadStatesFile(statesFile)
+    countriesMap = loadCountriesFile(countriesFile)
+    stationsMap = {}
+    with open(stationsFile, 'r') as file:
+        for row in file:
+            try:
+                station = {}
+                station['id'] = row[0:11].strip()
+                countryCode = row[0:2].strip()
+                if len(countryCode) > 0:
+                    station['country_code'] = countryCode
+                    station['country'] = countriesMap[countryCode]
+                station['location'] = {
+                    'lat': float(row[12:20].strip()),
+                    'lon': float(row[21:30].strip())
+                }
+                station['elevation'] = float(row[31:37].strip())
+                if countryCode == 'US':
+                    stateCode = row[38:40].strip()
+                    if len(stateCode) > 0:
+                        station['state_code'] = stateCode
+                        station['state'] = statesMap[stateCode]
+                station['name'] = row[41:71].strip()
+                gsn_flag = row[72:75].strip()
+                if len(gsn_flag) > 0:
+                    station['gsn_flag'] = gsn_flag
+                hcn_crn_flag = row[76:78].strip()
+                if len(hcn_crn_flag) > 0:
+                    station['hcn_crn_flag'] = hcn_crn_flag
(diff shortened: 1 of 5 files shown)


noaa/challenges/default.json
```diff
@@ -0,0 +1,61 @@
+    {
+      "name": "append-no-conflicts",
+      "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green and we want to ensure that we don't use the query cache. Document ids are unique so all index operations are append only. After that a couple of queries are run.",
+      "default": true,
+      "index-settings": {
+        "index.number_of_shards": 1,
```
jpountz: :+1:

noaa/operations/default.json
```diff
@@ -0,0 +1,112 @@
...
+                    "ASN00001018",
+                    "ASN00001019",
+                    "ARM00087121",
+                    "ASN00004032",
+                    "ASN00004046",
+                    "ASN00004019",
+                    "ASN00003105",
+                    "ASN00003100",
+                    "ASN00004083"
+                  ]
```
jpountz: can we have a simple `term` query like the disjunction has? Otherwise if there is a change in performance in that query, it might not be obvious whether it is related to the terms query or to the range?

noaa/track.json
```diff
@@ -0,0 +1,28 @@
+{% import "rally.helpers" as rally with context %}
+
+{
+  "short-description": "Daily weather measurement summaries from around the globe.",
+  "description": "Indexes 10M+ weather measurement summaries from NOAA.",
```
jpountz: maybe document where the data was retrieved?

noaa/operations/default.json
jpountz: if we want to benchmark both the point and the doc values query, it might also help to have one conjunction with a range that matches most documents and a term query that matches between 0.1 and 1%% of the index, and another conjunction where the range matches 2x fewer documents than the range.

noaa/track.json
martijnvg: I added a link in the README, I think that is sufficient?

noaa/operations/default.json
martijnvg: > can we have a simple term query like the disjunction has?

I think you missed the `range_query_range_field_in_conjunction_with_term_query` query above this one?

> it might also help to have one conjunction with a range that matches most document
 and a term query that matches between 0.1 and 1%% of the index,

A weather station in this data set has at most 366 document which is 0,014% of the total amount of documents. So I think the 0.1 and 1% case is covered.

What query could be used for matching most of the docs, that on its own doesn't have a lot of overhead that could interfere with the benchmark? A term range? match_all ?

martijnvg: > A weather station in this data set has at most 366 document which is 0,014% of the total amount of documents. So I think the 0.1 and 1% case is covered.

Arg, I made a mistake. A simple term query for weather station is 0,003%. The `terms` query matches with 5856 documents and that is 0,05%. So what I'll do is increase the number terms in the the `terms` query to get at least to 0,1%

noaa/track.json
jpountz: oh right, I missed it!

noaa/operations/default.json
jpountz: I'm a bit worried that the overhead of merging postings of multiple terms will add noise. Maybe we could cross this dataset with stations (ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/ghcnd-stations.txt) in order to be able to index more metadata with all documents such as geo coordinates, state and elevation of the station. Then I believe we could find some states that have significant numbers of records?

noaa/track.json
jpountz: totally

noaa/operations/default.json
martijnvg: :+1: I'll add more metadata to the documents.

colings86: ++ @martijnvg I can update my python script to do this if you want?

martijnvg: @colings86 Thanks that would be great. Note that for creating this track I did made some modifications to your script, mainly around the fact that it needs to be converted to a json file. This is what I have now: https://gist.github.com/martijnvg/72a3711cb26fd84f196e9a1c4a41d038

noaa/README.txt
```diff
@@ -0,0 +1,5 @@
+Dataset containing daily weather measurement from NOAA:
+ftp://ftp.ncdc.noaa.gov/pub/data/ghcn/daily/by_year/
+
+The dataset has been processed by: https://gist.github.com/colings86/078e85a1131324471f4f10c73570d678
```
danielmitterdorfer: Can you just compress the zip file from the gist and dump it here so it is self-contained? Also, the gist contains instructions, especially: 

>Sort files using something like sort --field-separator=',' --key=1,2 -o ~/Downloads/2017-sorted.csv ~/Downloads/2017.csv

And I think you should document how you sorted the files.

noaa/challenges/default.json
```diff
@@ -0,0 +1,82 @@
...
+        "index.number_of_replicas": 0,
+        "index.queries.cache.enabled": false
+      },
+      "schedule": [
+        {
+          "operation": "index",
+          "#COMMENT": "This is an incredibly short warmup time period but it is necessary to get also measurement samples. As this benchmark is rather about search than indexing this is ok.",
+          "warmup-time-period": 10,
```
danielmitterdorfer: Is this short warmup time period warranted here? I think this is only necessary for percolator (where indexing throughput is not interesting anyway). Ideally we'd have at least 240 seconds here.

noaa/challenges/default.json
```diff
@@ -0,0 +1,82 @@
...
+      "name": "append-no-conflicts-index-only",
+      "description": "Indexes the whole document corpus using Elasticsearch default settings.",
+      "index-settings": {
+        "index.number_of_replicas": 0
+      },
+      "schedule": [
+        {
+          "operation": "index",
+          "#COMMENT": "This is an incredibly short warmup time period but it is necessary to get also measurement samples. As this benchmark is rather about search than indexing this is ok.",
+          "warmup-time-period": 10,
```
danielmitterdorfer: Same as above for the warmup time period. If possible this should be at least 240 seconds.

noaa/track.json
danielmitterdorfer: The comment says 10M+ weather measurements but it's actually only 2.5M.

noaa/challenges/default.json
```diff
@@ -0,0 +1,82 @@
...
+          "clients": 8
+        }
+      ]
+    }
```
danielmitterdorfer: Nit: Missing new line

noaa/operations/default.json
```diff
@@ -0,0 +1,203 @@
...
+                    "gte": -10,
+                    "lte": 10
+                  }
+                }
+              }
+            ]
+          }
+        }
+      }
+    }
```
danielmitterdorfer: Nit: Missing new line

noaa/track.json
martijnvg: @danielmitterdorfer The doc count is actually 10914068, so I'll just update it to that. I would expect Rally to fail with an error, because the `document-count` in `track.json` was incorrect.

martijnvg:
I've updated the PR.

noaa/track.json
danielmitterdorfer: Oh Rally does not count the documents again but I may add this feature I've just raised https://github.com/elastic/rally/issues/296.

jpountz:
@martijnvg could you map the station code as a keyword so that it does not get the text/keyword dual mapping? In general, I think it'd be better to map all fields explicitly and disable dynamic mappings.

martijnvg:
@jpountz yes: 659d6970a9e0b94babec4b7d40e9cc1dbccde04c

PR #41, opened 2018-04-05
Switch to explicit index management

### LINKED ISSUE #37: Tracks should only wait for green for the indices they care about
When running tracks like `nested`, it waits for the entire cluster to be green, rather than waiting for only the `sonested` index to be green. The track operations should ignore the other (potentially non-allocated) indices on the node.

danielmitterdorfer:
With this commit we switch all tracks to explicit index management
(instead of index auto-management). As this change is breaking, we also
increase the track version from 1 to 2 thus requiring at least Rally
0.9.0 to use these tracks.

This allows us to remove also some legacy code that has registered
the `refresh` operation for older versions of Rally (newer versions 
already include this operation in Rally core).

Relates elastic/rally#380
Closes #37


percolator/challenges/default.json
```diff
@@ -99,33 +118,4 @@
           "target-throughput": 50
         }
       ]
-    },
-    {
-      "name": "append-no-conflicts-index-only",
```
dliappis: We don't need this challenge any more? I would have assumed we'd just need to replace the operations here with a delete / create / index, but what am I missing?

danielmitterdorfer: Percolator really is about querying, indexing is just a means to an end here and I figured we can get rid of it right away. If somebody really wants this (which they shouldn't), then they can just use task filtering to get it back.

pmc/track.json
```diff
@@ -1,25 +1,13 @@
...
-        {
-          "name": "articles",
-          "mapping": "mappings.json",
-          "documents": "documents.json.bz2",
-          "document-count": 574199,
-          "compressed-bytes": 5928712141,
-          "uncompressed-bytes": 23256051757
-        }
-      ]
+      "types": ["articles"]
```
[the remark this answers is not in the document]
danielmitterdorfer: That was the necessary evil of backwards-compatibility. :)

PR #44, opened 2018-05-13
Set number of shards explicitly for all tracks

danielmitterdorfer:
With this commit we set the number of shards explicitly to the default
of five shards in order to keep results consistent.

Relates elastic/elasticsearch#30539


danielmitterdorfer:
Thanks for the review! I'll create a separate PR for consistent handling of such index settings in other tracks as well.

PR #50, opened 2018-09-17
Add ML-related challenge

danielmitterdorfer:
With this commit we add a new challenge `append-ml` to the `nyc_taxis`
track that can be used to benchmark the machine learning module in
Elasticsearch. The challenge requires that Rally is invoked
appropriately so it will activate ML in Elasticsearch.

### DIFF
--- a/nyc_taxis/track.py
+++ b/nyc_taxis/track.py
@@ -1,0 +1,13 @@
+import time
+
+
+def wait_for_ml_lookback(es, params):
+    while True:
+        response = es.xpack.ml.get_datafeed_stats(datafeed_id=params["datafeed-id"])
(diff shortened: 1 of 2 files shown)


nyc_taxis/track.py
```diff
@@ -0,0 +1,13 @@
+import time
+
+
+def wait_for_ml_lookback(es, params):
+    while True:
```
dliappis: Should we allow waiting for ever, if something goes wrong, or allow a timeout parameter as an escape hatch?

danielmitterdorfer: I'd not do it at the moment because we usually have timeouts on other levels anyway.

PR #51, opened 2018-11-09
Add update challenge to nyc_taxis

danielmitterdorfer:
With this commit we add a new challenge `update` to the `nyc_taxis` track. We
also include a new track parameter `store_type` which is intentionally
undocumented because we intend to remove it again after an initial experimation
phase.

### DIFF
--- a/nyc_taxis/README.md
+++ b/nyc_taxis/README.md
@@ -60,6 +60,9 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using 
 * `bulk_size` (default: 10000)
 * `bulk_indexing_clients` (default: 8): Number of clients that issue bulk indexing requests.
 * `ingest_percentage` (default: 100): A number between 0 and 100 that defines how much of the document corpus should be ingested.
+* `conflict_probability` (default: 25): A number between 0 and 100 that defines the probability of id conflicts. This requires to run the respective challenge.
+* `on_conflict` (default: "index"): Whether to use an "index" or an "update" action when simulating an id conflict.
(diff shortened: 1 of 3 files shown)


nyc_taxis/README.md
```diff
@@ -60,8 +60,11 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using
 * `bulk_size` (default: 10000)
 * `bulk_indexing_clients` (default: 8): Number of clients that issue bulk indexing requests.
 * `ingest_percentage` (default: 100): A number between 0 and 100 that defines how much of the document corpus should be ingested.
+* `conflict_probability` (default: 25): A number between 0 and 100 that defines the probability of id conflicts. This requires to run the respective challenge.
```
dliappis: How about instead of "This requires to run the respective challenge." to simply and explicitly say "Only used by the `update` challenge".

nyc_taxis/README.md
```diff
@@ -60,8 +60,11 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using
...
+* `on_conflict` (default: "index"): Whether to use an "index" or an "update" action when simulating an id conflict.
```
dliappis: Same comment as ^^

nyc_taxis/README.md
```diff
@@ -60,8 +60,11 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using
...
+* `recency` (default: 0): A number between 0 and 1 that defines whether to bias towards more recent ids when simulating conflicts. See the [Rally docs](http://esrally.readthedocs.io/en/latest/track.html#bulk) for the full definition of this parameter. This requires to run the respective challenge.
```
dliappis: Same comment as ^^

PR #52, opened 2018-11-26
Add update challenge for http_logs track

dliappis:


http_logs/challenges/default.json
```diff
@@ -220,4 +220,52 @@
...
+        {
+          "operation": "delete-index"
+        },
+        {
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {%- if index_settings is defined %} {{index_settings | tojson}} {%- else %} {
+              "index.number_of_shards": {{number_of_shards | default(1)}},
+              "index.number_of_replicas": {{number_of_replicas | default(0)}},
+              "index.store.type": "{{store_type | default('mmapfs')}}"
```
danielmitterdorfer: Should we remove this for the time being? I don't think we need this parameter for every track.

dliappis: As discussed, I'll leave this here as we removed it when we also remove it from nyc_taxis.

PR #54, opened 2018-12-06
Fix benchmarks for scripting queries

mayya-sharipova:
Fix benchmarks for expression, painless-static, painless-dynamic
so that they always return positive scores, as returning negative
scores is forbidden from 7.0

### DIFF
--- a/geonames/operations/default.json
+++ b/geonames/operations/default.json
@@ -113,7 +113,7 @@
               {
                 "script_score": {
                   "script": {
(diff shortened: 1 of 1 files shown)


ebadyano:
@danielmitterdorfer @dliappis from a few runs i did it look like the changes shouldn't affect the scores. I will repeat the run tomorrow to confirm that the difference I see is within usual fluctuations.

ebadyano:
You can merge it anytime. Thank you!

Issue #19, opened 2017-04-20
sparse fields track


LucaWintergerst:
We should consider adding a track with very sparse fields, comparing how doc_values behave over time. This is particularly interesting once we move to lucene 7.

<img width="1150" alt="screen shot 2017-04-20 at 1 08 12 pm" src="https://cloud.githubusercontent.com/assets/11661400/25225582/8d01f23c-25ca-11e7-9e42-8b03bcaf77d9.png">

The run from14:30-16:30 with `doc_values: false`, 
The one from 16:30-17:30 with `doc_values: true`, 
The  very last run is with doc values disabled for all fields

The data has around 2200 fields in total, split across 30 types. The one type with the most documents has 200-300 fields.
The decrease in performance comparing the first two runs is significant, with around 30-40%.
Furthermore the indexing rate also keeps slowing down, the more data gets indexed, which does not happen as much when doc_values are disabled (see run 1)

This test was run on the following hardware:
14 cores
4 SSDs, multiple data paths
30GB heap size

The cluster was CPU bound during all runs.

The merges can't keep up in the second run and "indexing throttled" messages were showing up in the logs

LucaWintergerst:
after re-running the test with with just one type and 200 fields, the performance did not change significantly. Almost no change was visible in monitoring. Unfortunately I don't have exact numbers.

jpountz:
@LucaWintergerst FYI Elasticsearch master is on Lucene 7 since Tuesday.

The Nested track has sparse fields by design due to the use of `nested` fields: fields that exist in the parent do not exist in children and vice-versa. The `geonames` dataset also has the `elevation` field which is only present in 26% of documents.

I'd be fine with adding a track that has more sparse fields, but only if it is realistic. Our recommendation still is and will always be to model documents in a way that fields are dense whenever possible.

LucaWintergerst:
I do understand and fully support our stance against sparse fields and therefore our recommendation for dense documents, but from what we see our customers do this does not always apply. 
Often times the source data only contains a subset of the fields which are possible to appear in a document. While it is certainly possible to model the data or indices to counteract sparsity, it creates and additional overhead that a user might not be willing to pay without understanding why he should even care. 

It can also be hard to defend this stance without having credible data in the form of benchmarks to convince the user otherwise.
How bad is sparsity really? How much does it impact indexing, searching, index size and so on?
I'm sure that you (or we) can answer these questions but I would still like to have data that I can show people to convince them otherwise. 
[...]

PR #56, opened 2018-12-12
Add a track for Metricbeat data

pcsanwald:
This PR creates a track for metricbeat data, which I'd like to use to benchmark the autohisto aggregation. I've marked this "WIP (work in progress)" because I need to generate a realistic dataset, but, I'd like any feedback on what I've done so far also, as this is the first rally track I've created.

I hope that metricbeat data will be broadly useful, and not just useful for my use case, per the discussion in #19. My idea for generating a dataset is to have one with two metricbeat modules running, that collect beats at different intervals, perhaps an elasticsearch module running hourly, and a system module running on the minute. We'd then run the autohisto aggregation on a field that's only present in elasticsearch (`node_stats` maybe), across a time range and bucket number combination that would result in minute level buckets.


metricbeat/challenges/default.json
```diff
@@ -0,0 +1,179 @@
...
+            "index": "metricbeat",
+            "request-params": {
+              "wait_for_status": "{{cluster_health | default('green')}}",
+              "wait_for_no_relocating_shards": "true"
+            }
+          }
+        },
+        {
+          "operation": "index-append",
+          "warmup-time-period": 0,
```
danielmitterdorfer: I guess bulk-indexing the document corpus is considered more of a setup task here (i.e. you're not interested in bulk-indexing throughput)? In that case setting the warmup time-period to zero is fine.

metricbeat/challenges/default.json
```diff
@@ -0,0 +1,179 @@
...
+          "operation": "force-merge",
+          "clients": 1
+        },
+        {
+          "name": "refresh-after-force-merge",
+          "operation": "refresh",
+          "clients": 1
+        },
+        {
+          "operation": "index-stats",
```
danielmitterdorfer: I don't think we need to benchmark index stats here?

metricbeat/challenges/default.json
```diff
@@ -0,0 +1,179 @@
...
+          "clients": 1,
+          "warmup-iterations": 0,
+          "iterations": 1000,
+          "target-throughput": 100
+        },
+        {
+          "operation": "node-stats",
```
danielmitterdorfer: I don't think we need to benchmark node stats here?

metricbeat/challenges/default.json
```diff
@@ -0,0 +1,179 @@
...
+          "operation": "default",
+          "clients": 1,
+          "warmup-iterations": 0,
+          "iterations": 1000,
+          "target-throughput": 50
+        }
+      ]
+    },
+    {
+      "name": "append-no-conflicts-index-only",
```
danielmitterdorfer: I would not include any of the other challenges *unless* you are interested in those metrics.

metricbeat/challenges/default.json
```diff
@@ -0,0 +1,179 @@
...
+        },
+        {
+          "operation": "node-stats",
+          "clients": 1,
+          "warmup-iterations": 0,
+          "iterations": 1000,
+          "target-throughput": 100
+        },
+        {
+          "operation": "default",
```
danielmitterdorfer: Is it intentional that the aggregation queries are not referenced here?

metricbeat/files.txt
```diff
@@ -0,0 +1 @@
+documents.json
```
danielmitterdorfer: This file is needed for the script that prepares the corpus for offline usage (see https://github.com/elastic/rally-tracks/blob/master/download.sh). I think when you're done this should only contain the compressed corpus file (+ a smaller file that contains the test corpus, see our [docs](https://esrally.readthedocs.io/en/stable/adding_tracks.html#adding-support-for-test-mode)).

metricbeat/index.json
```diff
@@ -0,0 +1,9216 @@
...
+  "settings": {
+    "index.number_of_shards": {{number_of_shards | default(1)}},
+    "index.number_of_replicas": {{number_of_replicas | default(0)}},
+    "index.mapping.total_fields.limit": 20000,
+    "index.search.slowlog.level" : "debug",
+    "index.search.slowlog.threshold.fetch.debug" : "0s",
+    "index.search.slowlog.threshold.query.debug" : "0s"
+  },
+  "mappings": {
+    "doc": {
```
danielmitterdorfer: Can we use `_doc` here for consistency with the other tracks?

metricbeat/track.json
```diff
@@ -0,0 +1,32 @@
+{% import "rally.helpers" as rally with context %}
+{
+  "version": 2,
+  "description": "Metricbeat data",
+  "indices": [
+    {
+      "name": "metricbeat",
+      "body": "index.json",
+      "types": ["doc"]
```
danielmitterdorfer: Can we use `_doc` here for consistency with the other tracks?

metricbeat/track.json
```diff
@@ -0,0 +1,32 @@
...
+    }
+  ],
+  "corpora": [
+    {
+      "name": "metricbeat",
+      "base-url": "http://0.0.0.0:8000",
```
danielmitterdorfer: This property is optional so you can also remove it while you're iterating.

pcsanwald:
@danielmitterdorfer I think I've addressed most of your comments, this is ready for another review I think.

Note that I have gotten an actual slug of metricbeat data from `infra-stats`, so, when we are ready to merge this, I can upload that to S3 and update the config to account for the zipped version, etc.

metricbeat/files.txt
danielmitterdorfer: I think this should contain:

```
documents.json.bz2
documents-1k.json.bz2
```

For details please see my comment above.

metricbeat/operations/default.json
```diff
@@ -0,0 +1,65 @@
+    {
+      "name": "index-append",
+      "operation-type": "bulk",
+      "bulk-size": {{bulk_size | default(10000)}},
+      "ingest-percentage": {{ingest_percentage | default(100)}}
+    },
+    {
+      "name": "default",
```
danielmitterdorfer: Are you interested in measuring performance of `match_all`? If not I suggest to remove it from the track.

metricbeat/index.json
```diff
@@ -0,0 +1,10962 @@
+{
+  "settings": {
+    "index.number_of_shards": {{number_of_shards | default(1)}},
+    "index.number_of_replicas": {{number_of_replicas | default(0)}},
+    "index.mapping.total_fields.limit": 20000,
+    "index.search.slowlog.level" : "debug",
+    "index.search.slowlog.threshold.fetch.debug" : "0s",
+    "index.search.slowlog.threshold.query.debug" : "0s"
+  },
+  "mappings": {
```
danielmitterdorfer: Meanwhile types have been removed on master so this will issue deprecation warnings. I have extracted the Metricbeat mapping from 7.0.0-beta1 in https://gist.github.com/danielmitterdorfer/72686146100dea7b775f0efdc80bd2f1. Can you please use that mapping instead so we avoid the deprecation warnings?

metricbeat/track.json
```diff
@@ -0,0 +1,31 @@
+{% import "rally.helpers" as rally with context %}
+{
+  "version": 2,
+  "description": "Metricbeat data",
+  "indices": [
+    {
+      "name": "metricbeat",
+      "body": "index.json",
+      "types": ["_doc"]
```
danielmitterdorfer: As types have been removed on master you can just remove this property (the most recent version of Rally will allow you to do this).

pcsanwald:
certainly no urgency on this, but I think I've addressed review comments and this should be ready for another review.

pcsanwald:
I've pinged daniel and we are uploading the track data

metricbeat/track.json
```diff
@@ -0,0 +1,30 @@
...
+  "description": "Metricbeat data",
+  "indices": [
+    {
+      "name": "metricbeat",
+      "body": "index.json"
+    }
+  ],
+  "corpora": [
+    {
+      "name": "metricbeat",
```
danielmitterdorfer: can you please add the following property:

```json
"base-url": "http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora/metricbeat"
```

metricbeat/track.json
```diff
@@ -0,0 +1,30 @@
...
+      "documents": [
+        {
+          "source-file": "documents.json",
```
danielmitterdorfer: This should be `documents.json.bz2` so it is using the compressed version.

danielmitterdorfer:
The track data are now uploaded to S3 and I tested it locally. I left two more comments (see above). Can you please address them @pcsanwald? Then I think we're good to merge.

metricbeat/track.json
pcsanwald: I pushed b9326

pcsanwald:
@danielmitterdorfer thanks for the review! I believe I have addressed both your comments.

PR #61, opened 2019-01-02
Adds a geoshapes track

### LINKED ISSUE #60: Add a track for geo_shape
We should add a track for `geo_shape`.

@iverase @nknize @imotov Who wants to own this?

imotov:
Adds a track from OSM-derived geoshape data.

Closes #60


geoshape/files.txt
```diff
@@ -0,0 +1,2 @@
+documents.json.bz2
+documents-1k.json.bz2
```
jpountz: is the latter still needed?

jpountz:
Oh, nevermind I just saw that you added Daniel as a reviewer already!

jpountz:
@imotov I'm curious whether you already ran this benchmark with both legacy quad-tree based geoshapes and the new bkd-backed geo shapes?

imotov:
@jpountz I tried to run it on 6.5.4 with default settings but it died after going through 3% of the data set. I suspect it probably ran out of memory but I don't know for sure.

The result for master can be found [here](https://gist.github.com/imotov/12e36f31c654151cd46fa965c0e1dcf9). There are some failures that still need to be investigated. I am pretty sure the indexing issues are related to https://github.com/elastic/elasticsearch/issues/26286 but search issues are somewhat puzzling. I am planning to look into it next.

jpountz:
+1 to skip update benchmarks. Updates make some sense due to the fact that they trigger increased merging activity and that multidimensional points are slower than other Lucene data structures at merging when the number of dimensions is greater than 1. That said I could easily live without it.

geoshape/README.md
```diff
@@ -0,0 +1,31 @@
+## Geopoint track
+
+This track is based on [PlanetOSM](http://wiki.openstreetmap.org/wiki/Planet.osm) data.
+
+### Example Document
+
+```json
+{
```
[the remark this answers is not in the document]
imotov: Sure. The script basically takes the file that @nknize provided in https://github.com/elastic/rally-tracks/issues/60#issuecomment-449483186 and converts it into a single element JSON, so we are unlikely to run it ever again. But I will add it just in case.

geoshape/index.json
```diff
@@ -0,0 +1,19 @@
+{
+  "settings": {
+    "index.number_of_shards": {{number_of_shards | default(5)}},
```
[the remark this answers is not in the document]
imotov: Makes sense.

imotov:
I pushed some changes.

> However, indexing failed immediately with...

I spent some time analyzing the failures. It looks like we have 3 types of failures there. As I mentioned before, some of them seems to be related to  elastic/elasticsearch#26286. However, I also found some instances of elastic/elasticsearch#36883 and a few instances of `Invalid polygon, interior cannot share more than one point with the exterior`. For the last set of errors some cases seem to be pretty [obvious](https://gist.github.com/imotov/f73f9cd68a63713c922e5dd16de26689) but in others [not as much]( https://gist.github.com/imotov/993c4c8d99f7c2de4cf73844288f043e). I am still spot checking the failures.

> I also checked your gist quickly and from the large variation in indexing throughput it seems to me the system is not properly warmed up.

The test was running for several hours and I noticed significant slowdown towards the end of the test. So, I think it might be something else.

danielmitterdorfer:
I reran the benchmark once more on the same environment and let it continue when errors occur. This is the error rate (on a per document basis, not per bulk request) over time:

per-doc-error-rate

I also created a chart for the indexing throughput over time:

indexing-throughput

In both charts, the warmup is the grey part and the actual measurement phase is the green part. Leaving the errors aside, I think it's problematic that the benchmark never reaches a steady state. While at the beginning of the measurement phase we reach roughly 45500 docs/s, throughput declines over time and reaches 10900 docs/s at the end of the benchmark. I think it is dangerous to report a median throughput in that case because it suggests that a higher throughput can be reached over a longer period of time when in fact it is heavily dependent on how long we execute the benchmark. One thing I am wondering about is *why* we never reach steady state. Maybe @nknize has more thoughts on this?

jpountz:
> why we never reach steady state

I suspect it's related to the fact that merging becomes increasingly costly as the index size grows with multi-dimensional points since the whole BKD tree needs to be rebuilt from scratch. Geo points and range fields have the issue as well, but geo shapes make it worse by using 4 indexed dimensions when ranges and geo points only use 2.

geoshape/challenges/default.json
```diff
@@ -0,0 +1,60 @@
...
+          "name": "refresh-after-force-merge",
+          "operation": "refresh",
+          "clients": 1
+        },
+        {
+          "operation": "polygon",
+          "clients": 1,
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
```
danielmitterdorfer: The empirical data that I have gathered on hardware that is similar to our nightly environment suggests that this throughput is too high. With service times around 2.5 seconds I suggest a target throughput of 0.3 (meaning that a query will be scheduled roughly every 3 seconds)

geoshape/challenges/default.json
```diff
@@ -0,0 +1,60 @@
...
+        },
+        {
+          "operation": "bbox",
+          "clients": 1,
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
```
danielmitterdorfer: The empirical data that I have gathered on hardware that is similar to our nightly environment suggests that this throughput is too high. With service times around 3 to 3.5 seconds I suggest a target throughput of 0.25 (meaning that a query will be scheduled every 4 seconds)

danielmitterdorfer:
That sounds reasonable. If I understand you correctly this means we will never reach steady state though. I wonder how we can make the benchmark meaningful w.r.t. to throughput metrics.

FWIW the complete benchmark has finished by now and I did not see any errors for any of the queries (in Igor's benchmark results we see a positive error rate for `polygon` queries).

imotov:
> That sounds reasonable. If I understand you correctly this means we will never reach steady state though. I wonder how we can make the benchmark meaningful w.r.t. to throughput metrics.

Maybe we can create a new index every 5mln records or so? It is strange though, I was kind of expecting that limit on the maximum segment size should take care of that. 

> in Igor's benchmark results we see a positive error rate for `polygon` queries

Yeah, my machine was unresponsive at the end of the indexing run. I am not sure if it was thermal throttling or something else.

jpountz:
> I was kind of expecting that limit on the maximum segment size should take care of that.

This is a good point, the index is much larger than 5GB in the end so merging activity should reach a steady state at some point.

iverase:
Not sure if relevant but I would not expect a constant indexing throughput because of the nature of the data. It is not the same to index a document with a point (one triangle) than a document with a polygon with thousand of points (~thousand of triangles + tessellator overhead). 

I think the batch size might be too big as some of the documents can be quite big (polygons with thousand of points). In my experience I had to limit the batch size to 2500.

We should expect 8751 documents to fail.

imotov:
> Not sure if relevant but I would not expect a constant indexing throughput because of the nature of the data. 

Indeed, the test file starts with a whole bunch of LINESTRINGS followed by MULTILINES and then POLYGONS. However, I thought we randomize the records that we index. If I am mistaken, then this might be indeed an issue.

imotov:
As per our discussion, I am going to split this PR into 3 separate tracks for LINESTRINGS, MULTILINES and POLYGONS and will remove all shapes that elasticsearch cannot process at the moment.

danielmitterdorfer:
> As per our discussion, I am going to split this PR into 3 separate tracks for LINESTRINGS, MULTILINES and POLYGONS and will remove all shapes that elasticsearch cannot process at the moment.

Thanks for iterating @imotov. Just a suggestion: You can keep everything in one track but you could create three different [corpora](https://esrally.readthedocs.io/en/stable/track.html#corpora) and three challenges (one for each corpus). That makes it a bit easier to maintain because you can keep everything in the same track. I'm happy to assist if you have more specific questions.

imotov:
I ran the new test without any errors over night and got interesting results. The first graph is the average throughput for the first two hours of the run and the second is the average for the entire run:
screen shot 2019-01-12 at 7 31 23 am
screen shot 2019-01-12 at 7 30 35 am

danielmitterdorfer:
Thanks for rerunning it. Could this be explained by https://github.com/elastic/rally-tracks/pull/61#issuecomment-451421760?

> merging becomes increasingly costly as the index size grows with multi-dimensional points since the whole BKD tree needs to be rebuilt from scratch.

Btw, I think it is great that we have isolated the three different types now and can see how they behave.

imotov:
>  Could this be explained by #61 (comment)?

I am not sure. The index is much bigger than a maximum segment size. I don't understand the mechanism that prevents it from stabilizing. We shouldn't modify the older segments, so I am not sure what exactly kicks in that slows down the overall performance. The change is quite drastic though especially at the beginning where we index linestrings for 1 hour and the performance steadily drops from 24,000 to 8,000. That's 3x difference. At the end of the test our throughput is 2,000, which is 10x slower than when we started.

> Btw, I think it is great that we have isolated the three different types now and can see how they behave.

@danielmitterdorfer I wonder if it would make sense to also index them in 3 different indices instead of a single index as I do at the moment. Any other ideas on how to make this test more meaningful? My main concern is that if settle on using this overall test running time as  "a canary in a coal mine" whatever is causing the slowdown on the large index will overshadow whatever smaller regressions and improvements we will introduce until we figure out and fix the source of this slowdown.

danielmitterdorfer:
>> Could this be explained by #61 (comment)?

> I am not sure. 

You could try rerunning with a profiler (`--telemetry=jfr`to capture a flight recording) or maybe we need to add run a benchmark with more verbose logs enabled so we can see what is happening?

> I wonder if it would make sense to also index them in 3 different indices instead of a single index as I do at the moment. 

That would make sense to me, yes.

danielmitterdorfer:
@imotov is this PR superseded by #62 or do you intent to still work on this one?

imotov:
> s this PR superseded by #62 or do you intent to still work on this one?

@danielmitterdorfer I would say #62 is a parallel effort. Knowing how geoshapes work for points is very useful for the effort to consolidate geo_point and geo_shape data types but it is not sufficient as a  geo_shape metric. We still need a full-blown geo_shape test with different geo_shape types.

I think I addressed your comments from this PR as much as I could. If you see any additional issues I can address, please let me know. The slowdown through the test is still present, but I think it's more a reflection of reality rather than a flaw in the test. Hopefully, @iverase's merge optimization will reduce this issue.

danielmitterdorfer:
Thanks for the update @imotov! I'll do a final pass very soon.

geoshape/README.md
```diff
@@ -0,0 +1,28 @@
...
+  "shape": "LINESTRING(-1.8212114 52.5538901, -1.8205573 52.554324)"
+}
+```
+
+### Parameters
+
+This track allows to overwrite the following parameters with Rally 0.8.0+ using `--track-params`:
+
+* `bulk_size` (default: 500)
```
danielmitterdorfer: This differs from the value in the track (100) + also I am not sure whether the same bulk size for all corpora is right. I wonder whether we need to introduce dedicated parameters per corpus?

geoshape/index.json
```diff
@@ -0,0 +1,19 @@
+{
+  "settings": {
+    "index.number_of_shards": {{number_of_shards | default(1)}},
+    "index.number_of_replicas": {{number_of_replicas | default(0)}}
+  },
+  "mappings": {
+    "_doc": {
```
danielmitterdorfer: This works for `6.x` but not on master anymore as types have been removed now.

geoshape/operations/default.json
```diff
@@ -0,0 +1,62 @@
+    {
+      "name": "index-append-linestrings",
+      "operation-type": "bulk",
+      "bulk-size": {{bulk_size | default(100)}},
```
danielmitterdorfer: As I've mentioned in my feedback for the README, I'm not sure whether we should have a different bulk size per corpus?

geoshape/track.json
```diff
@@ -0,0 +1,73 @@
+{% import "rally.helpers" as rally with context %}
+
+{
+  "version": 2,
+  "description": "Shapes from PlanetOSM",
+  "indices": [
+    {
+      "name": "osmlinestrings",
+      "body": "index.json",
+      "types": ["_doc"]
```
danielmitterdorfer: Support for types has been removed on master, so this works against 6.x but not against master.

geoshape/track.json
```diff
@@ -0,0 +1,73 @@
...
+    }
+  ],
+  "corpora": [
+    {
+      "name": "linestrings",
+      "base-url": "http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora/geoshape",
+      "target-index": "osmlinestrings",
+      "target-type": "_doc",
```
danielmitterdorfer: Support for types has been removed on master, so this works against 6.x but not against master.

PR #64, opened 2019-02-06
Allow _doc type in master

dliappis:
https://github.com/elastic/elasticsearch/pull/38270 made current master
(7.0.0) reject mappings with type `_doc` unless
`?include_type_name=true` is passed during index creation.

Workaround the rejection for now by explicitly setting
`include_type_name` to `true` in every create_index operation.


dliappis:
Note that https://github.com/elastic/rally/pull/642 is a prerequisite to this.

PR #65, opened 2019-02-12
Adds a track that indices geopoints as geoshapes (backport to 6)

imotov:
This track contains the same data as the geopoint
track and performs the same operations (when they
are supported) except it indexes points as
geoshapes.

Relates #60


imotov:
I started backporting this to 5.x when I realized that test documents are currently formatted as WKT, and support for this format was added only in 6.x. So, I would have to reformat data as geojson in order to add support for 5.x and below. Does it make sense to do that?

danielmitterdorfer:
> Does it make sense to do that?

IMHO it does not make sense so let's keep this as is. Thank you so much for the effort you put into this @imotov! We really appreciate this a lot.

PR #66, opened 2019-02-14
Remove document types

danielmitterdorfer:
With this commit we remove document types from all Rally tracks as they
have been removed with Elasticsearch 7.0.0.

Relates elastic/rally#647


danielmitterdorfer:
Note that this PR requires elastic/rally#647 and thus we will only merge it after Rally 1.0.4 has been released in order not to break users (they still need to upgrade Rally so this change will work).

PR #67, opened 2019-03-04
Add pipeline support for Eventdata track

alexsapran:
This change introduces the option to allow the eventdata track to be ingested
with the use of the pipeline plugins geo-ip and user-agent.

To start using pippeline with this track you should use the new challenge
`append-no-conflicts-with-ingest-pipeline` and set the appropriate track param
`ingest_pipeline_name` to use of the baseline, geoip, useragent, combined
pipelines.


alexsapran:
Closing in favor of https://github.com/elastic/rally-eventdata-track/pull/31

PR #69, opened 2019-03-19
Adapt geopointshape track to track conventions

### LINKED ISSUE #68: esrally run benchmark always download geopointshape track data  to geopointshapes folder.
esrally run benchmark always download geopointshape track data  to geopointshapes folder even  run ./dowload geopointshape  to download track data to local storage.

```json
    {
      "name": "geopointshapes",
      "base-url": "http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora/geopointshape",
      "documents": [
        {
          "source-file": "documents.json.bz2",
          "document-count": 60844404,
          "compressed-bytes": 493367095,
          "uncompressed-bytes": 2780550484
        }
      ]
    }
```

danielmitterdorfer:
With this commit we align the name of the corpus in the track
`geopointshape` with the track's name. We also define a unique index
name across all our tracks (previously the index name in that track was
identical with the index name in the `geopoint` track which could lead
to problematic situations).

Closes #68

### DIFF
--- a/geopointshape/challenges/default.json
+++ b/geopointshape/challenges/default.json
@@ -16,7 +16,7 @@
           "name": "check-cluster-health",
           "operation": {
             "operation-type": "cluster-health",
-            "index": "osmgeopoints",
+            "index": "osmgeoshapes",
             "request-params": {
               "wait_for_status": "{{cluster_health | default('green')}}",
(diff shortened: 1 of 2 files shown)


danielmitterdorfer:
Thanks for the review! :)

PR #71, opened 2019-04-12
Allow to override number of shards for updates

danielmitterdorfer:
With this commit we allow to override the number of shards even for
the challenge `append-fast-with-conflicts` in `geonames`.

### DIFF
--- a/geonames/challenges/default.json
+++ b/geonames/challenges/default.json
@@ -270,7 +270,7 @@
             "operation-type": "create-index",
             "settings": {
               "index.refresh_interval": "30s",
-              "index.number_of_shards": 6,
+              "index.number_of_shards": {{number_of_shards | default(6)}},
               "index.translog.flush_threshold_size": "4g"
             }
           }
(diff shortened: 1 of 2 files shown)


geonames/challenges/default.json
```diff
@@ -270,7 +270,7 @@
             "operation-type": "create-index",
             "settings": {
               "index.refresh_interval": "30s",
-              "index.number_of_shards": 6,
+              "index.number_of_shards": {{number_of_shards | default(6)}},
```
dliappis: I think we should also allow overriding at least `number_of_replicas` as well.

dliappis: Also there can still be an issue if someone tries to override everything using `index_settings`, but I guess we don't want to deal with this in this PR.

danielmitterdorfer: No, overriding `number_of_replicas` actually works because this definition is merged with the original one. I've pushed a commit that mentions that overriding `index_settings` is unsupported for this challenge.

dliappis: Ah thanks!

danielmitterdorfer:
Thanks!

PR #77, opened 2019-05-29
Add sort operations to geonames and http_logs

mayya-sharipova:


http_logs/challenges/default.json
```diff
@@ -77,6 +77,20 @@
           "iterations": 200,
           "#COMMENT": "Throughput is considered per request. So we issue one scroll request per second which will retrieve 25 pages",
           "target-throughput": 1
+        },
+        {
+          "operation": "desc_sort_timestamp",
+          "clients": 1,
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 1.5
```
danielmitterdorfer: Can we please lower the target throughput to 1 operation per second? I did a test on our nightly environment and it could not reach that throughput.

http_logs/challenges/default.json
```diff
@@ -77,6 +77,20 @@
...
+        },
+        {
+          "operation": "asc_sort_timestamp",
+          "clients": 1,
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 1.5
```
danielmitterdorfer: Can we please lower the target throughput to 1 operation per second? I did a test on our nightly environment and it could not reach that throughput.

http_logs/challenges/default.json
mayya-sharipova: @danielmitterdorfer Thanks Daniel, this has been addressed in the last commit. I wonder if we even need to decrease the throughput further, say to `0.5`.  I have the following results from running benchmarks on my laptop:
```
 |   Min Throughput |           desc_sort_timestamp |      0.65 |  ops/s |
 |Median Throughput |           desc_sort_timestamp |      0.66 |  ops/s |
 |   Max Throughput |           desc_sort_timestamp |      0.66 |  ops/s |
```
```
 |   Min Throughput |            asc_sort_timestamp |      0.76 |  ops/s |
 |Median Throughput |            asc_sort_timestamp |      0.76 |  ops/s |
 |   Max Throughput |            asc_sort_timestamp |      0.76 |  ops/s |
```

danielmitterdorfer: You raise a great point. Unfortunately, the target throughput is an absolute number that is the same regardless on which system the benchmark is running. For our latency-oriented benchmarks we aim for a high utilization (roughly 70%) on our nightly hardware (see this [blog post](http://robharrop.github.io/maths/performance/2016/02/20/service-latency-and-utilisation.html) for the rationale). Consequently, this means that on other systems this throughput might be still too high or too low. We have an idea how to allow that Rally is  adaptive to the system on which the benchmark is being run but for the time being I think a target throughput of 1 ops/s makes most sense for our nightlies. I hope that makes sense?

mayya-sharipova:
@danielmitterdorfer  It would be useful for us to have these benchmarks for latest 7.x as well. Does it correspond to correspond to rally's branch 7?

danielmitterdorfer:
>  It would be useful for us to have these benchmarks for latest 7.x as well. Does it correspond to correspond to rally's branch 7?

Yes, 7.x corresponds to the branch `7` in rally-tracks. I've backported your changes to the branches `7`, `7.1` and `7.0` so these queries will now work for all versions starting with Elasticsearch 7.0.0.

Issue #79, opened 2019-06-19
Use the default store type instead of assuming one


danielmitterdorfer:
We currently allow to pass the [index store type](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-store.html) as a track parameter for some tracks and default to `hybridfs` if nothing is passed. However, this is making some assumptions, e.g. that memory mapping is allowed (which can be turned off explicitly with `node.store.allow_mmap`). There is a better default though: We can just default to the correct default value which is `fs`.

PR #80, opened 2019-06-27
Change default store_type to fs

ebadyano:
Closes https://github.com/elastic/rally-tracks/issues/79

### DIFF
--- a/geonames/index.json
+++ b/geonames/index.json
@@ -2,7 +2,7 @@
   "settings": {
     "index.number_of_shards": {{number_of_shards | default(5)}},
     "index.number_of_replicas": {{number_of_replicas | default(0)}},
-    "index.store.type": "{{store_type | default('hybridfs')}}",
+    "index.store.type": "{{store_type | default('fs')}}",
     "index.requests.cache.enable": false
   },
   "mappings": {
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -247,7 +247,7 @@
             "settings": {%- if index_settings is defined %} {{index_settings | tojson}} {%- else %} {
               "index.number_of_shards": {{number_of_shards | default(1)}},
               "index.number_of_replicas": {{number_of_replicas | default(0)}},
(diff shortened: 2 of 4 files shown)


ebadyano:
@dliappis One concern I have is that Daniel very recently delivered a chance in https://github.com/elastic/rally-tracks/commit/f1e41437e1ef29b29f34a532699f82ff4346954a where he used `hybridfs` as a default and not `fs`; wonder if I'm missing something for #79. Also I ran nested with the default changes to `fs`, seems to work fine. Any ideas?

dliappis:
> where he used `hybridfs` as a default and not `fs`

As per https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-store.html#file-system:

> fs
> Default file system implementation. This will pick the best implementation depending on the operating environment, which is currently hybridfs on all supported systems but is subject to change.

So currently (as of ES 7.0, please do double check the versions in the docs) `fs` equals `hybridfs` so we are ok.

I'd be careful about backporting this PR though to older branches esp. since `fs` became available after 7.0.

danielmitterdorfer:
> I'd be careful about backporting this PR though to older branches esp. since fs became available after 7.0.

@ebadyano, @dliappis I think it is safe to backport this all the way until `5` as `fs` is [available there as well](https://www.elastic.co/guide/en/elasticsearch/reference/5.0/index-modules-store.html#file-system)? However, he `store_type` track parameter is currently available in `7`, `7.2`, `7.1` and `7.0` and I think we should backport this PR at least to these branches as well.

dliappis:
Thanks @danielmitterdorfer 👍 on backporting to `7`, `7.2`, `7.1` and `7.0`.

ebadyano:
Backported to 7, 7.1, 7.0.  I don't see 7.2 branch under rally-tracks..

danielmitterdorfer:
> I don't see 7.2 branch under rally-tracks..

You're right. Thanks for backporting it!

PR #81, opened 2019-07-05
Allow configure `conflicts` for updates

dnhatn:
With this commit, we allow overriding the conflict type for update challenge of some tracks. One use case is to pass `sequential` for `conflicts` and `0` for `conflict_probability` to disable Elasticsearch's auto-generated id for append-only indices.


dnhatn:
@danielmitterdorfer Thanks for looking. I pushed changes. Can you please take another look?

dnhatn:
Thanks @danielmitterdorfer :)

PR #83, opened 2019-09-04
Adjust to new parameter source API

danielmitterdorfer:
With this commit we change the existing parameter sources to conform to
a new API for parameter sources in Rally. We also keep the old methods
for backwards compatibility with older releases of Rally.

Relates elastic/rally#763


danielmitterdorfer:
This is motivated by elastic/rally#763 and should be reviewed together but can be merged independently.

danielmitterdorfer:
Thanks for the reviews!

PR #85, opened 2019-09-25
Replace interval parameter with fixed_interval and calendar_interval

drawlerr:
The `interval` parameter is now deprecated, so we must now use `fixed_interval` or `calendar_interval` in our queries.

### DIFF
--- a/http_logs/operations/default.json
+++ b/http_logs/operations/default.json
@@ -89,7 +89,7 @@
           "by_hour": {
             "date_histogram": {
               "field": "@timestamp",
-              "interval": "hour"
+              "fixed_interval": "hour"
             }
           }
         }
--- a/http_logs/operations/default.json
+++ b/http_logs/operations/default.json
@@ -89,7 +89,7 @@
           "by_hour": {
             "date_histogram": {
               "field": "@timestamp",
-              "fixed_interval": "hour"
+              "fixed_interval": "1h"
             }
           }
         }
--- a/metricbeat/operations/default.json
+++ b/metricbeat/operations/default.json
@@ -46,7 +46,7 @@
           "occurrences_over_time": {
             "date_histogram": {
               "field": "@timestamp",
-              "fixed_interval": "hour"
+              "fixed_interval": "1h"
             }
           }
         }
--- a/metricbeat/operations/default.json
+++ b/metricbeat/operations/default.json
@@ -46,7 +46,7 @@
           "occurrences_over_time": {
             "date_histogram": {
               "field": "@timestamp",
-              "interval": "hour"
+              "fixed_interval": "hour"
             }
           }
         }
--- a/nyc_taxis/operations/default.json
+++ b/nyc_taxis/operations/default.json
@@ -101,7 +101,7 @@
           "dropoffs_over_time": {
             "date_histogram": {
               "field": "dropoff_datetime",
-              "interval": "day"
+              "fixed_interval": "day"
             }
           }
         }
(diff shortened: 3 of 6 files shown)


pmc/operations/default.json
```diff
@@ -54,7 +54,7 @@
           "articles_over_time": {
             "date_histogram": {
               "field": "timestamp",
-              "interval": "month"
+              "fixed_interval": "month"
```
danielmitterdorfer: This fails with a rather cryptic error message:

```json
{
  "error" : {
    "root_cause" : [
      {
        "type" : "x_content_parse_exception",
        "reason" : "[8:33] [date_histogram] failed to parse field [fixed_interval]"
      }
    ],
    "type" : "x_content_parse_exception",
    "reason" : "[8:33] [date_histogram] failed to parse field [fixed_interval]",
    "caused_by" : {
      "type" : "illegal_argument_exception",
      "reason" : "failed to parse [month]",
      "caused_by" : {
        "type" : "number_format_exception",
        "reason" : "For input string: \"mont\""
      }
    }
  },
  "status" : 400
}
```

digging in the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-datehistogram-aggregation.html#_fixed_intervals) reveals the cause:

> [...] fixed intervals cannot express other units such as months, since the duration of a month is not a fixed quantity. Attempting to specify a calendar interval like month or quarter will throw an exception.
>
> If we try to recreate the "month" calendar_interval from earlier, we can approximate that with 30 fixed days

Can you please adjust this accordingly (here and in the other affected places)? I'd expect that this *might* also change latency numbers. As a follow-up to this PR, can you please check our nightlies and add an according annotation should that be the case?

danielmitterdorfer:
Another note for a follow-up: `interval` has been [deprecated with 7.2](https://www.elastic.co/guide/en/elasticsearch/reference/7.2/search-aggregations-bucket-datehistogram-aggregation.html#_calendar_and_fixed_intervals) so we should also backport this also to `7` (no additional PR needed, just cherry-pick, test all affected tracks locally and then directly push to that branch).

pmc/operations/default.json
danielmitterdorfer: Upon reconsideration I'd suggest that we move to `calendar_interval` here for the simple reason that it also broadens the coverage of our benchmarks, i.e. with benchmarks for `fixed_interval` and `calendar_interval` we'd notice regressions over time in both.

drawlerr: 👍

http_logs/operations/default.json
```diff
@@ -89,7 +89,7 @@
           "by_hour": {
             "date_histogram": {
               "field": "@timestamp",
-              "interval": "hour"
```
polyfractal: Note: this was previously interpreted as a calendar interval, but has been changed to fixed.  The old parsing logic treated "singular" values (hour, day, month, etc) as calendar if it was parseable as calendar.

So in this case it should be `calendar_interval: hour`

metricbeat/operations/default.json
```diff
@@ -46,7 +46,7 @@
           "occurrences_over_time": {
             "date_histogram": {
               "field": "@timestamp",
-              "interval": "hour"
```
polyfractal: Same as above

nyc_taxis/operations/default.json
```diff
@@ -101,7 +101,7 @@
           "dropoffs_over_time": {
             "date_histogram": {
               "field": "dropoff_datetime",
-              "interval": "day"
```
polyfractal: same as above

PR #86, opened 2019-10-03
Fix bad intervals

drawlerr:
Bad parameters in previous PR threw off our benchmarks.  Addressed as per comments on #85

### DIFF
--- a/http_logs/operations/default.json
+++ b/http_logs/operations/default.json
@@ -89,7 +89,7 @@
           "by_hour": {
             "date_histogram": {
               "field": "@timestamp",
-              "fixed_interval": "1h"
+              "calendar_interval": "hour"
             }
           }
         }
--- a/metricbeat/operations/default.json
+++ b/metricbeat/operations/default.json
@@ -46,7 +46,7 @@
           "occurrences_over_time": {
             "date_histogram": {
               "field": "@timestamp",
-              "fixed_interval": "1h"
+              "calendar_interval": "hour"
             }
           }
         }
--- a/nyc_taxis/operations/default.json
+++ b/nyc_taxis/operations/default.json
@@ -101,7 +101,7 @@
           "dropoffs_over_time": {
             "date_histogram": {
               "field": "dropoff_datetime",
-              "fixed_interval": "day"
+              "calendar_interval": "day"
             }
           }
         }


polyfractal:
Thanks @drawlerr! 

Should we also disable request caching on the benchmarks while we're at it?  If the query is a match_all, the agg is the same and the segments haven't been updated since the last query, the request cache will just return the last response.  Usually in 3-5ms. :)

I think it's probably been an oversight until this point -- unless there is a reason to we're leaving it on -- but we should probably disable it for our benchmarks since we're trying to analyze the agg itself, not the caching effect.

dliappis:
> I think it's probably been an oversight until this point -- unless there is a reason to we're leaving it on -- but we should probably disable it for our benchmarks since we're trying to analyze the agg itself, not the caching effect.

For all tracks cache is disabled via https://www.elastic.co/guide/en/elasticsearch/reference/current/shard-request-cache.html#_enabling_and_disabling_caching e.g. in [http_logs](https://github.com/elastic/rally-tracks/http_logs) it can be seen [here](https://github.com/elastic/rally-tracks/blob/8dbe6a9176a8cc10d85918cfb594969f58e366ae/http_logs/index.json#L5), so I think we are good regarding request caching?

dliappis:
For the record, I think that the reason the latency went down for `nyc_taxis` so much with https://github.com/elastic/rally-tracks/pull/85 wasn't due to caching (as explained above) but because it was erroring out (and we don't immediately terminate a race when errors happen).
The error received is:

```
	("Request returned an error. Error type: transport, Description: x_content_parse_exception ({'error': {'root_cause': [{'type': 'x_content_parse_exception', 'reason': '[1:177] [date_histogram] failed to parse field [fixed_interval]'}], 'type': 'x_content_parse_exception', 'reason': '[1:177] [date_histogram] failed to parse field [fixed_interval]', 'caused_by': {'type': 'illegal_argument_exception', 'reason': 'failed to parse setting [date_histogram.fixedInterval] with value [day] as a time value: unit is missing or unrecognized'}}, 'status': 400})", None)
```

polyfractal:
Aha, didn't notice the index setting (was only looking for the request parameter).  Thanks for clearing that up @dliappis!

dliappis:
Thank _you_ @polyfractal for being on top of the results and the comments leading to this PR!

polyfractal:
All credit in that department goes to @talevy for spotting the discrepancy :) :heart:

talevy:
correction: @drawlerr discovered it. I only relayed

polyfractal:
Haha, it's a game of telephone :)

dliappis:
@drawlerr Thanks for merging! Things are looking normal again:

image

What is still needed is to annotate the chart so that it's easy to figure out what was the reason for the anomaly in the future.

PR #90, opened 2019-11-13
http_logs add force merge to 1 segment

mayya-sharipova:
Measure the performance of sort operations
after force merging to 1 segment.

PR https://github.com/elastic/elasticsearch/pull/48533  adds a new merge policy that interleaves old and new
segments on force merge. This checks the sort
performance with this policy after docs are merged to
1 segment.

### DIFF
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -96,7 +96,8 @@
           "name": "force-merge-1-seg",
           "operation": {
             "operation-type": "force-merge",
-            "max-num-segments": 1
+            "max-num-segments": 1,
+            "request-timeout": 7200
           },
           "clients": 1
         },
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -91,6 +91,35 @@
           "warmup-iterations": 200,
           "iterations": 100,
           "target-throughput": 0.5
+        },
+        {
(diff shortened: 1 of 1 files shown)


http_logs/challenges/default.json
```diff
@@ -91,6 +91,35 @@
           "warmup-iterations": 200,
           "iterations": 100,
           "target-throughput": 0.5
+        },
+        {
+          "name": "force-merge-1-seg",
+          "operation": {
+            "operation-type": "force-merge",
+            "max-num-segments": 1
```
danielmitterdorfer: Can you please also add `request-timeout`and set it to a value that is high enough (I guess `7200` (2 hours) should suffice) to ensure we'll wait long enough for this operation to finish? Otherwise we might run queries while the force merge is still ongoing in the background.

mayya-sharipova: @danielmitterdorfer Thanks for the feedback. Addressed in 77faa8b

mayya-sharipova:
@danielmitterdorfer  we would like to see these benchmarks starting from ES 7.6.  For this, should I backport this PR to `rally-tracks` branch 7?

danielmitterdorfer:
> For this, should I backport this PR to rally-tracks branch 7?

@mayya-sharipova yes, that would do the trick. :)

PR #92, opened 2019-11-25
Add reindex operation

henningandersen:
Add reindex operation to http_logs track in order to
verify reindex performance before and after resilient
reindex implementation as well as keep an eye on it
for the future.

Relates elastic/elasticsearch#42612

### DIFF
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -121,22 +121,6 @@
           "warmup-iterations": 200,
           "iterations": 100,
           "target-throughput": 0.5
-        },
-        {
-          "name": "reindex",
-          "operation": {
-            "operation-type": "reindex",
-            "body": {
-              "source": {
-                "index": "logs-*"
-              },
-              "dest": {
-                "index": "reindexed-logs"
-              }
-            },
-            "request_timeout": 7200
-          },
-          "clients": 1
         }
       ]
     },
@@ -331,7 +315,7 @@
     },
     {
       "name": "append-no-conflicts-index-reindex-only",
-      "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only. After indexing, same data are reindexed",
+      "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only. After indexing, same data are reindexed.",
       "schedule": [
         {
           "operation": "delete-index"
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -121,6 +121,22 @@
           "warmup-iterations": 200,
           "iterations": 100,
           "target-throughput": 0.5
+        },
+        {
+          "name": "reindex",
+          "operation": {
+            "operation-type": "reindex",
+            "body": {
+              "source": {
+                "index": "logs-*"
+              },
+              "dest": {
+                "index": "reindexed-logs"
+              }
+            },
+            "request_timeout": 7200
+          },
+          "clients": 1
         }
       ]
     },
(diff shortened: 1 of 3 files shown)


http_logs/challenges/default.json
```diff
@@ -312,4 +328,65 @@
...
+          "operation": "force-merge",
+          "clients": 1
+        },
+        {
+          "name": "refresh-after-force-merge",
+          "operation": "refresh",
+          "clients": 1
+        },
+        {
+          "name": "reindex",
```
dliappis: If you don't want to use a custom runner, you use the [raw-request operation](https://esrally.readthedocs.io/en/stable/track.html#raw-request).

It would then look like:

```
{
  "name": "reindex",
  "operation": {
    "operation-type": "raw-request",
    "method": "POST",
    "path": "/_reindex",
    "body": {
      "source": {
        "index": "logs-*"
      },
      "dest": {
        "index": "reindexed-logs"
      }
    },
    "request-params": {
      "request_timeout": 7200
    }
  },
  "clients": 1
}
```

http_logs/challenges/default.json
```diff
@@ -312,4 +328,65 @@
...
+            "operation-type": "reindex",
+            "body": {
+              "source": {
+                "index": "logs-*"
+              },
+              "dest": {
+                "index": "reindexed-logs"
+              }
+            },
+            "request_timeout": 7200
```
dliappis: Did you intend to make this operation blocking too? In other words, do you want to add `wait_for_completion=true` so that there is a record of how long this operation takes?

http_logs/track.py
```diff
@@ -0,0 +1,5 @@
+def reindex(es, params):
```
dliappis: See the earlier comment about the use of [raw-request](https://esrally.readthedocs.io/en/stable/track.html#raw-request) if you'd like to avoid adding a custom runner.

http_logs/challenges/default.json
```diff
@@ -121,6 +121,22 @@
           "warmup-iterations": 200,
           "iterations": 100,
           "target-throughput": 0.5
+        },
+        {
```
dliappis: What is the intention of adding reindex to the default `append-no-conflicts` challenge? Do you want to get some sort of graph displayed in https://elasticsearch-benchmarks.elastic.co/#tracks/http-logs/nightly/default/30d and if yes what metrics?

http_logs/challenges/default.json
dliappis: One FYI thing to be generally aware of is that if there are proxies involved in the path between Rally and Elasticsearch the proxy may timeout the request abruptly (if `wait_for_completion=true`) and then the Elasticsearch Python client will retry.

http_logs/challenges/default.json
henningandersen: The primary purpose of the custom runner is to ensure that the `request_timeout` is passed to the python "driver" since it uses a low timeout by default. This was based on elastic/rally#567 and elastic/rally#669. I think that if I add it into request-params, it will simply be sent to the elasticsearch server as a http request-parameter?

http_logs/challenges/default.json
henningandersen: Reindex defaults to `wait_for_completion=true`, so I think that part is OK?

I am not sure there is anything to do about the proxy issue? Is there maybe a way to instruct the python client to never retry and instead fail (unless we have scenarios where a retry is expected)? That would be a general rally change I think?

http_logs/challenges/default.json
henningandersen: Good question. I think it could be either docs/s or total run-time (docs/s is probably better, but harder to find)? What will I need to do in order to get it to show up?

http_logs/challenges/default.json
dliappis: I see this can be confusing.

For `raw-request`, any `request_params` specified in the operation get [passed as params](https://github.com/elastic/rally/blob/89b9e791380201b6e9c855c78f7420b20b878a9b/esrally/driver/runner.py#L1250) -- by Rally -- to the [es.transport.perform_request](https://elasticsearch-py.readthedocs.io/en/master/connection.html?highlight=transport%20connection#elasticsearch.Transport.perform_request) method. In turn this method when it [detects the request_timeout key](https://github.com/elastic/elasticsearch-py/blob/35afe32a6dc34bf8cdd946a3723b59b2d0b375a4/elasticsearch/transport.py#L343) in the `params` it passes it as [timeout](https://github.com/elastic/elasticsearch-py/blob/35afe32a6dc34bf8cdd946a3723b59b2d0b375a4/elasticsearch/transport.py#L358) in the connection method.

So to summarize either for specific Rally operations that directly support setting `request_timeout` as added in elastic/rally#567 / elastic/rally#669 / your custom runner or defining it as a key in the `request-params` dict of the `raw-request` operation, ultimately it will be converted to a connection `timeout` by the Elasticsearch Python client.

http_logs/challenges/default.json
dliappis: Ah great, I missed that `true` is the default for `reindex`! I don't think there's anything to be done, I just raised it as a FYI in case you want to run it against an external environment and get surprising results.

http_logs/challenges/default.json
dliappis: To get metrics (depending on what the `_reindex` API returns) you should stick with the runner that you've created and then return a tuple or dict as described [here](https://esrally.readthedocs.io/en/stable/adding_tracks.html#custom-runners) -- under `This function can return:` -- where you can define `weight/unit`. You can additionally specify more KV pairs that will be appended in the meta section of the metric records.

henningandersen: I removed the addition to the default challenge and added docs/s metric for reindex in c86dcc2ee620ed2b6e06e45da817aa69bb5f496e

henningandersen:
This should be ready for another round, sorry about the delay on this one.

PR #93, opened 2019-11-29
Set cluster-settings via API

danielmitterdorfer:
With this commit we avoid the soon to be deprecated track property
`cluster-settings` in the `pmc` track and instead implement a fallback
using the update cluster settings API. Newer versions of Rally will
expose this API via a runner but for older versions we need to implement
it in the track.


danielmitterdorfer:
Thanks for the review!

PR #104, opened 2020-03-03
Ensure segments are merged

danielmitterdorfer:
With this commit we change all tracks to increase the request timeout of the
`force-merge` task to ensure the force merge finishes before we move on. We also
include another task that waits until all ongoing segment merges finish. Older
Rally versions will entirely rely on the increased request timeout for force
merges but newer Rally versions (every version including elastic/rally#925) will
evaluate the condition and wait accordingly until the system is quiet.

Relates elastic/rally#925

### DIFF
--- a/eventdata/challenges/default.json
+++ b/eventdata/challenges/default.json
@@ -29,8 +29,10 @@
           "clients": {{bulk_indexing_clients | default(8)}}
         },
         {
-          "operation": "force-merge",
-          "clients": 1
+          "operation": {
+            "operation-type": "force-merge",
+            "request-timeout": 7200
+          }
         }
       ]
     }
--- a/eventdata/challenges/default.json
+++ b/eventdata/challenges/default.json
@@ -33,6 +33,19 @@
             "operation-type": "force-merge",
             "request-timeout": 7200
           }
+        },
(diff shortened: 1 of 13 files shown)


eventdata/challenges/default.json
```diff
@@ -29,8 +29,23 @@
           "clients": {{bulk_indexing_clients | default(8)}}
         },
         {
-          "operation": "force-merge",
-          "clients": 1
```
dliappis: Nit:

Removing `"clients": 1`, even though the right thing as [it is the default](https://esrally.readthedocs.io/en/stable/track.html#schedule) makes things a bit inconsistent as elsewhere (in most of the other tasks) we still have `"clients": 1`.
Does it make sense to retain it and in a follow-up PR remove all the `"clients": 1` in tasks?

danielmitterdorfer: I thought I take the opportunity to remove some unnecessary noise here. As discussed elsewhere I will create a follow-up PR which removes the rest of these lines.

danielmitterdorfer: I've created https://github.com/elastic/rally-tracks/pull/105.

PR #107, opened 2020-03-10
Use keyword field for term query and terms agg

danielmitterdorfer:
With this commit we use the keyword subfield in the `geonames` track for
the `term` query and aggregations that aggregate on terms. While the
terms aggregation produces correct results on a `text` field, the `term`
query does not return any results (which is unintended).


danielmitterdorfer:
Thank you for the review!

PR #111, opened 2020-04-07
Backwards-compatible implementation of ML lookback

danielmitterdorfer:
With this commit we implement the ML lookback for `nyc_taxis` in a
backwards-compatible way. As newer versions of Elasticsearch (and thus
the corresponding client) have dropped the `_xpack` path component we
are falling back to the raw transport API.

Relates elastic/rally#812

### DIFF
--- a/nyc_taxis/track.py
+++ b/nyc_taxis/track.py
@@ -4,7 +4,7 @@ import time
 
 def wait_for_ml_lookback(es, params):
     while True:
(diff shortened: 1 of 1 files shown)


danielmitterdorfer:
Note that this does **not** target the `master` branch but branch `5` instead.

danielmitterdorfer:
Thanks for the review! :)

PR #121, opened 2020-06-02
Add operations for geohash and geotile aggs

nik9000:
Performance tests for elastic/elasticsearch#57483


hub-cap:
Any particular `car` requirements to run these? Ideally would these land in 7 and master?

PR #131, opened 2020-09-10
Observability Track

gingerwizard:
First pass at a new observability track. Pending agreement on usage i will push docs. 

Whilst initially focusing on logging (indexing only), this track aims to be as flexible as possible and should be re-usable across metics, logs and APM with future applicability to security. 

Two primary goals of this track:

1. Support use of representative datasets and control the ratios according to common user usage patterns 
2. Ensure we can produce data with different date profiles (not just limited to the original corpus) matching to the current time period (so queries with `now-x` match) 
3. Keep configuration options simple and in align with knowledge available to the user - for logging `daily_logging_volume_gb` and `number_of_days`.

**The track corpus and current challenges are examples only and are subject to change once we identify appropriate datasets and appropriate parameter defaults**. A few small test datasets are provided for testing in the `./data`. These are also likely to be removed prior to commit. Initially a `logging-indexing` challenge is provided only.

This track aims to mirror Elastic solutions as closely as possible and thus uses datastreams and index templates. These aren't natively supported in Rally and are hence implemented through custom parameter sources and runners here. As datastreams and index template are not a primary configuration option on tracks, they are configured through indices and legacy template configuration respectively i.e. Any indices declared on the track are considered to be datastreams and legacy templates index templates. Ideally, we will remove this abstraction by addressing https://github.com/elastic/rally/issues/1054 and https://github.com/elastic/rally/issues/1053. The custom runners and parameter sources can in turn be removed.

As mentioned above, the user specifies volumes to be indexed via  `daily_logging_volume_gb` and `number_of_days`.

A `process-data` parameter source produces a dataset to satisfy these volumes using the corpus defined in the track. Importantly, this dataset is written to disk (supports many clients for performance) for later consumption by the bulk generator (with its own custom parameter source for consuming these files). All normal behaviour of the bulk operator is supported except conflicts i.e. user has full flexibility as to the number of clients and throughput. 

The generated file is produced according to configured `corpora_ratios`. The timestamps of each document are set according to a profile - for now we just support deterministic but this is easy to expand. The first document has timestamp `end_date - number_of_days`. The `end_date` defaulting to the value `now` but timestamps and offsets are supported.

Finally, we support 2 modes - `raw` and `processed`. I think we need to formalise this at a track level better. Right now it simply uses a different corpus set depending on the value. The intention here is:

1. `raw` (default) - Uses  data which represents the data originating from our beats. This uses the ingest pipelines defined in any index templates and associated with datastreams to be used. This mode is most representative of real world usage. 
2. `processed`  - Sends the data is it would look post ingestion. This mode assumes all ingest pipelines have been removed from templates using the `remove_pipelines` parameter on the `create-all-index-templates` source. Primary motivation for this is so we can do a side by side analysis when inspecting regressions and exclude ingestion as a cause. 

We considered the possibility of using a single source (processed data) with the `process-data`  parameter source being responsible for dropping fields in raw mode and producing the raw data. The challenge here is what fields to drop - it varies per module and any allow list or deny list could be extensive slowing data generation. 

Ingest pipelines and ilm policies are also loaded through custom runners.

Finally, few final thoughts:

1. Bulk size will likely change per data stream in reality - currently configurable for whole track. Can be addressed by making datastreams a first class configuration option.
2. We rely on ILM associated with the datastreams to rollover indices.
3. To improve file generation speed the `process-data` parameter source supports multiple clients. If using > 1 client,  each client is given an equal portion of each source file. An output file is produced per client. We therefore don't respect the ordering of the original data.
7. Readers currently read one line at a time. We may want to consider reading more than 1 line for efficiency.
8. The deterministic profile for timestamp generation is currently allowing milliseconds increments - for logs this might be sufficient but we might want to move to micro or even nano for metrics on larger volumes.
9. Do we need to formalise `raw` and `processed` better e.g. require the user to specify both for a track.

### DIFF
--- a/geonames/challenges/default.json
+++ b/geonames/challenges/default.json
@@ -77,13 +77,13 @@
           "operation": "term",
           "warmup-iterations": 500,
           "iterations": 1000,
-          "target-throughput": 150
+          "target-throughput": 140
         },
         {
           "operation": "phrase",
           "warmup-iterations": 500,
           "iterations": 1000,
-          "target-throughput": 150
+          "target-throughput": 140
         },
         {
           "operation": "country_agg_uncached",


gingerwizard:
@danielmitterdorfer as well

danielmitterdorfer (commented):
Some high level comments on this track:

* I think there is also infrastructure missing to execute the tests (initialization of the virtual environment, invocation of the test command, ...). As discussed offline it makes sense to add a Makefile for this.
* The track can also be greatly simplified if we implement https://github.com/elastic/rally/issues/335. This would allow us to create a dedicated component for file processing and remove some workarounds.

I also have some observations about style (I did not comment on each of them):

* Logger initialization should be moved to specific classes / functions
* Use double-quotes for strings
* Use f-strings if possible
* Add missing new-lines at the end of files

observability/runners/data_processor.py
```diff
@@ -0,0 +1,72 @@
...
+        self._file_source = None
+        self._output_folder = None
+        self._ops = 0
+
+    def write_meta_file(self):
+        with open(os.path.join(self._output_folder, f"{self._client_index}.metadata"), "w") as meta_file:
+            meta_file.write(str(int(self._ops)))
+
+    async def __call__(self, es, params):
+        es.noop()
```
danielmitterdorfer, line 53: One workaround that I'd try so we can avoid this change in Rally, is:

```python
        rc = es.request_context.get()
        rc["request_start"] = 0
        rc["request_end"] = 0
```

As this is only temporary anyway (can be removed once https://github.com/elastic/rally/issues/335 is implemented) I would try to keep workarounds as contained as possible.

PR #132, opened 2020-09-29
Changes for runtime fields

nik9000:
Adds an option to load the `unparsed` data in the `http_logs` track and
create all fields as `runtime` fields that reference the raw `message`
field.


http_logs/index.json
```diff
@@ -1,6 +1,6 @@
 {
   "settings": {
-    "index.number_of_shards": {{number_of_shards | default(5)}},
+    "index.number_of_shards": {{number_of_shards | default(1)}},
```
nik9000, line 3: This seemed useful for me when I was testing the amount of space that fields take up. I'm happy to remove it from this PR if we'd like.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
     "properties": {
       "@timestamp": {
         "format": "strict_date_optional_time||epoch_second",
-        "type": "date"
+        "type": "date",
+        "doc_values": {{ timestamp_doc_values | default(true) | tojson }},
+        "index": {{ timestamp_index | default(true) | tojson }}
```
nik9000, line 17: I added these and similar statement to be able to test the size of the on disk index after indexing. If we'd like to separate these out into a separate PR I'd be happy with it.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
...
-        "index": false,
-        "doc_values": false
-      },
-      "request": {
-        "type": "text",
-        "fields": {
-          "raw": {
-            "ignore_above": 256,
-            "type": "keyword"
+      {%- if runtime_grok is defined %}
```
nik9000, line 19: This implements using my `grok` prototype. I'm 10000% sure we aren't going to keep this API, but it is something.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
...
-      },
-      "size": {
-        "type": "integer"
-      },
-      "geoip" : {
-        "properties" : {
-           "country_name": { "type": "keyword" },
-           "city_name": { "type": "keyword" },
-           "location" : { "type" : "geo_point" }
+      {%- elif runtime_script_grok is defined %}
```
nik9000, line 48: This implements using the a `script` to do grok-like-stuff. This is implemented in the 7.x branch but we're actively debating the API again.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
...
+          "doc_values": true
+        },
+        "clientip": {
+          "type": "runtime",
+          "runtime_type": "ip",
+          "script": "String m = doc[\"message\"].value; int end = m.indexOf(\" \"); emit(m.substring(0, end));"
+        },
+        "request": {
+          "type": "runtime",
+          "runtime_type": "keyword", {# This should be text with a subtype of keyword #}
```
nik9000, line 61: This is interesting! Runtime fields don't actually implement text at all so this isn't really an apples to apples comparison. I'd like to know what to do with this change.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
...
+          "runtime_type": "long",
+          "script": "String m = doc[\"message\"].value; int end = m.lastIndexOf(\" \"); int start = m.lastIndexOf(\" \", end - 1) + 1; emit(Long.parseLong(m.substring(start, end)));"
+        },
+        "size": {
+          "type": "runtime",
+          "runtime_type": "long",
+          "script": "String m = doc[\"message\"].value; int start = m.lastIndexOf(\" \") + 1; emit(Long.parseLong(m.substring(start)));"
+        }
+      {%- else %}
+        "message": {
```
nik9000, line 75: And this is how things used to be.

http_logs/index.json
```diff
@@ -12,38 +12,118 @@
...
+          "type": "keyword",
+          "index": false,
+          "doc_values": false
+        },
+        "clientip": {
+          "type": "ip",
+          "doc_values": {{ doc_values | default(true) | tojson }},
+          "index": {{ index | default(true) | tojson }}
```
nik9000, line 83: Except that I've added all of these so you can turn off indexing or doc values if you need to get a size estimate.

http_logs/operations/default.json
```diff
@@ -79,6 +87,58 @@
         }
       }
     },
+    {
+      "name": "200s-in-range",
```
nik9000, line 91: These typify the kinds of queries that we hope are useful with runtime fields - ones where you can use the `@timestamp` to eliminate many options and then let runtime fields slowly check everything.

noaa/index.json
```diff
@@ -13,258 +13,426 @@
     },
     "properties": {
       "AWDR": {
-        "type": "keyword"
+        "type": "keyword",
+        "doc_values": {{ doc_values | default(true) | tojson }},
+        "index": {{ index | default(true) | tojson }}
```
nik9000, line 18: This is the same space checking change but for noaa which is useful because it is has lots and lots and lots of fields.

nik9000:
I've opened this more as a place to share my work then because I expect we'll want to merge it. I think we'll likely want to pick parts of this that we'd like to merge, but I'm not clear *which* parts and when.

http_logs/index.json
nik9000: I think we don't actually *use* the `text` field in our tests so this is "ok".

http_logs/challenges/default.json
```diff
@@ -23,11 +23,22 @@
             }
           }
         },
-        {
-          "operation": "index-append",
-          "warmup-time-period": 240,
-          "clients": {{bulk_indexing_clients | default(8)}}
-        },
+        {%- if runtime_script_grok is defined %}
```
dliappis, line 26: @nik9000 In order to visualize the queries in https://elasticsearch-benchmarks.elastic.co/#tracks/http-logs/nightly/default/30d we will actually need a second challenge. The query visualizations and the chart generator will not current visualize more than one operations in the same chart (plus it'd get very noisy given that they already visualize 4 percentiles).

So I suggest we create a new challenge e.g. `append-no-conflicts-with-runtime-fields` or something like that, where we have all the content of the `append-no-conflicts` challenge plus the operations in the conditionals here. We can skip the `if` in the new challenge.

http_logs/index.json
dliappis: Yeah this will be a surprise to users relying on defaults, I think we should try to keep this default value.

http_logs/index.json
```diff
@@ -9,19 +9,52 @@
     "_source": {
       "enabled": {{ source_enabled | default(true) | tojson }}
     },
+    {%- if runtime_script_grok is defined %}
```
dliappis, line 12: Similarly I think we should limit the amount of files that contain conditionals. I suggest we have a separate `index-runtimefields.json` and apply it conditional in the central `track.json` where we anyway have some conditionals.

dliappis (changes requested):
I left some comments, mainly regarding necessary changes to allow us to create the necessary visualizations.

http_logs/index.json
nik9000: ++

http_logs/index.json
nik9000: I've removed these.

http_logs/challenges/default.json
nik9000: I have this condition that I call copy-and-paste blindness. If I see a bunch of stuff that looks copy and pasted I can never find the small bits that differ. Is there any way I can write this as a function and call it twice to get two named challenges? That way the if statement stay? Otherwise I'll never be able to find the difference.

http_logs/index.json
nik9000: I hadn't realized that was a thing. I'm happy to do that.

http_logs/challenges/default.json
dliappis: I hear you. You can leverage the `{{ rally.collect(parts="") }}` helper to reference common sections.

Example approach:

1. Create a `common` dir under challenges, e.g. like:

```
├── challenges
│   ├── common
│   │   └── default-schedule.json
│   └── default.json
```

2. Move the the schedule from the `append-no-conflicts` challenge to `common/default-schedule.json`.
3. Change the `append-no-conflicts` challenge to be:

    ```
      {
        "name": "append-no-conflicts",
        "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only. After that a couple of queries are run.",
        "default": true,
        "schedule": [
          {{ rally.collect(parts="common/default-schedule.json") }}
        ]
      },
    ```

4. Create a second challenge with a different name + description, same content, that relies on use specifying the `runtime_script_grok` track parameter.

This is just an idea, one could split sections from the schedule of the default challenge to more files and include as required.

dliappis (changes requested):
Thanks for iterating and changing things based on comments!

The last thing we need to still, to allow us to create separate charts, is to create a new challenge for the runtime fields. I've left a [comment](https://github.com/elastic/rally-tracks/pull/132/files#r525301630) suggesting an approach to reduce copy/pasta lemme know if you need assistance with it.

nik9000:
@dliappis I've pushed the dry-stuff. I feel bad that the two challenges are the same and you only get a different thing with a track param. We could totally set variables before importing the template to get *some* of the changes, but the corpa change is the big one that can't do that.

dliappis:
> @dliappis I've pushed the dry-stuff. I feel bad that the two challenges are the same and you only get a different thing with a track param. We could totally set variables before importing the template to get _some_ of the changes, but the corpa change is the big one that can't do that.

@nik9000 I must be doing something wrong, I don't see any additional commits (or signs of a force push) since I reviewed yesterday and don't see two separate challenges. Is there something I am missing?

nik9000:
> @nik9000 I must be doing something wrong, I don't see any additional commits (or signs of a force push) since I reviewed yesterday and don't see two separate challenges. Is there something I am missing?

I suppose it'd help if I'd actually pushed..... I've pushed now though!

http_logs/challenges/default.json
```diff
@@ -3,166 +3,14 @@
...
-          "warmup-iterations": 10,
-          "iterations": 100,
-          "target-throughput": 0.5
-        }
+        {{ rally.collect(parts="common/default-schedule.json") }}
+      ]
+    },
+    {
+      "name": "append-runtime-script-grok",
+      "description": "Indexes the whole document corpus using scripts to extract fields.",
```
dliappis, line 11: Maybe we can add in the description that this relies on setting the track param `runtime_script_grok` to `true`.

http_logs/challenges/common/default-schedule.json
```diff
@@ -0,0 +1,183 @@
...
+  "iterations": 100,
+  "target-throughput": 0.5
+},
+{
+  "name": "asc-sort-with-after-timestamp-after-force-merge-1-seg",
+  "operation": "asc_sort_with_after_timestamp",
+  "warmup-iterations": 10,
+  "iterations": 100,
+  "target-throughput": 0.5
+}
```
dliappis, line 183: nit: add newline

http_logs/challenges/common/default-schedule.json
```diff
@@ -0,0 +1,183 @@
+{
```
dliappis, line 1: With this level of indentation the rendered track by Rally looks like:

```
  "challenges": [
        {
      "name": "append-no-conflicts",
      "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only. After that a couple of queries are run.",
      "default": true,
      "schedule": [
        {
  "operation": "delete-index"
},
{
  "operation": {
    "operation-type": "create-index",
    "settings": {}
  }
},
```

to make things look proper when rendered internally by Rally (you can check those rendered files in your $TMPDIR, they are useful for debugging) we should indent by 8 characters

dliappis (commented):
Thanks for iterating! This looks very close to ready, left a few small comments.

dliappis (commented):
This is basically LGTM.

@nik9000 would it be possible to address the comments in https://github.com/elastic/rally-tracks/pull/132/files#r527070658 , https://github.com/elastic/rally-tracks/pull/132/files#r527063535 and https://github.com/elastic/rally-tracks/pull/132/files#r527063075 before we merge?

dliappis:
Flagged as backport pending; will do this after a few days.

dliappis:
Backported to 7 in https://github.com/elastic/rally-tracks/commit/98a30cef87d41acf166aea04436976a7c8a42ea7

will not backport further to 7.1/7.0.

PR #137, opened 2020-10-15
Lower query throughput in geonames

danielmitterdorfer:
With this commit we lower the target throughput for two queries in the
`geonames` track. For both we want to target roughly 70% utilization:

* `term`: We measured a median service time of roughly 7ms which
corresponds to a target throughput of 100 ops/s with 70% utilization.
* `phrase`: We measured a median service time of roughly 6ms which
corresponds to a target throughput of 110 ops/s with 70% utilization.


dliappis (approved):
LGTM thanks!

PR #138, opened 2020-10-26
Remove random score functions in geonames

mayya-sharipova:
Random score queries, having random components, don't guarantee
stable query service time (and latency) and create problems
for investigations.

There were originally added in PR #53 to compare the performance
between function_score and script_score. As we have enough
data for this comparison (random_score query in script_score showed
to perform better than in funciton_score), we can remove these
operations now.

### DIFF
--- a/geonames/challenges/default.json
+++ b/geonames/challenges/default.json
@@ -147,18 +147,6 @@
           "target-throughput": 1.5
         },
         {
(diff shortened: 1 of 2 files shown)


hub-cap (approved):
LGTM, will merge and backport

PR #140, opened 2020-11-10
Add EQL track

matriv:
Add a new track to benchmark EQL performance.

### DIFF
--- a/eql/track.json
+++ b/eql/track.json
@@ -1,0 +1,158 @@
+{% import "rally.helpers" as rally with context %}
+{
+  "version": 2,
+  "description": "Tracker-generated track for eql",
+  "default": true,
+  "indices": [
+    {
+      "name": "endgame-4.28.2-000001",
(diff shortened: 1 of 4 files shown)


eql/track.json
```diff
@@ -0,0 +1,158 @@
+{% import "rally.helpers" as rally with context %}
+{
+  "version": 2,
+  "description": "Tracker-generated track for eql",
```
dliappis, line 4: Let's change this description to something useful.

dliappis (approved):
I left a comment about changing the (auto generated) description.

~Also let's mention in the README that this track is designed to be executed against Elasticsearch master only~
Offline I understood this will work also with 7.10+ so after merging we should backport to the `7` branch.

Otherwise LGTM, no need for another review cycle.

matriv:
@dliappis That's correct we want it in >= 7.10

eql/track.json
matriv: Sure, done!

PR #125, opened 2020-06-05
Add operation for sub-string-terms

nik9000:
Adds two operations to compare the speed of the `terms` agg run on
strings as a sub-aggregation to test https://github.com/elastic/elasticsearch/pull/57758

PR #124, opened 2020-06-04
Iterate on date_histogram metrics

nik9000:
Iterates a bit on the date_histogram metrics, filtering "bad" data and increasing the throughput targets accordingly.

PR #118, opened 2020-05-26
Add metrics to date-histogram

nik9000:
This'll be useful for comparing the performance of `auto_date_histogram`
with `date_histogram` and for comparing `auto_date_histogram` under some
new optimization that I'm working on.

PR #113, opened 2020-04-28
Add a track for date_histogram

nik9000:
Adds a track to nyc taxis to test `date_histogram` performance.

PR #142, opened 2020-11-16
Fix filter on date_histogram in nyc_taxis

nik9000:
It looks like there is a bug in one of the `date_histogram`s of
nyc_taxis. The `query` filtering the histogram outside the request body
so it doesn't do anything.

### DIFF
--- a/nyc_taxis/operations/default.json
+++ b/nyc_taxis/operations/default.json
@@ -121,16 +121,16 @@
     {
       "name": "date_histogram_calendar_interval",
       "operation-type": "search",
-      "query": {
-        "range": {
-          "dropoff_datetime": {
-            "gte": "2015-01-01 00:00:00",
-            "lt": "2016-01-01 00:00:00"
-          }
-        }
-      },
       "body": {
         "size": 0,
+        "query": {
+          "range": {
+            "dropoff_datetime": {
+              "gte": "2015-01-01 00:00:00",
+              "lt": "2016-01-01 00:00:00"
+            }
+          }
+        },
         "aggs": {
           "dropoffs_over_time": {
             "date_histogram": {


dliappis (approved):
LGTM

One unrelated observation is that none of the queries seems to have a target throughput (therefore they run as fast as they can). I don't know if this is on purpose, but given this operation is exclusively used by the `date-histogram` challenge which isn't part of nightlies, I guess this is ok.

nik9000:
> One unrelated observation is that none of the queries seems to have a target throughput (therefore they run as fast as they can).

I think I added them with target_intervals just because it was easier to do the math on my side.

nik9000:
@dliappis should I backport this? I'm not familiar with the backport policy for this repo but I'm game for whatever.

dliappis:
> @dliappis should I backport this? I'm not familiar with the backport policy for this repo but I'm game for whatever.

@nik9000 I checked and we don't have the specific track (originally introduced in #113) backported in e.g. branch `7`. Would you like to have this in `7` too? If yes, we should backport the foundational work in #113 , #124 , #125 and then this.

nik9000:
For the most part I use this track to make sure I'm not breaking date_histogram optimizations. If you don't need them backported I don't need them backported.

dliappis:
> For the most part I use this track to make sure I'm not breaking date_histogram optimizations. If you don't need them backported I don't need them backported.

I did some digging and #124 has already been backported to 7 / 7.0 / 7.1 so for the sake of consistency, I'll backport #113, #118, #125, #142 also.

nik9000:
Thanks!

dliappis:
I've now backported #113, #118, #124 plus this PR to 7 / 7.0 / 7.1

PR #143, opened 2020-11-17
Reduce percolator-with-content-google target throughput

dliappis:
Latency charts have occasional spikes because we can't reach the defined
target throughput.

Lower the target throughput of the operation
"percolator_with_content_google" to 27 ops/s.


danielmitterdorfer (approved):
LGTM

PR #145, opened 2020-11-24
Tune EQL queries to get more consistent results.

matriv:
Increase iterations for 2 "fast" queries from 50 to 100 and their warmup
iterations, and also increase the warmup iterations for the rest except
for the `_slow` which seem pretty consistent from the previous nightly
runs.


danielmitterdorfer (approved):
LGTM

PR #148, opened 2020-12-08
Switch runtime_script_grok to _source

nik9000:
This switches the `runtime_script_grok` track to load from `_source` by
default and makes it possible to configure it to load from either
`_source`, a `wildcard` field, or a `keyword` field.

Loading values from `_source` is more realistic for the "saving space"
use case and we should see that in the index size. Loading from
`wildcard` would be more realistic in the "flexibility" use case so that
is useful to run locally, but for now it makes sense to run the "saving
space" use case in rally. Loading values from a `keyword` field is
useful mostly as a baseline to compare to the `wildcard` field. We
*think* `wildcard` is always better at flexibility and `_source` is
always better at saving space, but being able to test `keyword` will
validate our theories.

### DIFF
--- a/http_logs/README.md
+++ b/http_logs/README.md
@@ -45,7 +45,8 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using 
 * `cluster_health` (default: "green"): The minimum required cluster health.
 * `ingest_pipeline`: Only applicable for `--challenge=append-index-only-with-ingest-pipeline`, selects which ingest
 node pipeline to run. Valid options are `'baseline'` (default), `'grok'`  and `'geoip'`. For example: `--challenge=append-index-only-with-ingest-pipeline --track-params="ingest_pipeline:'baseline'" `
-* `runtime_script_grok`: If defined the challenges load the `unparsed` set of documents, indexing the `@timestamp` and the raw `message` field. All of the other fields are declared as `runtime` fields extracting the value usig a script.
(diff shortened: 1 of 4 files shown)


danielmitterdorfer (approved):
It took a while to review this as I tested each of the three possible configurations. The benchmarks themselves look fine and I only found one typo in the docs. LGTM (no need for another review around after fixing the typo and the merge conflict).

nik9000:
I believe I've resolved the typo and merge conflict. I'll give the three options another roll through today to double check everything. @danielmitterdorfer, in the past y'all have merged merge my PRs and backported them. Would you like to do that still? I keep opening PRs over here so maybe I should learn to do it myself?

danielmitterdorfer:
> In the past y'all have merged merge my PRs and backported them. Would you like to do that still? I keep opening PRs over here so maybe I should learn to do it myself?

It would be great if you could pick this up. Backports are done by cherry-picking the commit on the right branch (in your case it's only branch `7`). Contrary to the ES repo, we don't to backport PRs but rather push directly to the remote. In order to avoid any disruptions in the nightly benchmarks over the weekend we don't merge PRs on the last day of the week (i.e. on Fridays).

nik9000:
Got it! I can do that.

nik9000:
I've merged and backported! I'll try and remember to look at the nightlies in the morning.

PR #150, opened 2020-12-09
Fix two operations around runtime fields

### LINKED ISSUE #149: 200s-in-range query latency is unstable
In the [nightly benchmarks](https://elasticsearch-benchmarks.elastic.co/#tracks/http-logs/nightly/default/30d) we've observed that latency is unstable for the query `200s-in-range`:

nightly-basic-http_logs-4g-200s-in-range-latency

When we inspect latency over time in a single benchmark, we see that it is consistenly rising:

latency-200s-in-range

Service time in a single benchmark is around 26ms:

nightly-basic-http_logs-4g-200s-in-range-service-time

This suggests that the target throughput is too high for that configuration and we proposing dropping it, so one request is issued every 30ms, meaning target througput should be lowered to 1 / 30ms = 33 ops/s.

nik9000:
This fixes the throughput for the new `200s-in-range` when run against
standard indexed fields to be something more reasonable. It also fixes
the `term` query test for runtime field. Previously it was running
against an unmapped field so it would return nothing very very quickly.
Now it runs against the runtime field and shows how slow queries against
a runtime field can be when you don't add another query to limit the
docs we have to look at.

Closes #149

### DIFF
--- a/http_logs/challenges/common/default-schedule.json
+++ b/http_logs/challenges/common/default-schedule.json
@@ -1,4 +1,4 @@
-{
+        {
           "operation": "delete-index"
         },
         {
@@ -70,9 +70,9 @@
         },
         {
           "operation": "term",
-          "warmup-iterations": 500,
-          "iterations": 100,
-          "target-throughput": 50
+          "warmup-iterations": {%- if runtime_script_grok is defined %}50{% else %}500{% endif %},
+          "iterations": {%- if runtime_script_grok is defined %}10{% else %}100{% endif %},
+          "target-throughput": {%- if runtime_script_grok is defined %}0.125{% else %}50{% endif %}
         },
         {
           "operation": "range",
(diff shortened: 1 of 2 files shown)


http_logs/challenges/common/default-schedule.json
```diff
@@ -70,9 +70,9 @@
         },
         {
           "operation": "term",
-          "warmup-iterations": 500,
-          "iterations": 100,
-          "target-throughput": 50
+          "warmup-iterations": {%- if runtime_script_grok is defined %}50{% else %}500{% endif %},
+          "iterations": {%- if runtime_script_grok is defined %}10{% else %}100{% endif %},
+          "target-throughput": {%- if runtime_script_grok is defined %}0.125{% else %}50{% endif %}
```
nik9000, line 75: On my desktop this runs in about seven seconds so I think targeting 8 seconds in CI is a good start. We can tune it from there.

danielmitterdorfer (approved):
Thanks for the fix. LGTM

dliappis (approved):
LGTM

PR #151, opened 2020-12-21
Lower throughput target for runtime 200s in range

nik9000:
Now that using `_source` as the origin of runtime fields has settled I'm
looking into the throughput targets. It looks like 400s in range is
fine. The service time for 200s in range varies between 60ms and 140ms.
That variation ain't great. It turns out that we're targeting a
throughput of 15 operations per second which is a little tight for the
60ms service time. This switches the new target to 5 operations per
second which has a ton of extra space, even for the 140ms service time.
Hopefully we'll see some more stability in the service time after this.

### DIFF
--- a/http_logs/challenges/common/default-schedule.json
+++ b/http_logs/challenges/common/default-schedule.json
@@ -84,7 +84,7 @@
           "operation": "200s-in-range",
           "warmup-iterations": 500,
           "iterations": 100,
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM. @DJRickyB FYI this will change throughput chart and we'll need an annotation.

dliappis:
Backported in 7 via e59a12f1089d4edd29fe6b350eaf4785df9d0b8b

PR #152, opened 2020-12-21
Fix term for source runtime fields

nik9000:
When runtime fields are based on `_source` (added in #148) we were mapping
`request.keyword` instead of `request.raw` what was causing term queries
to quickly fail, giving incorrect timings.


dliappis (approved):
LGTM

dliappis:
Backported in 7 via 7994c4b983b2839b6ef8fe0003f7f79398dda1f5

PR #153, opened 2020-12-23
Make slow runtime field query target realistic

nik9000:
In #152 we fixed a query against runtime fields so it would not longer
erroniously report that the request took 10ms. The error took 10ms to
generate..... OK! Now that we run the actual request against _source (
see #148 for that) our old througput target of about 8 seconds per
request ain't gonna work - `_source` is slower. About 18 seconds per
request. This updates the throughput target.


DJRickyB (approved):
LGTM!  Thank you

DJRickyB:
backported to 7

PR #157, opened 2021-02-10
reduce target-throughput for tasks that are a little tight in nightlies

DJRickyB:
service_time increased slightly for geonames tasks "country_agg_uncached" and "expression" as a result of recent benchmark environment changes.  These increases caused a multi-fold increase in measured latency, which is affecting readability in our nightly benchmarks charts.  This change adjusts the throughput target for these two tasks and will accompany an annotation in the charts on elasticsearch-benchmarks.elastic.co

### DIFF
--- a/geonames/challenges/default.json
+++ b/geonames/challenges/default.json
@@ -90,7 +90,7 @@
           "operation": "country_agg_uncached",
           "warmup-iterations": 200,
           "iterations": 100,
-          "target-throughput": 3.6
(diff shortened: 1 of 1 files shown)


nik9000 (approved):
I'm coll with this so long as we know that its part of the environment change. If `country_agg_uncached` gets worse for other reasons I wanna know!

dliappis (approved):
LGTM, let's annotate when it becomes effective.

PR #160, opened 2021-03-04
Fix error in field name

nik9000:
I made a copy-and-paste error which broke all non-grok fields. I guess I
didn't test well enough. I'm sorry!


nik9000:
I haven't fully tested this locally but it sure looks less broken.....

nik9000:
Locally these look much better:
```
curl -HContent-Type:application/json -XPOST 'localhost:39200/_search?pretty&error_trace' -d'{
  "query": {
    "bool": {
      "must": [
        {
          "range": {
            "@timestamp": {
              "gte": "1998-05-01T00:00:00Z",
              "lt": "1998-05-02T00:00:00Z"
            }
          }
        },
        {
          "match": {
            "dissect.from_source.status": "400"
          }
        }
      ]
    }
  }
}'
```
has results now!

danielmitterdorfer (approved):
Thanks for the fix!

PR #163, opened 2021-03-18
Add benchmark for top_metrics fetching keyword

nik9000:
Tests for the speed cost of implementing
https://github.com/elastic/elasticsearch/pull/70579

### DIFF
--- a/noaa/challenges/default.json
+++ b/noaa/challenges/default.json
@@ -338,6 +338,13 @@
           "warmup-iterations": 10,
           "iterations": 50,
           "target-throughput": 1
+        },
+        {
+          "operation": "last_country_code_per_station_top_metrics_5000",
+          "clients": 1,
+          "warmup-iterations": 10,
+          "iterations": 50,
+          "target-throughput": 1
         }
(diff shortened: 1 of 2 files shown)


danielmitterdorfer (approved):
Thanks! LGTM

dliappis:
@nik9000 how far back should we backport this (specifically wondering about 7.x)?

nik9000:
It won't work before about 7.13. Before that it's just errors. Might be
7.12. But doesn't need to go back that far because we aren't running in
nighties. Just 7.latest is cool

dliappis:
> It won't work before about 7.13. Before that it's just errors. Might be 7.12. But doesn't need to go back that far because we aren't running in nighties. Just 7.latest is cool

We have a branch `7.12` which covers `7.12` (inclusive) and everything newer than that. We are good with that I guess?

nik9000:
great!

dliappis:
Cherry picked in 7.12 via fadc9a887b3a52681551171b00cf416e51e60ff9

PR #164, opened 2021-03-26
Fix the metricbeat track

nik9000:
It uses parameters that don't work against ES's master branch.

### DIFF
--- a/metricbeat/index.json
+++ b/metricbeat/index.json
@@ -3,10 +3,7 @@
     "index.number_of_shards": {{number_of_shards | default(1)}},
     "index.number_of_replicas": {{number_of_replicas | default(0)}},
     "index.requests.cache.enable": false,
-    "index.mapping.total_fields.limit": 20000,
-    "index.search.slowlog.level" : "debug",
-    "index.search.slowlog.threshold.fetch.debug" : "0s",
-    "index.search.slowlog.threshold.query.debug" : "0s"
+    "index.mapping.total_fields.limit": 20000
   },
   "mappings" : {
     "_meta" : {


metricbeat/index.json
```diff
@@ -3,10 +3,7 @@
     "index.number_of_shards": {{number_of_shards | default(1)}},
     "index.number_of_replicas": {{number_of_replicas | default(0)}},
     "index.requests.cache.enable": false,
-    "index.mapping.total_fields.limit": 20000,
-    "index.search.slowlog.level" : "debug",
-    "index.search.slowlog.threshold.fetch.debug" : "0s",
-    "index.search.slowlog.threshold.query.debug" : "0s"
```
dliappis, line 9: For master according to the [docs](https://www.elastic.co/guide/en/elasticsearch/reference/master/index-modules-slowlog.html) the above three would become:

```
"index.search.slowlog.threshold.fetch.debug": "0s"
"index.search.slowlog.threshold.query.debug": "0s"
```

I am not sure why they were added in the original PR: https://github.com/elastic/rally-tracks/pull/56 to be honest, but since these are dynamic settings we could leverage something else (e.g. an operation before the query we are after) to debug things? Or make these conditional? WDYT?

dliappis (commented):
Thank you. Left a question.

metricbeat/index.json
nik9000: I can change them to their new versions if you'd prefer. I figured we could add them back if we felt like we needed them. I don't think the track itself wants them.

dliappis: If they aren't wanted I am good.

dliappis:
Thank you @nik9000 for the fix!

PR #165, opened 2021-04-01
Benchmark for date_histo with one bucket

nik9000:
This adds a benchmark for `date_histogram` when a single bucket covers
the entire index. We expect this to be quite common for buckets with a
size near a day. Its fairly common for folks to use "daily" indices that
contain *about* a day's worth of data. That day isn't likely to line up
*exactly* with daily buckets. But for buckets marginally larger than a
day and for indices even a little smaller than a day I think it'll be
quite common.

Benchmark for https://github.com/elastic/elasticsearch/pull/71180

### DIFF
--- a/noaa/challenges/default.json
+++ b/noaa/challenges/default.json
@@ -430,6 +430,13 @@
           "target-interval": 3
         },
         {
+          "operation": "date-histo-entire-range",
+          "clients": 1,
+          "warmup-iterations": 100,
+          "iterations": 500,
+          "target-interval": 0.2
+        },
+        {
           "operation": "date-histo-numeric-terms",
           "clients": 1,
           "warmup-iterations": 10,
--- a/noaa/operations/default.json
+++ b/noaa/operations/default.json
@@ -839,6 +839,21 @@
       }
     },
     {
+      "name": "date-histo-entire-range",
+      "operation-type": "search",
+      "body": {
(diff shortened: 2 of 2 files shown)


dliappis (approved):
LGTM.

dliappis:
@nik9000 will merge this on Monday (typically we won't merge track changes on Fri to prevent chances of broken nightlies during the weekend). Question: how far back should this be backported?

nik9000:
Thanks!

dliappis:
@nik9000 regarding backporting, how far back can this go?

nik9000:
It can run as far back as you'd like. I likey won't run it on old branches
though. The code should work on most versions of es after .say, 1.0.

dliappis:
> The code should work on most versions of es after .say, 1.0.

Do you mean after `7.12.0` or?

nik9000:
> Do you mean after `7.12.0` or?

I mean its safe to backport it as far as you like. But 7.12 is certainly far enough for me. 7.13 even.

dliappis:
> I mean its safe to backport it as far as you like. But 7.12 is certainly far enough for me. 7.13 even.

Backported to 7.12 using 11cee00fe9548510ebcb71138d8f981e27be225e

nik9000:
Thanks!

PR #166, opened 2021-04-01
reduce impact of warmups in runtime fields on overall challenge duration

DJRickyB:


DJRickyB:
most of these are excluded from nightly runs but six of the `400s-in-range` tests do execute, meaning this should yield back roughly 2 hours and 12 minutes to nightly

ebadyano (approved):
LGTM, thank you

PR #168, opened 2021-04-20
Use one primary shard for geopointshape

dliappis:
Switch from the default of `5` to `1` for primary shards of the
geopointshape track. Using `5` was legacy copy/paste from geopoint
back from the times when the ES default was `5`.

A follow up commit will switch the geopoint default to `1` as well.


iverase (approved):
LGTM. I think it makes much more sense than having 5 shards for that amount of data

dliappis:
Backported to `7.12`, `7.2`, `7.1`, `7.0`, `7`, `6`

PR #171, opened 2021-04-28
update to use new on_prepare_track API

DJRickyB:
elastic/rally#1228 broke the RuntimeFieldResolver with an API change, this updates the API to the new paradigm

also relates elastic/rally#1253

### DIFF
--- a/http_logs/track.py
+++ b/http_logs/track.py
@@ -1,8 +1,8 @@
 from copy import copy
 import re
 
-from esrally import exceptions
 from esrally.track import loader
+
 
 def reindex(es, params):
     result = es.reindex(body=params.get("body"), request_timeout=params.get("request_timeout"))
@@ -27,7 +27,7 @@ class RuntimeFieldResolver(loader.TrackProcessor):
                     task.operation.params = self._replace_field(f"{impl}.from_{source}.", task.operation.params)
 
     def on_prepare_track(self, track, data_root_dir):
-        return True
+        return []
 
     def _replace_field(self, field, t):
         if t == 'path' or t == 'status':
@@ -48,3 +48,4 @@ def register(registry):
     else:
         registry.register_runner("reindex", reindex)
     registry.register_track_processor(RuntimeFieldResolver())
+    registry.register_track_processor(loader.DefaultTrackPreparator())


http_logs/track.py
```diff
@@ -48,3 +48,4 @@ def register(registry):
     else:
         registry.register_runner("reindex", reindex)
     registry.register_track_processor(RuntimeFieldResolver())
+    registry.register_track_processor(loader.DefaultTrackPreparator())
```
dliappis, line 51: Increase backwards compatibility using something like:

```python
        # TODO remove this backwards compatibility hatch after several Rally releases
        # ref: https://github.com/elastic/rally/pull/1228 and https://github.com/elastic/rally/issues/1166
        class EmptyTrueList(list):
            def __bool__(self):
                return True

            def __eq__(self, other):
                if isinstance(other, bool):
                    return True

        return EmptyTrueList()
```

To make life easier, this is the diff with my suggestion:

```
diff --git a/http_logs/track.py b/http_logs/track.py
index 53dc2db..bc50f2d 100644
--- a/http_logs/track.py
+++ b/http_logs/track.py
@@ -4,6 +4,7 @@ import re
 from esrally import exceptions
 from esrally.track import loader
 
+
 def reindex(es, params):
     result = es.reindex(body=params.get("body"), request_timeout=params.get("request_timeout"))
     return result["total"], "docs"
@@ -13,6 +14,7 @@ async def reindex_async(es, params):
     result = await es.reindex(body=params.get("body"), request_timeout=params.get("request_timeout"))
     return result["total"], "docs"
 
+
 class RuntimeFieldResolver(loader.TrackProcessor):
     PATTERN = re.compile('.+-from-(.+)-using-(.+)')
 
@@ -27,7 +29,17 @@ class RuntimeFieldResolver(loader.TrackProcessor):
                     task.operation.params = self._replace_field(f"{impl}.from_{source}.", task.operation.params)
 
     def on_prepare_track(self, track, data_root_dir):
-        return True
+        # TODO remove this backwards compatibility hatch after several Rally releases
+        # ref: https://github.com/elastic/rally/pull/1228 and https://github.com/elastic/rally/issues/1166
+        class EmptyTrueList(list):
+            def __bool__(self):
+                return True
+
+            def __eq__(self, other):
+                if isinstance(other, bool):
+                    return True
+
+        return EmptyTrueList()
 
     def _replace_field(self, field, t):
         if t == 'path' or t == 'status':
@@ -41,6 +53,7 @@ class RuntimeFieldResolver(loader.TrackProcessor):
             }
         return t
 
+
 def register(registry):
     async_runner = registry.meta_data.get("async_runner", False)
     if async_runner:
@@ -48,3 +61,10 @@ def register(registry):
     else:
         registry.register_runner("reindex", reindex)
     registry.register_track_processor(RuntimeFieldResolver())
+    try:
+        registry.register_track_processor(loader.DefaultTrackPreparator())
+    except TypeError as e:
+        # TODO remove this backwards compatibility hatch after several Rally releases
+        # ref: https://github.com/elastic/rally/pull/1228 and https://github.com/elastic/rally/issues/1166
+        if e == "__init__() missing 1 required positional argument: 'cfg'":
+            pass
```

http_logs/track.py
```diff
@@ -1,9 +1,9 @@
 from copy import copy
 import re
 
-from esrally import exceptions
```
dliappis, line 4: See comment below, if we go the backwards compatibility route, we need to keep it imported.

dliappis (commented):
Thanks for coming up quickly with this.

I am concerned about the popularity of the `http_logs` track and breaking it's use when we backport this to the `7.12` branch with recently released Rally versions when running benchmarks against ES>=7.12.0 and ES<8.0.0

I suggested some backwards compatibility hacks, which we could get rid of after a certain amount of releases.

dliappis:
After an offline discussion with @DJRickyB we agreed that it's best to handle backwards compatibility on the Rally side.
For now merging this in master only, which means that running http_logs against ES master branches will require using Rally from master and we'll hold off a backport of this until we come up with a backwards compatible change on the Rally side.

DJRickyB:
@dliappis found something further...

see: https://discuss.elastic.co/t/rally-on-docker-will-not-list-tracks-cannot-list-could-not-register-track-plugin/271510

for the `list` command (and possibly others) we rely on master rally-tracks as there's no distribution to be gleaned, thus this PR has broken existing rally release functionality. reverting this until we have a backward-compatible solution

PR #172, opened 2021-04-28
Revert "update to use new on_prepare_track API"

DJRickyB:
Reverts elastic/rally-tracks#171.  see https://github.com/elastic/rally-tracks/pull/171#issuecomment-828598581


DJRickyB:
cheating in to unblock users

PR #173, opened 2021-04-28
on_prepare_track update with backward compatibility

DJRickyB:
After #171 , compatibility was broken for functionality that relied on `master` rally-tracks, specifically the default behavior of `info` and `list races` in existing releases.  This re-implements 171 with backward functionality.  

Tested `race`, `info`, and `list races` with each esrally 2.2.0 and elastic/rally master

### DIFF
--- a/http_logs/track.py
+++ b/http_logs/track.py
@@ -61,10 +61,9 @@ def register(registry):
     else:
         registry.register_runner("reindex", reindex)
     registry.register_track_processor(RuntimeFieldResolver())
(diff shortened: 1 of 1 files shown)


dliappis:
@DJRickyB I backported this to `7.12` in 6d44309b32aa9a6f22dca6bbf14fd74dcb5743ab in order to ensure that the current Rally master can run http_logs against ES >=7.12.0. Note that it doesn't any further backporting as track processors existing in rally-tracks/http_logs only in 7.12 and master (at the moment).

PR #180, opened 2021-06-10
Decrease target-throughput of painless in geonames

dliappis:
Painless static and dynamic are spiking unnecessarily in
https://elasticsearch-benchmarks.elastic.co, with the service_time
occasionally just below 700ms.

Lower target-thoughput so that latency will not increase
as long as queries get serviced within ~714ms.


ebadyano (approved):
lgtm

dliappis:
Will merge on Monday.

PR #182, opened 2021-06-17
Deprecating eventdata track

ebadyano:
Added user-info message to deprecated the track. The track will be remove in a follow up pr soon to reduce maintenance effort. This is not the [eventdata](https://github.com/elastic/rally-eventdata-track) track that we use in many benchmarks but a version with a static corpus.

### DIFF
--- a/eventdata/challenges/default.json
+++ b/eventdata/challenges/default.json
@@ -1,5 +1,6 @@
     {
       "name": "append-no-conflicts",
+      "user-info": "This track is deprecated. Please use https://github.com/elastic/rally-eventdata-track.",
       "description": "Indexes the whole document corpus using Elasticsearch default settings. We only adjust the number of replicas as we benchmark a single node cluster and Rally will only start the benchmark if the cluster turns green. Document ids are unique so all index operations are append only.",
       "default": true,
       "schedule": [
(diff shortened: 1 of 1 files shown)


eventdata/challenges/default.json
```diff
@@ -1,5 +1,6 @@
     {
       "name": "append-no-conflicts",
+      "user-info": "This track is deprecated. Please use https://github.com/elastic/rally-eventdata-track.",
```
danielmitterdorfer, line 3: Can we please add this property for all challenges? Also, I suggest to change the wording slightly: `This track is deprecated and will be removed soon. ...`. Wdyt?

danielmitterdorfer (commented):
Thanks! I left a comment regarding adding this for all challenges of this track.

Can you please also [explain the motivation for this change](https://chris.beams.io/posts/git-commit/#why-not-how) in the PR description? Unfortunately, this will then not be automatically reflected in your commit once you merge the PR so you need to copy&paste it then.

eventdata/challenges/default.json
ebadyano: I will update the wording. For other challenges should I just add this property with empty content?

danielmitterdorfer (approved):
Thanks for iterating! LGTM

ebadyano:
@danielmitterdorfer Thank you for the review

PR #184, opened 2021-06-28
Improve consistency of geo tracks

imotov:
The execution speed of geo tracks may change drastically depending on 
the BKD tree structures in different segments. This change ensures that 
we have only one segment before running the search queries.


danielmitterdorfer (approved):
Thanks! LGTM

danielmitterdorfer:
I've backported the PR to all relevant branches.

PR #187, opened 2021-07-06
Add `cluster` param to eql track for CCS

matriv:
Add a new param: `cluster` which can be set to indicate
which configured remote cluster to hit when running the EQL
benchmarks.

### DIFF
--- a/eql/track.py
+++ b/eql/track.py
@@ -4,9 +4,9 @@ import logging
 async def eql(es, params):
     logger = logging.getLogger(__name__)
     logger.info("Provided params [%s].", params)
-    cluster = ""
-    if params.get("cluster"):
-        cluster = params.get("cluster") + ":"
+    if cluster := params.get("cluster", ""):
+        cluster += ":"
+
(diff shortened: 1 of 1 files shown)


matriv:
PS. This is just for the future, used it only manually to execute CCS related benchmarks.

eql/track.py
```diff
@@ -4,8 +4,11 @@
 async def eql(es, params):
     logger = logging.getLogger(__name__)
     logger.info("Provided params [%s].", params)
+    cluster = ""
+    if params.get("cluster"):
+        cluster = params.get("cluster") + ":"
```
dliappis, line 9: The three lines above can be replaced with:

```
if cluster := params.get("cluster", ""):
    cluster += ":"
```

dliappis (approved):
LGTM. I left a suggestion for something simpler.

matriv:
Backported to `7.14`, `7.12` & `7.2`.

dliappis:
Thanks for all the work here!

PR #191, opened 2021-07-16
Increase iterations for date histogram aggregation

danielmitterdorfer:
We have seen warmup issues in our nightly benchmarks for the date
histogram aggregation recently. With this commit we increase the number
of warmup and also measurement iterations to provide more time to warmup
and get more stable measurement results.


dliappis (approved):
LGTM

Issue #193, opened 2021-07-19
Force-merge to one segment by default


danielmitterdorfer:
Historically our tracks resort to the default behavior of the force-merge API and did not specify the desired number of segments. Recently we modified the force-merge behavior in several tracks (see https://github.com/elastic/rally-tracks/pull/185 and https://github.com/elastic/rally-tracks/pull/186) to improve stability of benchmark results. As this is desirable in general, we should change the default behavior in all tracks that peform any actions after indexing (see also #192) to force-merge to one segment unless specified otherwise (i.e. setting `max_num_segments` to `null`). Each track's README should include a clear indication of this behavior *and* an example how to achieve the behavior of force-merge.

ebadyano:
After trying force merge with noaa, we decided that there was no noticeable improvement with stability, so doing force merge to one segment by default did not prove to be a suitable stability solution and needs to be done as needed by a benchmark. The change for noaa was reverted in https://github.com/elastic/rally-tracks/pull/204. closing this issue.

PR #194, opened 2021-07-21
Force-merge to one segment in noaa by default

ebadyano:
Updated the default for force-merge parameter max_num_segments to 1, to
improve the stability of the benchmark and to be consistent with the
other tracks.

Relates to https://github.com/elastic/rally-tracks/issues/193

### DIFF
--- a/noaa/README.md
+++ b/noaa/README.md
@@ -53,7 +53,7 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using 
 * `index_settings`: A list of index settings. Index settings defined elsewhere (e.g. `number_of_replicas`) need to be overridden explicitly.
 * `cluster_health` (default: "green"): The minimum required cluster health.
 * `error_level` (default: "non-fatal"): Available for bulk operations only to specify ignore-response-error-level.
-* `max_num_segments` (default: 1): An integer specifying the max amount of segments the force-merge operation should use. To unset, set the parameter to `null`.
+* `max_num_segments` (default: 1): An integer specifying the max amount of segments the force-merge operation should use. To unset, set the parameter to `null` with [track parameters via JSON file](https://esrally.readthedocs.io/en/stable/command_line_reference.html#track-params).
 
 ### License
 
--- a/noaa/README.md
+++ b/noaa/README.md
@@ -53,6 +53,7 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using 
 * `index_settings`: A list of index settings. Index settings defined elsewhere (e.g. `number_of_replicas`) need to be overridden explicitly.
 * `cluster_health` (default: "green"): The minimum required cluster health.
 * `error_level` (default: "non-fatal"): Available for bulk operations only to specify ignore-response-error-level.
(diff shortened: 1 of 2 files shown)


danielmitterdorfer (commented):
Thanks! I left some comments.

danielmitterdorfer:
@ebadyano you've requested another review but I don't see any changes after my prior review. Is it possible that you've only applied the changes locally but did not push them?

ebadyano:
@danielmitterdorfer Daniel, could please you check if you can see it now. I think you are right and I must have not pushed the changes. Thank you!

noaa/README.md
```diff
@@ -53,6 +53,7 @@ This track allows to overwrite the following parameters with Rally 0.8.0+ using
 * `index_settings`: A list of index settings. Index settings defined elsewhere (e.g. `number_of_replicas`) need to be overridden explicitly.
 * `cluster_health` (default: "green"): The minimum required cluster health.
 * `error_level` (default: "non-fatal"): Available for bulk operations only to specify ignore-response-error-level.
+* `max_num_segments` (default: 1): An integer specifying the max amount of segments the force-merge operation should use. To unset, set the parameter to `null`.
```
danielmitterdorfer, line 56: IMHO we should explicitly state here that this requires that track parameters are provided via a JSON file (with a link to the docs). Passing `null` via the command line parameter `--track-params` is not supported.

danielmitterdorfer (commented):
Thanks! I left an improvement idea. I also noticed that this logic is not applied to the `top_metrics` and `aggs` challenges yet.

ebadyano:
> Thanks! I left an improvement idea. I also noticed that this logic is not applied to the `top_metrics` and `aggs` challenges yet.

oh good catch, thank you! added to those challenges as well.

ywelsch:
I was wondering if force-merging down to one segment by default will lead to less coverage in the aggs track (e.g. global ordinals no longer being needed) to track performance regressions. Perhaps worth looping in @nik9000?

nik9000:
> I was wondering if force-merging down to one segment by default will lead to less coverage in the aggs track (e.g. global ordinals no longer being needed) to track performance regressions. Perhaps worth looping in @nik9000?

It'll certainly change things! We don't run the aggs track nightly. We said that wasn't a priority for the next few months. It's important, but its waiting.

Anyway, I think it's fine. But I'd prefer to keep the aggs track without the force merge. It just feels like a more common case for us. We'll get more bumpy results but those bumps are "real life".

ebadyano:
chatted with Nik on slack and while he would prefer to not use force-merge t o1 segment for terms aggs, as we would be measuring slightly different code paths and it is to be something aware for the future he is okay with us to proceed with the change for all challenges.

@danielmitterdorfer is it okay to merge the change? Thanks, Evgenia

danielmitterdorfer:
Thanks for the clarification. Let's move forward with this and merge it.

danielmitterdorfer (approved):
LGTM

ebadyano:
@danielmitterdorfer thank you for the review! I will merge  this on Monday.

danielmitterdorfer:
@ebadyano I noticed that the `backport pending` label is still applied; do you intend to backport this?

PR #196, opened 2021-08-12
Add concurrent indexing/querying challenge to PMC track

DJRickyB:


pmc/challenges/indexing-querying.json
```diff
@@ -0,0 +1,128 @@
...
+      "clients": {{bulk_indexing_clients | default(8)}},
+      "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
+    },
+    {
+      "name": "refresh-after-index",
+      "operation": "refresh"
+    },
+    {
+      "operation": {
+        "operation-type": "force-merge",
```
ebadyano, line 54: did you mean to also `force_merge_max_num_segments` to this new challenge?

pmc/challenges/indexing-querying.json
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+      "parallel": {
+        "tasks": [
+          {
+            "name": "concurrent-index-append",
+            "operation": {
+              "operation-type": "bulk",
+              "bulk-size": {{bulk_size | default(500)}},
+              "ingest-percentage": 50
```
ebadyano, line 83: do we only want to index half of the data?

b-deam: I think this makes sense, we ingest the first half here:
https://github.com/elastic/rally-tracks/blob/ee2a1f06940b62f34a696a6a0bebf32e7e25993d/nyc_taxis/challenges/indexing-querying.json#L32-L37

Then during querying, we ingest the remaining 50%:
https://github.com/elastic/rally-tracks/blob/ee2a1f06940b62f34a696a6a0bebf32e7e25993d/nyc_taxis/challenges/indexing-querying.json#L61-L82

DJRickyB:
I originally thought that might be the case as well but i was mistaken.
Same 50% each time, could probably put it up to 100 but this is what I've
been using in recent benchmarks

pmc/challenges/default.json
```diff
@@ -143,6 +143,9 @@
         {
           "operation": {
             "operation-type": "force-merge",
+{%- if force_merge_max_num_segments is defined %}
```
ebadyano, line 146: I think this added force merge max segments to `append-no-conflicts-index-only` challenge, but not the new one

ebadyano (approved):
lgtm

b-deam (commented):
LGTM 2️⃣

PR #197, opened 2021-08-16
Force-merge to one segment in nested by default

ebadyano:
Updated the default for force-merge parameter max_num_segments to 1, to
improve the stability of the benchmark and to be consistent with the
other tracks.

Relates to #193

### DIFF
--- a/nested/challenges/default.json
+++ b/nested/challenges/default.json
@@ -153,6 +153,7 @@
             "operation-type": "force-merge",
             "request-timeout": 7200{%- if force_merge_max_num_segments is not none %},
             "max-num-segments": {{force_merge_max_num_segments}}
+             {%- endif %}
           }
         },
         {
--- a/nested/challenges/default.json
+++ b/nested/challenges/default.json
@@ -1,3 +1,4 @@
+    {% set force_merge_max_num_segments = (max_num_segments | default(1)) %}
     {
       "name": "nested-search-challenge",
       "description": "Indexes the document corpus for an hour using Elasticsearch default settings. After that randomized nested queries are run.",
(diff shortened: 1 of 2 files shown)


nested/challenges/default.json
```diff
@@ -148,7 +151,8 @@
         {
           "operation": {
             "operation-type": "force-merge",
-            "request-timeout": 7200
+            "request-timeout": 7200{%- if force_merge_max_num_segments is not none %},
```
danielmitterdorfer, line 154: The `{%- endif %}` is missing in this block?

If I run `esrally info --track-path=~/.rally/benchmarks/tracks/default/nested` after I've checked out your PR I get:

```
[ERROR] Cannot info. Could not load '/Users/daniel/.rally/benchmarks/tracks/default/nested/track.json'. The complete track has been written to '/var/folders/pg/gbmwwwms3n50fttdmnqd4sl40000gn/T/tmpxb0h_1qa.json' for diagnosis.
	Unexpected end of template. Jinja was looking for the following tags: 'elif' or 'else' or 'endif'. The innermost block that needs to be closed is 'if'.
```

danielmitterdorfer (changes requested):
Unfortunately the track cannot be loaded as is. If that is fixed, the change should be fine.

nested/challenges/default.json
ebadyano: oh, thank you for catching!

danielmitterdorfer (approved):
LGTM

pquentin:
@DJRickyB Do you know what the status of this is? Any reason not to merge?

ebadyano:
will close it without merging, as currently nested doesn't have a stability issue and so force-merge to one segment is unnecessary see https://github.com/elastic/rally-tracks/issues/193#issuecomment-1248381716

PR #200, opened 2021-10-11
Deprecate sorted indexing challenge in geonames

danielmitterdorfer:
The sorted indexing challenge in the `geonames` track turns out to be
unstable. As we have other options that are more reliable (such as
`http_logs`) we deprecate this challenge for the `geonames` track.

### DIFF
--- a/geonames/challenges/default.json
+++ b/geonames/challenges/default.json
@@ -269,6 +269,7 @@
     {
       "name": "append-sorted-no-conflicts",
       "description": "Indexes the whole document corpus in an index sorted by country_code field in ascending order. Document ids are unique so all index operations are append only.",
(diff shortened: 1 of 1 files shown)


geonames/challenges/default.json
```diff
@@ -269,6 +269,7 @@
     {
       "name": "append-sorted-no-conflicts",
       "description": "Indexes the whole document corpus in an index sorted by country_code field in ascending order. Document ids are unique so all index operations are append only.",
+      "user-info": "This challenge is deprecated and will be removed soon. Use the corresponding challenge in http_logs to benchmark sorted indices.",
```
DJRickyB, line 272: TIL about `user-info`.  Can we document it? Seems useful

PR #201, opened 2021-10-11
Reduce target throughput for noaa disjunction big range

DJRickyB:
In our nightly benchmarks, we see some spikiness in `nightly-basic-noaa-add-defaults-range_field_disjunction_big_range_small_term_query-latency`, particularly after Aug 16th, which is an annotated regression.
image

Given we are charting latency, rather than service_time, it seems that perhaps the task is over-scheduled, and that the smaller spikes are probably very similar to the larger spikes (e.g. that we have "runaway" latency).  To check this theory, we chart service_time:
image

Our theory holds true.  This change increases the target allowable latency from an effective 166.7 ms to 200 ms by reducing target throughput from 6 to 5.

### DIFF
--- a/noaa/challenges/default.json
+++ b/noaa/challenges/default.json
@@ -107,7 +107,7 @@
           "operation": "range_field_disjunction_big_range_small_term_query",
           "warmup-iterations": 100,
           "iterations": 500,
-          "target-throughput": 6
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM thank you

PR #206, opened 2021-10-15
http_logs: default to ignore non-fatal errors during index-append

inqueue:
With this PR we set `ignore-response-error-level: non-fatal` for index-append in the http_logs default track schedule. This resolves an issue where index-append could abort if a non-fatal network timeout occurs during bulk indexing. If the client does encounter a timeout, we want the client to ignore the error and continue retrying.

### DIFF
--- a/http_logs/challenges/common/default-schedule.json
+++ b/http_logs/challenges/common/default-schedule.json
@@ -32,7 +32,8 @@
         {
           "operation": "index-append",
           "warmup-time-period": 240,
-          "clients": {{bulk_indexing_clients | default(8)}}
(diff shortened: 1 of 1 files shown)


DJRickyB (approved):
looks good, thank you

dliappis:
@inqueue FYI I rebased it as far back as this could go i.e.:

7.14 -> 1571871647ba36aa33e7bc28410ace02408f5f9d
7.12 -> 040e28f7aa3c223a8ea1ca95a231b88c5d737205
7.2 -> 31df6b80869306c51dabb49d5f3e8c4daf229ab5

PR #210, opened 2021-11-05
Additional challenges for EQL track

Luegg:
Additional challenges. See `eql/README.md` for more explanations about the query selection and parametrization.

### DIFF
--- a/eql/track.json
+++ b/eql/track.json
@@ -1,5 +1,30 @@
 {% import "rally.helpers" as rally with context %}
 {% set cluster = cluster | default('') %}
+
+{% set task_options_10qps %}
+"tags": ["fast"],
+"target-throughput": 10,
+"clients": 2,
+"warmup-iterations": 50,
+"iterations": 100
+{% endset %}
+
+{% set task_options_5qps %}
+"tags": ["fast"],
+"target-throughput": 5,
+"clients": 2,
+"warmup-iterations": 50,
+"iterations": 100
+{% endset %}
+
+{% set task_options_03qps %}
+"tags": ["slow"],
+"target-throughput": 0.3,
+"clients": 2,
+"warmup-iterations": 25,
+"iterations": 50
+{% endset %}
+
 {
   "version": 2,
   "description": "EQL benchmarks based on endgame index of SIEM demo cluster",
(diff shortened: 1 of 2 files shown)


eql/track.json
```diff
@@ -53,21 +56,25 @@
         "ingest-percentage": {{ingest_percentage | default(100)}}
       },
       "clients": {{bulk_indexing_clients | default(8)}},
-      "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
+      "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
+      "tags": "setup"
```
Luegg, line 60: turns out tags are pretty helpful when working with rally, I've also updated Marios document to explain how they can be used for local testing (https://docs.google.com/document/d/1RiG-Itu532RDHQMq6FSt4iEXE0_42K15aayuYDBGk40/edit#heading=h.3um5dshkzbne)

bpintea (commented):
I think this would be useful to add, but as far as replacing the current ones would go, I think keeping those sequences that use string matching (`sequence_2stage_equalityFilter_maxspan1m_fetch1000_size500_slow`) and functions (`sequence_3stage_startsWithfilter_maxspan30m_fetch1000_size100_tail100_head50`) would be valuable.

`sequence_2stage_nofilter_fetch1000_size_1000` and `sequence_2stage_nofilter_fetch1000_size1000_tail1000_head200` (and maybe also `sequence_4stage_nofilter_maxspan5m_fetch1000_size100_head100_tail50`) could maybe be compressed into one query with non-optional but null fields query, since they're otherwise covered by the new queries.

Luegg:
@bpintea I've added another query to cover `startsWith` and eventually replace `sequence_3stage_startsWithfilter_maxspan30m_fetch1000_size100_tail100_head50`. We already have a few occurrences of terms queries in the new tasks so I'm unsure wether we need more to cover `sequence_2stage_equalityFilter_maxspan1m_fetch1000_size500_slow`.

@danielmitterdorfer I've added you as a reviewer mostly to make sure we don't inadvertently blow up the benchmarks too much. I expect that all the additional queries will add ~20 minutes of runtime to the EQL benchmark.

danielmitterdorfer (approved):
Thanks! LGTM

bpintea (approved):
LGTM

PR #211, opened 2021-11-05
remove dependency on mapper-size plugin in EQL track

Luegg:
It looks like the mapper-size plugin is not really needed and I couldn't figure out what it has been used for. Since it makes some steps in the workflow more cumbersome, I think it's best to remove it. First step would be to remove the dependency on the plugin from the track.

Creating this as a separate PR to be able to isolate potential effects on performance (I don't expect any though).


danielmitterdorfer (approved):
LGTM

PR #216, opened 2021-11-09
SQL benchmarks

Luegg:
Introduces SQL benchmarks using the existing NOAA dataset: 33M measurements from weather stations, 6GB in index. The dataset is ideal for running analytical queries like average temp over 12 months in all stations in Canada and seems to match SQL like use cases best (of the already existing datasets).

### DIFF
--- a/noaa/challenges/sql.json
+++ b/noaa/challenges/sql.json
@@ -62,7 +62,6 @@
         {
           "name": "refresh-after-index",
           "operation": "refresh",
-          "clients": 1,
           "tags": ["setup"]
         },
         {
@@ -75,7 +74,6 @@
         {
           "name": "refresh-after-force-merge",
           "operation": "refresh",
-          "clients": 1,
           "tags": ["setup"]
         },
         {
--- a/noaa/challenges/sql.json
+++ b/noaa/challenges/sql.json
@@ -1,0 +1,275 @@
+    {% set task_options_10qps %}
+      "target-throughput": 10,
+      "clients": 2,
+      "warmup-iterations": 50,
+      "iterations": 100,
+      "tags": ["queries"]
+    {% endset %}
+
+    {% set task_options_5qps %}
+    "target-throughput": 5,
+    "clients": 2,
+    "warmup-iterations": 50,
+    "iterations": 100,
+    "tags": ["queries"]
+    {% endset %}
+
+    {% set task_options_1qps %}
+    "target-throughput": 1,
+    "clients": 2,
+    "warmup-iterations": 25,
+    "iterations": 50,
+    "tags": ["queries"]
+    {% endset %}
+
+    {
+      "name": "sql",
+      "description": "Checks the performance of SQL",
+      "default": false,
+      "schedule": [
+        {
+          "operation": "delete-index",
+          "tags": ["setup"]
+        },
+        {
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {{index_settings | default({}) | tojson}}
+          },
+          "tags": ["setup"]
+        },
+        {
+          "name": "check-cluster-health",
+          "operation": {
+            "operation-type": "cluster-health",
+            "index": "weather-data-2016",
+            "request-params": {
+              "wait_for_status": "{{cluster_health | default('green')}}",
+              "wait_for_no_relocating_shards": "true"
+            },
+            "retry-until-success": true
+          },
+          "tags": ["setup"]
+        },
+        {
+          "operation": "index",
+          "#COMMENT": "This is an incredibly short warmup time period but it is necessary to get also measurement samples. As this benchmark is rather about search than indexing this is ok.",
+          "warmup-time-period": 10,
+          "clients": {{bulk_indexing_clients | default(8)}},
+          "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
+          "tags": ["setup"]
+        },
+        {
+          "name": "refresh-after-index",
+          "operation": "refresh",
+          "clients": 1,
+          "tags": ["setup"]
+        },
+        {
+          "operation": "force-merge",
+          "tags": ["setup"],
+          "clients": 1{%- if max_num_segments is defined %},
+          "max-num-segments": {{max_num_segments}}
+          {%- endif %}
+        },
+        {
+          "name": "refresh-after-force-merge",
+          "operation": "refresh",
+          "clients": 1,
+          "tags": ["setup"]
+        },
+        {
+          "operation": {
+            "name": "increase-max-open-scoll-context",
+            "operation-type": "put-settings",
+            "body": {
+              "transient" : {
+                  "search.max_open_scroll_context" : "2000"
+              }
+            }
+          },
+          "tags": ["setup"]
+        },
+        {
+          "operation": {
+            "name": "select_const",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT 1"
+            }
+          },
+          {{ task_options_10qps }}
+        },
+        {
+          "operation": {
+            "name": "select_star",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT * FROM \"weather-data-2016\""
+            }
+          },
+          {{ task_options_10qps }}
+        },
+        {
+          "operation": {
+            "name": "select_field",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT station.name FROM \"weather-data-2016\""
+            }
+          },
+          {{ task_options_10qps }}
+        },
+        {
+          "operation": {
+            "name": "select_scriptField",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT LCASE(SUBSTRING(station.name, 0, 2)) FROM \"weather-data-2016\""
+            }
+          },
+          {{ task_options_10qps }}
+        },
+        {
+          "operation": {
+            "name": "select_stringScriptFilter",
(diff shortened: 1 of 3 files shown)


noaa/challenges/sql.json
```diff
@@ -0,0 +1,275 @@
...
+          },
+          "tags": ["setup"]
+        },
+        {
+          "operation": {
+            "name": "select_const",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT 1"
+            }
```
bpintea, line 99: Is this one needed?

noaa/challenges/sql.json
```diff
@@ -0,0 +1,275 @@
...
+          },
+          {{ task_options_10qps }}
+        },
+        {
+          "operation": {
+            "name": "select_field",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT station.name FROM \"weather-data-2016\""
+            }
```
bpintea, line 119: Needed?

noaa/challenges/sql.json
```diff
@@ -0,0 +1,275 @@
...
+        },
+        {
+          "operation": {
+            "name": "select_stringScriptFilterWithLimit",
+            {# https://github.com/elastic/elasticsearch/issues/80523 #}
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT station.name FROM \"weather-data-2016\" WHERE LCASE(SUBSTRING(station.name, 0, 2)) = 'sa' AND station.elevation > 2800 LIMIT 10"
+            }
+          },
```
bpintea, line 151: Given the non limited query, do we actually need this one?
There is an issue opened on it, but I'm not sure if we should treat the benchmarks as a test case collection.

noaa/challenges/sql.json
```diff
@@ -0,0 +1,275 @@
...
+        {
+          "operation": {
+            "name": "select_stringScriptSort",
+            "operation-type": "sql",
+            "body": {
+              "query": "SELECT station.name FROM \"weather-data-2016\" WHERE station.elevation > 2800 ORDER BY LCASE(SUBSTRING(station.name, 0, 2))"
+            }
+          },
+          {{ task_options_1qps }}
+        },
```
bpintea, line 173: Are there significant difference between numeric and string sorting?
And the same comment would apply to the rest of the numeric vs string operations.

bpintea (commented):
Generally looks good, but I'm wondering if we need so many queries to test with, worrying  not to start glossing over them in time, having in this proposed format 18 (?).
Do we obtain varying enough tests results to be worth splitting, for instance, numeric vs string script filtering/sorting?

Luegg:
> Generally looks good, but I'm wondering if we need so many queries to test with, worrying not to start glossing over them in time, having in this proposed format 18 (?). Do we obtain varying enough tests results to be worth splitting, for instance, numeric vs string script filtering/sorting?

I think we can generally be more generous with benchmark queries than what we used to be. The cost from the processing end is not too big (a 5qps query benchmark should take about 30s to run every night).

I also see your concern about making the weekly checks too time consuming but I thought we can also address it on the presentation side. Not all of the queries need to be presented with the same prominence.

Or can we even use anomaly detection to identify changes in the benchmark results?

On the upside, having more queries can help a lot with validating hypothesis when you see degradations in performance. E.g. `select_const` gives you a baseline for the cost of the whole SQL infrastructure. `select_field` gives you a baseline for fetching data without much processing. If you see that either of these changed together with other benchmarks, you immediately know that there might be some fundamental issues.

Concerning the numeric vs. string scripts, I agree that the distinction is a bit arbitrary but I noted that there is a bit of a chasm between the two types of queries.

From my local results, you see that queries using string scripts generally operate a lot slower:

image

(the top group of queries above 500ms all use string based scripts, the lower group around 100 to 200 is the numeric scripts)

bpintea:
> Or can we even use anomaly detection to identify changes in the benchmark results?

That would be something, I guess the perf team considered this already? I'll leave to to Daniel, if there are thoughts about it.

> select_const gives you a baseline for the cost of the whole SQL infrastructure.
> you immediately know that there might be some fundamental issues.

But would we need this baseline? If there would be a fundamental issue, the other queries will be impacted too, no?

> Concerning the numeric vs. string scripts, I agree that the distinction is a bit arbitrary but I noted that there is a bit of a chasm between the two types of queries.

The scale is different, true, but would we expect the fluctuation to also be different?

Luegg:
> > Or can we even use anomaly detection to identify changes in the benchmark results?
> 
> That would be something, I guess the perf team considered this already? I'll leave to to Daniel, if there are thoughts about it.
> 
> > select_const gives you a baseline for the cost of the whole SQL infrastructure.
> > you immediately know that there might be some fundamental issues.
> 
> But would we need this baseline? If there would be a fundamental issue, the other queries will be impacted too, no?

The baseline helps with a) detecting smaller fluctuations and b) being a shortcut for validating hypothesis.

> 
> > Concerning the numeric vs. string scripts, I agree that the distinction is a bit arbitrary but I noted that there is a bit of a chasm between the two types of queries.
> 
> The scale is different, true, but would we expect the fluctuation to also be different?

I'm not sure to be honest. I can think of a couple of things that are different (e.g. string scripts are probably more memory consuming) and there might be quite some execution paths that differ between the two.

I think in general we have the problem that the surface area of SQL is relatively large (for EQL probably as well). In order to make sure that we catch a majority of degradations there is probably no way around having a certain breadth of queries (it's probably still too few).

I think there are also other ways to improve the dashboards to make detection easier. E.g. we could group all the numeric queries in one panel and only show the 50th percentile.

costin:
This is a great start, LGTM.

noaa/track.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+
+async def sql(es, params):
+    logger = logging.getLogger(__name__)
+    logger.info("Provided params [%s].", params)
```
danielmitterdorfer, line 6: I guess that this was needed during development. Can we remove it now?

noaa/track.py
```diff
@@ -0,0 +1,20 @@
...
+    
+    body = params.get("body")
```
danielmitterdorfer, line 8: The body should be mandatory, thus I propose we retrieve it with `params["body"]` which would fail if the body is not provided.

noaa/track.py
```diff
@@ -0,0 +1,20 @@
...
+    # default page_timeout is not suited for benchmarks as it keeps the scroll contexts open for 45s
+    if 'page_timeout' not in body:
+        body['page_timeout'] = '100ms'
+
+    await es.sql.query(
```
danielmitterdorfer, line 13: Depending on the number of results returned, just calling the API via the client can cause significant enough overhead to skew the results. The overhead is due to parsing the entire response on the client side and converting it to JSON. For that reason we have specific code in Rally to avoid touching the response body in any way. I suggest that we move forward with this implementation but I'd like to ask you to raise an issue for Rally to add SQL as a dedicated runner in Rally which does not parse the response body and then use it in this track in a follow-up PR.

noaa/challenges/sql.json
```diff
@@ -0,0 +1,275 @@
...
+          "#COMMENT": "This is an incredibly short warmup time period but it is necessary to get also measurement samples. As this benchmark is rather about search than indexing this is ok.",
+          "warmup-time-period": 10,
+          "clients": {{bulk_indexing_clients | default(8)}},
+          "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
+          "tags": ["setup"]
+        },
+        {
+          "name": "refresh-after-index",
+          "operation": "refresh",
+          "clients": 1,
```
danielmitterdorfer, line 65: You can specify `clients` but the default value is `1` anyway so you might just as well rely on the default.

danielmitterdorfer (commented):
Thanks! I left a few comments, mostly around the custom runner.

noaa/track.py
Luegg: 👍 I'm also removing it from the `eql` custom operation then.

noaa/track.py
Luegg: That's good to know! The same issue applies for `eql`. So I'll raise it for both operations.

bpintea:
> The baseline helps with a) detecting smaller fluctuations and b) being a shortcut for validating hypothesis.

Not sure how this provides more info compared to the rest of the queries, but it's fine, I guess we can have it, see how it evolves and adjust if needed.

> I can think of a couple of things that are different (e.g. string scripts are probably more memory consuming) and there might be quite some execution paths that differ between the two.

I guess macro-benchmarking is more useful to us&users than determining what we can microoptimise. But not feeling very strong about it.

> I think there are also other ways to improve the dashboards to make detection easier. E.g. we could group all the numeric queries in one panel and only show the 50th percentile.

The 99/100 percentiles in the EQL tests seem to be now on a divergent path, compared to the the other two. We should wait with this narrowing down, at least until we understand what's going on there.

bpintea (approved):
LGTM

Luegg:
> Not sure how this provides more info compared to the rest of the queries, but it's fine, I guess we can have it, see how it evolves and adjust if needed.

👍 that's also what I thought, it's hard to get this right from the beginning anyway.

astefan (approved):
LGTM

noaa/track.py
danielmitterdorfer: Thanks for raising https://github.com/elastic/rally/issues/1382.

danielmitterdorfer (approved):
LGTM

PR #217, opened 2021-11-10
Add vectors search track

mayya-sharipova:
Elasticsearch 8.0 introduces indexed vectors and _knn_search operation
on them. This tracks is for benchmarking indexing, force merging
and _knn_seach operations for indexed vectors.

TODO:

- add deep1b dataset to rally-tracks.elastic.co

- add _knn_search operation to the challenge, once
https://github.com/elastic/rally/issues/1380 is implemented.


vectors_deep1b/_tools/parse.py
```diff
@@ -0,0 +1,20 @@
+import sys
+import struct
+import json
+from tqdm import tqdm
+
+def to_json(f):
+	for i in tqdm(range(num_vectors)) :
```
dliappis, line 7: `tqdm` is not part of the Python standard library. How about if we make the program more lenient and report if not present? I took a stab at a few improvements, e.g. also warning if the source file is missing WDYT?

```python
#!/usr/bin/env python3
import sys
import struct
import json

try:
	from tqdm import tqdm
	iterate = lambda i: tqdm(range(i))
except ModuleNotFoundError:
	print("Warning: [tqdm] package is not available and you won't be able to see progress.", file=sys.stderr)
	iterate = range

dims = 96
num_vectors = 9990000 


def to_json(f):
	for i in iterate(num_vectors):
		f.read(4)  # in .fvecs format for each vector the first 4 bytes represent dim
		vector = struct.unpack('f' * dims, f.read(dims * 4))

		record = {}
		record['vector'] = vector
		print(json.dumps(record, ensure_ascii=False))


if (len(sys.argv) != 2):
	print(f"Error: No .fvecs file. Rerun using [{sys.argv[0]} /path/to/deep10M.fvecs].")
	sys.exit(1)

with open(sys.argv[1], 'rb') as f:
	to_json(f)	
```

dliappis (commented):
Thanks for raising this!

I took a first look, starting at generating the dataset and left a suggestion for making the convert script a bit more user friendly.

As next I'll test the track itself.

mayya-sharipova:
@dliappis Thanks for the feedback on the script, great suggestions! I've push your suggestions in 48e9487215fe.

Also want to report results from my laptop while running:
```bash
./rally race --track-path=.../rally-tracks/vectors_deep1b/ --car="4gheap" --pipeline=from-sources --revision=current --user-tag="intention:vectors1"  --track-params="bulk_indexing_clients:1,ingest_percentage:10"
```
<details>
  <summary>Results</summary>
  
  ```txt
|                                                         Metric |         Task |       Value |   Unit |
|---------------------------------------------------------------:|-------------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |              |     1.11228 |    min |
|             Min cumulative indexing time across primary shards |              |   0.0116833 |    min |
|          Median cumulative indexing time across primary shards |              |    0.556142 |    min |
|             Max cumulative indexing time across primary shards |              |      1.1006 |    min |
|            Cumulative indexing throttle time of primary shards |              |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |              |           0 |    min |
| Median cumulative indexing throttle time across primary shards |              |           0 |    min |
|    Max cumulative indexing throttle time across primary shards |              |           0 |    min |
|                        Cumulative merge time of primary shards |              |     9.07122 |    min |
|                       Cumulative merge count of primary shards |              |           1 |        |
|                Min cumulative merge time across primary shards |              |           0 |    min |
|             Median cumulative merge time across primary shards |              |     4.53561 |    min |
|                Max cumulative merge time across primary shards |              |     9.07122 |    min |
|               Cumulative merge throttle time of primary shards |              |           0 |    min |
|       Min cumulative merge throttle time across primary shards |              |           0 |    min |
|    Median cumulative merge throttle time across primary shards |              |           0 |    min |
|       Max cumulative merge throttle time across primary shards |              |           0 |    min |
|                      Cumulative refresh time of primary shards |              |     4.45042 |    min |
|                     Cumulative refresh count of primary shards |              |          25 |        |
|              Min cumulative refresh time across primary shards |              |   0.0253167 |    min |
|           Median cumulative refresh time across primary shards |              |     2.22521 |    min |
|              Max cumulative refresh time across primary shards |              |      4.4251 |    min |
|                        Cumulative flush time of primary shards |              |     5.44108 |    min |
|                       Cumulative flush count of primary shards |              |           7 |        |
|                Min cumulative flush time across primary shards |              |   0.0421167 |    min |
|             Median cumulative flush time across primary shards |              |     2.72054 |    min |
|                Max cumulative flush time across primary shards |              |     5.39897 |    min |
|                                        Total Young Gen GC time |              |       2.202 |      s |
|                                       Total Young Gen GC count |              |         118 |        |
|                                          Total Old Gen GC time |              |           0 |      s |
|                                         Total Old Gen GC count |              |           0 |        |
|                                                     Store size |              |     1.56353 |     GB |
|                                                  Translog size |              | 1.02445e-07 |     GB |
|                                         Heap used for segments |              |           0 |     MB |
|                                       Heap used for doc values |              |           0 |     MB |
|                                            Heap used for terms |              |           0 |     MB |
|                                            Heap used for norms |              |           0 |     MB |
|                                           Heap used for points |              |           0 |     MB |
|                                    Heap used for stored fields |              |           0 |     MB |
|                                                  Segment count |              |           8 |        |
|                                                 Min Throughput | index-append |     6753.77 | docs/s |
|                                                Mean Throughput | index-append |     10894.2 | docs/s |
|                                              Median Throughput | index-append |     11498.8 | docs/s |
|                                                 Max Throughput | index-append |     11829.6 | docs/s |
|                                        50th percentile latency | index-append |     391.876 |     ms |
|                                        90th percentile latency | index-append |     451.808 |     ms |
|                                        99th percentile latency | index-append |      624.93 |     ms |
|                                       100th percentile latency | index-append |     709.739 |     ms |
|                                   50th percentile service time | index-append |     391.876 |     ms |
|                                   90th percentile service time | index-append |     451.808 |     ms |
|                                   99th percentile service time | index-append |      624.93 |     ms |
|                                  100th percentile service time | index-append |     709.739 |     ms |
|                                                     error rate | index-append |           0 |      % |


---------------------------------
[INFO] SUCCESS (took 952 seconds)
---------------------------------
  ```
</details>

vectors_deep1b/challenges/default.json
```diff
@@ -0,0 +1,55 @@
...
+        "operation-type": "cluster-health",
+        "request-params": {
+          "wait_for_status": "green"
+        },
+        "retry-until-success": true
+      }
+    },
+    {
+      "operation": "index-append",
+      "warmup-time-period": 0,
```
dliappis, line 27: Shouldn't we have some warmup time so that indexing results (esp. max / min) don't get skewed unreasonably due to jit warmup? Typically we've used something like 240s (https://github.com/elastic/rally-tracks/blob/fe23f98a77115d7762bf43a66053cc18d9b1decc/nyc_taxis/challenges/default.json#L33), which is usually derived by looking at the metrics store.

vectors_deep1b/operations/default.json
```diff
@@ -0,0 +1,19 @@
+{       
+  "name": "index-append",
+  "operation-type": "bulk",
+  "bulk-size": {{bulk_size | default(5000)}},
+  "ingest-percentage": {{ingest_percentage | default(100)}}
+},
+{
+  "name": "knn_search",
```
dliappis, line 8: I don't see this operation referenced anywhere in the challenge. Do you intend to add it later?

vectors_deep1b/index.json
```diff
@@ -0,0 +1,24 @@
+{
+  "settings": {
+    "index.number_of_replicas": 0
```
dliappis, line 3: By not specifying the num of primary shards, the ES default of `1` will be used. Just wanted to clarify this is what you want.

vectors_deep1b/challenges/default.json
```diff
@@ -0,0 +1,55 @@
...
+      "clients": {{bulk_indexing_clients | default(8)}}
+    },
+    {
```
dliappis, line 30: See the comment below regarding the `_knn_search` operation; I assume the operations below (force-merge etc.) are intended for use when the `_knn_search` operation gets added, but for now, for indexing only they are not needed right?

dliappis (commented):
Thanks for iterating @mayya-sharipova ! I've left a few questions, the main one is whether we should have a warmup time for indexing as per benchmarking best practices.

vectors_deep1b/challenges/default.json
mayya-sharipova: @dliappis  Thanks for the comment,  happy to add `warmup-time-period`, andI will follow your recommendation. about the correct number. 

Does `240` means that we start collecting `index-append` metrics only after 240s? 
Looking at the Median Throughput of   11000 docs/s,  it means that we start collecting metrics after indexing 2'640'000 docs, or around 25% dataset?   Does this look right to you or should it be smaller?

Also, I have noticed that if I ran `rally` with a smaller ingest_percentage, e.g. `ingest_percentage:10", I may not get any metrics in this case?

vectors_deep1b/challenges/default.json
mayya-sharipova: We have noticed that while indexing knn vectors, `force-merge` takes substantial time, and we are working on optimizing it, so we would like to start tracking it right now, even before running `_knn_search`.

vectors_deep1b/index.json
mayya-sharipova: Yes, for now having 1 shard is good.

vectors_deep1b/operations/default.json
mayya-sharipova: Yes, we want to add this operation as soon as it becomes [available in rally](https://github.com/elastic/rally/issues/1380) .   `_knn_search`  is a new API in elasticsearch, and has not been implemented in rally yet. Speaking of which, would it be possible for the benchmarking team to implement it, or should our search team do it?

mayya-sharipova:
@dliappis Thank you for another round of review. I've added some clarification questions.

vectors_deep1b/challenges/default.json
dliappis: > Does `240` means that we start collecting `index-append` metrics only after 240s?

@mayya-sharipova Correct. The warmup metrics are also available in the metrics store (`sample-type`: `warmup` vs `normal`).

> Also, I have noticed that if I ran `rally` with a smaller ingest_percentage, e.g. `ingest_percentage:10", I may not get any metrics in this case?

Right and while it may feel inconvenient, it actually protects the user from relying on performance figures that can't be 100% trusted.

A decent workaround is to use variable with a default value e.g.

`"warmup-up-period": {{ bulk_warmup | default(240) | int }}`

and thus if you run in a different env to the nightlies, you can override with `--track-params='bulk_warmup:0'`.

Regarding the actual default value, it's just what's used in other tracks, ideally we can run on the actual nightly hardware and using a simple visualization see if 240s is needed for warmup, or less.

vectors_deep1b/operations/default.json
dliappis: > Yes, we want to add this operation as soon as it becomes [available in rally](https://github.com/elastic/rally/issues/1380) . `_knn_search` is a new API in elasticsearch, and has not been implemented in rally yet. Speaking of which, would it be possible for the benchmarking team to implement it, or should our search team do it?

@mayya-sharipova Thanks for the link! I see in the docs that this is an experimental API and might change. If that's true, do you think it makes more sense to use [raw-request](https://esrally.readthedocs.io/en/latest/track.html?highlight=raw-request#raw-request) or a [custom runner](https://esrally.readthedocs.io/en/latest/adding_tracks.html?highlight=custom%20runner#custom-runners) to issue this request?

mayya-sharipova: Great question, considering that it is an experimental, and in the future we may move it under `_search`, I guess it makes sense to have a raw request for now.

vectors_deep1b/challenges/default.json
mayya-sharipova: @dliappis Thanks for the explanation, it makes sense. I've added modified the warmup period as suggested in 8667d92cca5d56ca3706fc64a8668bf19731aa59

mayya-sharipova:
@dliappis Thanks for another round of review. I am going for vacation,  and will be away for some time.  @jtibshirani can the best point of contact. 

@jtibshirani If you have time, feel free to add modifications to this PR.

jtibshirani:
I pushed some changes:
* Update the README to use a different link (still for the same dataset). This link is explicit about the license (CC-BY-4.0). I parsed the dataset and uploaded it to the rally-tracks GCS bucket.
* Add `_knn_search` operations and a `script_score` query that uses vector functions.
* Update the index settings to use 2 shards. This lets us exercise the result merging and fetching logic.

Notes: 
* On a local test, indexing and force-merging one shard (5 million out of 10 million total vectors) took ~2 hours. So the benchmark will take quite a while to complete!
* For now I just randomly chose 3 vectors from the dataset to use as query vectors. Ideally we would use a small "test set" containing many different vectors. This would be a good follow-up, but I wanted to keep the initial PR simple.
* Since kNN search uses an approximate algorithm, it's important to measure result accuracy in addition to latency/ throughput. This helps ensure we're using realistic parameters in benchmarks. We don't measure or report accuracy right now, but we should look into that in a follow-up.

dense_vector/challenges/default.json
```diff
@@ -0,0 +1,83 @@
...
+        "retry-until-success": true,
+        "include-in-reporting": false
+      }
+    },
+    {
+      "name": "knn-search-10-500",
+      "operation": "knn-search-10-500",
+      "warmup-iterations": 100,
+      "iterations": 100,
+      "clients": 4
```
jtibshirani, line 59: I wasn't sure what values to select here, feedback is very welcome.

dliappis: To answer this, first let's clarify if you deliberately want to run this as fast as possible; I don't see `target-throughput` specified, so it will do that. If this is the intended behavior, what value to choose here is a matter of what we want to benchmark. In this case it makes sense to keep this to `1` to avoid saturating the target.

One the other hand if you'd rather use a `target-throughput` we should run some tests on the target hardware to calculate the optimal target throughput and probably also specify a [different schedule](https://esrally.readthedocs.io/en/latest/track.html?highlight=schedule#schedule). In this case, we could also use >1 clients; in this case each client will run requests with a rate of `target throughput/clients`.

dliappis (commented):
@jtibshirani 

Thank you for iterating and improving the track so much!

> On a local test, indexing and force-merging one shard (5 million out of 10 million total vectors) took ~2 hours. So the benchmark will take quite a while to complete!

2hrs is quite a lot of time in nightly benchmarks. Would it make sense that for the nightlies we reduce the amount of data ingested via the track parameter `ingest_percentage`?

If we down that path, it may make sense to tweak the warmup and query iterations based on that.

dense_vector/challenges/default.json
jtibshirani: We discussed offline and agreed it would make sense to set `clients: 1` with no specific `target-throughput`. This lets us run the track as fast as possible, avoiding unnecessary complexity. We don't have a specific reason to test request parallelism too. The rally team is thinking of moving most "standard" tracks to this simple set-up.

jtibshirani:
> Would it make sense that for the nightlies we reduce the amount of data ingested via the track parameter ingest_percentage?

We can definitely do this. I would suggest `ingest_percentage: 20`, corresponding to 2 million total docs, or roughly 1 million per shard. I will test this locally and update the number of iterations.

dense_vector/challenges/default.json
```diff
@@ -0,0 +1,83 @@
...
+        },
+        "retry-until-success": true,
+        "include-in-reporting": false
+      }
+    },
+    {
+      "name": "knn-search-10-500",
+      "operation": "knn-search-10-500",
+      "warmup-iterations": 100,
+      "iterations": 100,
```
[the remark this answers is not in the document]
jtibshirani: This may be an obvious question, but could you explain why the iterations would scale with ingest percentage? It's not clear to me that there'd be a relationship.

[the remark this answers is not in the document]
dliappis: > This may be an obvious question, but could you explain why the iterations would scale with ingest percentage? It's not clear to me that there'd be a relationship.

I don't think it's an obvious question, so it's good we are discussing it. As I mentioned earlier it _could_ (but doesn't have to). As you mentioned in [your comment](https://github.com/elastic/rally-tracks/pull/217#issuecomment-989041810):

> I would suggest ingest_percentage: 20, corresponding to 2 million total docs, or roughly 1 million per shard. I will test this locally and update the number of iterations.

So if we update amount of data we ingest, it _might_ make sense to update the number of iterations (and warmup iterations?), and maybe there is a simple formula we could use.
However, I'd caution that evaluating the right amount of warmup and normal iterations depends on the hardware. Since we intend to run these on the [nightly hardware](https://elasticsearch-benchmarks.elastic.co/#), I suggest we make some measurements on similar hardware. Let's sync offline on how to do that.

rjurney:
It looks like this is the final issue holding up ANN support in Elastic? https://github.com/elastic/elasticsearch/issues/78473

jtibshirani:
Hello @rjurney, we already merged support for ANN in Elasticsearch. You can actually try it out as part of an 8.0 early access build (https://www.elastic.co/downloads/past-releases/elasticsearch-8-0-0-beta1). This PR adds a set of benchmarks (to a separate repo) to help us identify performance issues/ improvements. The results look pretty good so far, and we are on track to ship what we've merged.

jtibshirani:
I ran some experiments on the nightly benchmark hardware to select better warm-up periods. Here are some screenshots showing the warm-up cutoffs for `index-append` and `knn-search-10-500` (the other search operations are very similar):

Screen Shot 2021-12-14 at 6 57 15 PM
Screen Shot 2021-12-14 at 6 20 39 PM

I ran into some difficulties with the `index-append` operation. Setting a generous client timeout like 240s and using `--on-error=abort` showed that the bulk index operations sometimes stall and timeout. I managed to find a very stable configuration using `bulk_indexing_clients: 1` and a 4GB heap that does not trigger timeouts.

These timeouts really need to be investigated, in case there is a problem with dense vector indexing. I would prefer to merge the track with this working configuration, then follow-up with an investigation and fix in Elasticsearch.

dense_vector/index.json
```diff
@@ -0,0 +1,28 @@
+{
+  "settings": {
+    "index": {
+      "number_of_shards": 2,
+      "number_of_replicas": 0,
+      "refresh_interval": -1
```
dliappis, line 6: I am wondering, why do we need this as we aren't doing concurrent indexing and querying? My understanding is that starting with 7.3.0 there's no refresh if indices haven't received a search request within the last 30s, so there shouldn't be any refreshing in the bulk indexing task.

dliappis (commented):
Thank you for iterating here and tuning warmup iterations and indexing clients.

We are basically ready to merge, I left a comment with a question about the need to explicitly disable refreshes. I also wanted to clarify: we intent to run this on the nightly setup with `20` bulk ingest percentage right? This is defined elsewhere, so no need to do anything here, just wanted to understand what we want to do later on for nightly runs.

dense_vector/index.json
jtibshirani: Sorry about that, I did not mean to commit this. I reverted the change.

dliappis (approved):
LGTM! Thank you both @jtibshirani and @mayya-sharipova for the work here.

jtibshirani:
>  I also wanted to clarify: we intent to run this on the nightly setup with 20 bulk ingest percentage right?

That's correct. Here are the notable parameters I used in test runs: `--car=4gheap --client-options="timeout:240" --track-params="ingest_percentage:20"`.

dliappis:
> That's correct. Here are the notable parameters I used in test runs: `--car=4gheap --client-options="timeout:240" --track-params="ingest_percentage:20"`.

That's perfect. We'll deal with passing the ingest_percentage parameter and car in another repo as it's specific to nightlies.

For now feel free to (squash please) merge this!

I don't know if you've considered backporting this to any branch so that it can be run against released versions of ES.
There is branch logic in Rally (described in depth here: https://esrally.readthedocs.io/en/latest/track.html?highlight=cherry%20pick#custom-track-repositories) regarding which branch of rally-tracks will be used depending on the ES version is benchmarking against.

In short the most recent branch in the 7 series in this repo is `7.14` so if you'd backport to `7.14` the particular branch would be picked when ES target version is >=7.14.0 and <8.

PR #224, opened 2021-12-06
EQL: Adjust new benchmarks

Luegg:
Some of the new EQL benchmark tasks require adjustment because the target throughput is on the verge of causing congestion. As a result, the latency piles up and causes stark outliers in the aggregated latency stats. Visible in the following two graphs:

image
image


bpintea (approved):
Lgtm.

PR #229, opened 2021-12-13
EQL: Additional challenge using index sorting

Luegg:
Because most EQL queries result in one or more search requests with a query ordered by timestamp (descending), index sorting might be a sensible option if EQL is used heavily. Some initial tests revealed that the performance of most queries  increases by 20 - 50% with index sorting.

This PR adds another challenge to the EQL track that runs all of the new queries against a sorted index.

The new challenge should help to give informed advice to EQL users about the costs and benefits of using index sorting. It allows to compare both individual query performance as well as bulk indexing rates between the two configurations.

In order to share the queries between the two challenges, they have been extracted into `queries.json` and `legacy_queries.json` but have otherwise not been touched.


bpintea (approved):
I'm only wondering if `default` and `index_sorting` couldn't be composed to avoid dups, with the exception of index settings and included queries being the same. But otherwise LGTM

costin (approved):
LGTM

astefan (approved):
LGTM

b-deam (approved):
LGTM.

Luegg:
> I'm only wondering if `default` and `index_sorting` couldn't be composed to avoid dups, with the exception of index settings and included queries being the same. But otherwise LGTM

I was thinking the same thing but the abstractions in Jinja are a bit tricky to work with. In the end I thought this is the clearest version of the `/challenges` files.

PR #233, opened 2022-01-28
Adjust the parameters for dense vector operations

jtibshirani:
Previously these were set quite high, producing a recall above 99%. This change
bumps them down to more realistic values that still maintain recall over 90% I
tested recall over example queries using an external framework (ann-benchmarks)
since rally doesn't currently support this.


DJRickyB (approved):
LGTM

mayya-sharipova (approved):
LGTM as well

PR #237, opened 2022-02-14
index-append-with-timestamp-pipeline: set ignore non-fatal errors by default

inqueue:
Set `ignore-response-error-level: non-fatal` by default for task `index-append-with-timestamp-pipeline` to prevent aborting track execution on a non-fatal error. The same was implemented previously for `index-append` in https://github.com/elastic/rally-tracks/pull/206. Both fixes address errors we've encountered with http_logs encryption-at-rest benchmarks.

### DIFF
--- a/http_logs/challenges/common/default-schedule.json
+++ b/http_logs/challenges/common/default-schedule.json
@@ -26,7 +26,8 @@
         {
           "operation": "index-append-with-timestamp-pipeline",
           "warmup-time-period": 240,
-          "clients": {{bulk_indexing_clients | default(8)}}
+          "clients": {{bulk_indexing_clients | default(8)}},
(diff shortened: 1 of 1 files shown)


DJRickyB (approved):
LGTM

inqueue:
Backported to:

* 7.14 https://github.com/elastic/rally-tracks/commit/d103872233f7c79b12a17f2771d9369d4cd6d3dd
* 7.12 https://github.com/elastic/rally-tracks/commit/c68c8d3c4453c083ef6954a4d77e43366fceac13
* 7.2 https://github.com/elastic/rally-tracks/commit/974f77fd99bca4517b8b902a72cd5eeb66037e1c

PR #239, opened 2022-02-21
Run NYC taxis searches without target throughput too

pquentin:
This is part of an experiment that might end up with target-throughput removal across all tracks. This would be beneficial to avoid duplication for ARM benchmarks, which is why we're starting with NYC Taxis.

Tested with `esrally race --distribution-version=7.16.3 --track-path=~/src/rally-tracks/nyc_taxis --challenge=append-no-conflicts --test-mode`.


dliappis (approved):
LGTM!

Curious to see the results and where this experiment takes us.

PR #242, opened 2022-03-01
Use native SQL operation type and add paging queries

Luegg:
Replaces the custom `sql` operation type with the one introduced in https://github.com/elastic/rally/pull/1438 and adds new queries using the `pages` option. The new pagination queries cover all types of cursors.

The existing queries are unchanged and the impact on the results should be very small but we might see some decrease in latency. Mostly because the native operation type uses a streaming parser to extract the relevant parts of the response instead of parsing the whole response body.

CC @costin @astefan @bpintea @luigidellaquila

### DIFF
--- a/sql/README.md
+++ b/sql/README.md
@@ -13,6 +13,10 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `max_num_segments` (default: 1)
 * `query_percentage` (default: 100): Factor applied to the number of warmup-iterations and iterations for queries. Useful to run quick experiments but watch out for effects due to the shorter warmup period!
 
+## Testing
+
+Because some of the pagination queries expect more than 1k records to work, `--test-mode` is not supported for this track. Instead, the `ingest_percentage` and `query_percentage` track parameters can be used to test the track: `--track-params="ingest_percentage:1,query_percentage:2"`
+
 ## Query Selection
 
 The queries have been selected using the following criterias:
--- a/sql/track.json
+++ b/sql/track.json
@@ -87,7 +87,7 @@
   ],
   "corpora": [
     {
-      "name": "noaa",
+      "name": "noaa-sql",
       "base-url": "https://rally-tracks.elastic.co/noaa-sql",
       "documents": [
         {
--- a/sql/track.json
+++ b/sql/track.json
@@ -4,7 +4,7 @@
 {{ (iterations * (query_percentage | default(100)) / 100) | int }}
 {%- endmacro %}
 
-{% macro query_10qps(name, query) -%}
+{% macro query_10qps(name, query, pages=1, fetch_size=1000) -%}
 {
   "operation": {
     "name": "{{ name }}",
@@ -15,8 +15,10 @@
         Because ES < 7.16 uses `request_timeout` as `page_timeout` the values need to be identical but high enough such that
         the requests do not time out. #}
       "page_timeout": "1000ms",
-      "request_timeout": "1000ms"
-    }
+      "request_timeout": "1000ms",
+      "fetch_size": {{ fetch_size }}
+    },
+    "pages": {{ pages }}
   },
   "target-throughput": 10,
   "clients": 2,
@@ -26,7 +28,7 @@
 }
 {%- endmacro %}
 
-{% macro query_5qps(name, query) -%}
+{% macro query_5qps(name, query, pages=1, fetch_size=1000) -%}
 {
   "operation": {
     "name": "{{ name }}",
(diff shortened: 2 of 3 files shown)


Luegg:
As mentioned in https://github.com/elastic/rally/pull/1438#issuecomment-1049933540, this PR can only be backported once  https://github.com/elastic/rally/pull/1438 has been released.

michaelbaamonde:
I'm getting errors for all of the operations that attempt to fetch 5 pages when running in `test-mode`. Example:

```
esrally.exceptions.RallyError: Cannot run task [select_5pages]: Result set has been exhausted before all pages have been fetched, 4 page(s) remaining.
```
I have a hunch that this may be due to the truncation of the data set with `test-mode`, so I'm doing a full run to test that hypothesis.

michaelbaamonde:
@Luegg My local testing with the full data set (took a while... :smile:) also confirmed that `test-mode` is the cause. We've recently had some discussions within the team about some issues with `test-mode` in general, and this is one of them. Making it more robust is something that's on my radar, but no firm plans yet.

Your analysis of the available options lays things out nicely. I think there's two viable paths forward, both of which you touch on.

1. Don't [throw an exception](https://github.com/elastic/rally/pull/1438/files#diff-08281b42148c105b0855d07b562aaee68f560e448765fcf9c925b9163b16bfc5R2469) at all if the result set is exhausted before the specified number of pages is reached. I wouldn't want to try to handle `test-mode` differently within runners themselves, so it's an all-or-nothing kind of thing.
2. I'll just quote you here since you put it so well: Accept that test-mode does not work for SQL and add a note in the README.md on how to test it with ingest/query-percentage.

Option 1 is something I should have considered in the Rally PR. Apologies for that oversight. We do support pagination in the `Query` runner (for general searches), which does not treat early exhaustion as an error condition (see [this docstring](https://github.com/elastic/rally/blob/91634e9dbc57a277bce43326549aaad705233f67/esrally/driver/runner.py#L798).) This is an interesting question in general. If the user specifies a certain number of pages, should we interpret this as an invariant for the workload? I can see the argument for that, especially with a static corpus. On the other hand, we do offer [assertions](https://esrally.readthedocs.io/en/stable/track.html#operations), which can be used to check that a search returned a certain number of hits, treating a failure to do so as an error. Maybe that's a better way to do this, but... those also don't play nicely with `test-mode` at the moment :upside_down_face: , and we'd (I believe) also need to enhance the `sql` runner to optionally return detailed results.

I'm slightly leaning toward Option 1 for consistency's sake, but Option 2 isn't a bad one in the short term. In either case, we should add some clarifications to the Rally docs about `test-mode`'s current limitations. I'd like to hear @DJRickyB's thoughts on this since he's worked on pagination support in the past and may have more context on the considerations I mentioned re: Option 1.

DJRickyB:
In other operations, we handle pagination through the client by exercising logic using the number of hits retrieved and the number of pages expected. Using the error here for flow-control makes sense if that's the desired behavior but it's a departure from what we've done before and so I'm inclined to say leave it be.

**To disable test-mode capability for your track, just delete the `-1k.json` version of the corpora file** hosted publicly. Users will get an error saying this track doesn't support test mode. I think that's the most straightforward approach, especially given @Luegg:

>  I'd like to keep the check for having all pages fetched in prod mode

I don't think we should tweak the queries, or the runner, to be compatible with test-mode. It would be wonky and counter-productive

michaelbaamonde:
I'm good with that approach. A note in the README about how to test the track with the `ingest_percentage` and `query_percentage` params would be nice, as @Luegg suggested.

There's a lot we can do with `test-mode` going forward, and this PR has given me some more food for thought!

michaelbaamonde (approved):
LGTM. A note about how to test the track in the README would be nice, but not a blocker for me to approve. Thanks for all the work on this @Luegg!

Luegg:
Thanks for the considerations @michaelbaamonde and @DJRickyB. I've added a note to the readme and also ensured that the corpora name does no longer clash with the one from the noaa track. Up until now they shared the same path in the cache folder `~/.rally/benchmarks/data/`.

I'm not sure I have write access to delete the `-1k` file. Can one of you take this over?

Assertions would have certainly helped as well to achieve the semantics I was looking for. I think the runner also already exposes the `weight` property that corresponds to the number of pages. Since it should be save to change the error semantics in a non-breaking way in the future, I'll go forward with the current approach for now.

michaelbaamonde:
> Can one of you take this over?

Sure thing. Thanks again!

PR #243, opened 2022-03-02
refactor: conditionally include aggs tasks to the default challenge

salvatore-campagna:


noaa/challenges/default.json
```diff
@@ -356,205 +358,4 @@
           "target-throughput": 1
         }
       ]
-    },
-    {
-      "name": "aggs",
```
DJRickyB, line 361: we should keep this challenge, at least for transitional purposes. it can just unconditionally `include 'aggs-tasks.json'`

noaa/challenges/default.json
```diff
@@ -107,7 +107,9 @@
           "warmup-iterations": 100,
           "iterations": 500,
           "target-throughput": 5
-        }
+        }{%- if include_aggs is defined and include_aggs | default(false) %},
```
DJRickyB, line 110: this can be simplified, i think
```suggestion
        }{%- if include_aggs | default(false) %},
```

DJRickyB (changes requested):

noaa/challenges/default.json
dliappis, line 110: We could be a bit more resilient (and cover both strings and booleans). There are a couple of ways to pass track params in Rally and if you don't pass a json file (or inline json), the KV type definitions could be error prone for boolean type values. In practice, the conditional you have here, evaluates as `True` is there is a a typo like `esrally ... --track-params="include_aggs:FAlse".

Unfortunate standard jinja2 doesn't include the handy `| bool` filter that Ansible contains, but you could instead:

`{% if include_aggs | default(False) | string | lower == "true" %}`

Otherwise, if we want to be string about the datatype being a boolean, Rick's simplification above should work.

dliappis (commented):
Left a few comments

dliappis (approved):
LGTM thank you!

dliappis:
@salvatore-campagna note that we have a policy not to merge such changes before a weekend, to avoid possible breakage of nightly benchmarks. Let's merge on Monday.

salvatore-campagna:
@DJRickyB I applied the changes you requested. Can I proceed with merging them?

PR #244, opened 2022-03-02
Run `no_target` versions immediately before original ones

pquentin:
Tested with `esrally race --distribution-version=8.0.0 --track-path=~/src/rally-tracks/nyc_taxis --challenge=append-no-conflicts --test-mode`


DJRickyB (approved):
LGTM! Please change the title (and squash commit message, if necessary) before merging

PR #246, opened 2022-03-02
TSDB: Add an option to disable _source

nik9000:
This is a fairly standard option and we were looking into this in tsdb-land. So the option is useful to have.

### DIFF
--- a/tsdb/README.md
+++ b/tsdb/README.md
@@ -121,6 +121,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `number_of_replicas` (default: 0)
 * `number_of_shards` (default: 1)
 * `force_merge_max_num_segments` (default: unset): An integer specifying the max amount of segments the force-merge operation should use.
+* `source_enabled` (default: true): A boolean defining whether the `_source` field is stored in the index.
 * `index_mode` (default: time_series): Whether to make a standard index (`standard`) or time series index (`time_series`)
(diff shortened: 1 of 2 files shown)


tsdb/README.md
```diff
@@ -121,6 +121,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `number_of_replicas` (default: 0)
 * `number_of_shards` (default: 1)
 * `force_merge_max_num_segments` (default: unset): An integer specifying the max amount of segments the force-merge operation should use.
+* `source_enabled` (default: true): A boolean defining whether the `_source` field is stored in the index.
```
pquentin, line 124: Have you considered calling this `source_stored`?

pquentin (commented):
This looks good, however looking at the "store size" in test mode I see that disabling source takes more space. Trying with the full data now with the disk-usage-stats telemetry device

nik9000:
Yeah. There is this magic `recovery_source` made now. It gets dropped after
a while, but sticks around too long for the benchmark.

I personally have plans for recover_source.

nik9000:
Also! I can call the parameters whatever you like. Other tracks have this
parameter too. I should have copied the name. Consistency is really the
only argument I have for any particular name. But if you like another name
I'm cool with it.

pquentin (approved):
Thanks! I'll trust you that it works, lost patience with the full version (and learned about `ingest-percentage` too late).

Agreed that we should use a consistent name

PR #247, opened 2022-03-02
TSDB: sorted ingest

nik9000:
This adds and `ingest_order:sorted` option to the tsdb track. We believe
that in production data is quite likely to be ingested in mostly-sorted
order. And we were testing with unsorted data. This provides the option
to test with fully-sorted data. That's not *exactly* what we expect in
production, but it's much closer than the other option. In a future
follow up we'll add the option to test with mostly-sorted data.


pquentin (approved):
Thanks!

PR #249, opened 2022-03-04
Revert "Use native SQL operation type and add paging queries"

Luegg:
Reverts elastic/rally-tracks#242

At least one of the new queries (`aggregate_5pages `) fails with the complete dataset on the nightly with a `[{'type': 'query_phase_execution_exception', 'reason': 'Time exceeded'}]` error.


DJRickyB:
LGTM. Hopefully when this comes back we get back the corpora name, the macro, and the README entry

DJRickyB (approved):

Luegg:
Thanks 👍 Yes, these are all nice improvements I'd like to keep in the track. Except for the corpora name, the changes only make sense combined with the new pagination queries though.

PR #251, opened 2022-03-14
TSDB: Fix "duplicates"

nik9000:
Now there we're calculating the `_id` from dimensions I found that we
have duplicated `_id`s in our test data. But it's not because we
actually measure things three time. No! We just handn't annotated all of
the dimensions as `time_series_dimension`s.

### DIFF
--- a/tsdb/index.json
+++ b/tsdb/index.json
@@ -9,6 +9,8 @@
       {% if index_mode | default('time_series') is equalto 'time_series' %}
       "routing_path": [
         "metricset.name",
+        "host.name",
+        "kubernetes.container.name",
         "kubernetes.event.involved_object.uid",
         "kubernetes.node.name",
         "kubernetes.system.container.id",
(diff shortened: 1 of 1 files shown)


nik9000:
> But it's not because we
> actually measure things three time. No! We just handn't annotated all of
> the dimensions as `time_series_dimension`s.

I'm pretty sure we do have duplicates. In the `event` data set each metricbeat logs the event.

nik9000:
This doesn't remove all duplicates. But it does remove *some* duplicates. Which is lovely.

tsdb/index.json
```diff
@@ -7816,7 +7818,7 @@
               },
               "name": {
                 "type": "keyword",
-                "ignore_above": 1024
+                "time_series_dimension": true
```
imotov, line 7821: Nice! 😄

imotov (approved):
LGTM

PR #252, opened 2022-03-15
Adjust throughputs for some EQL queries after Lucene upgrade

Luegg:
After merging https://github.com/elastic/elasticsearch/pull/84540 there was a spectacular spike in latency for the 3 EQL queries `sequence_2stages_baseline`, `sequence_2stages_tailThenHead` and `sequence_2stages_maxspan1m`. A big part of the regression can be explained by congestion effects caused by a smaller, but still large, increase in service_time.

E.g. typical service times for a single run of `sequence_2stages_baseline` looked like this before:
image
Now, we see a 3-4 fold increase:
image

This PR adjusts the target throughput such that the congestion effect no longer occurs.

### DIFF
--- a/eql/queries.json
+++ b/eql/queries.json
@@ -90,7 +90,7 @@
         "query": "sequence [any where true] [any where true]"
       }
     },
-    {{ task_options_10qps }}
+    {{ task_options_1qps }}
   },
   {
     "operation": {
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM thanks

PR #253, opened 2022-03-15
[7.12] Adjust throughputs for some EQL queries after Lucene upgrade (#252)

Luegg:
Backports the following commits to 7.12:
 - Adjust throughputs for some EQL queries after Lucene upgrade (#252)


pquentin:
FYI You can push backports directly, you don't need pull requests.

PR #255, opened 2022-03-15
Fix error in tsdb track

nik9000:
When I fixed duplicates I left a non-dimension field in `routing_path`.
I didn't test it. Big oops. Now the track doesn't run because that's an
invalid index configuration. This fixes that.


imotov (approved):
LGTM

PR #256, opened 2022-03-17
Revert "Adjust throughputs for some EQL queries after Lucene upgrade"

Luegg:
With the merge of https://github.com/elastic/elasticsearch/pull/85025 the affected queries should again perform as before (or even slightly better).

Reverting the throughputs gives:
- faster benchmarks
- continuous metrics for the affected queries

### DIFF
--- a/eql/queries.json
+++ b/eql/queries.json
@@ -90,7 +90,7 @@
         "query": "sequence [any where true] [any where true]"
       }
     },
-    {{ task_options_1qps }}
+    {{ task_options_10qps }}
   },
   {
     "operation": {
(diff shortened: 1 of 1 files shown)


Luegg:
Currently we see a two step latency improvement in the nightlies:

image

(showing sequence_2stages_baseline with the extreme values cut off)

The first improvement on the 15th was due to the throughput adjustments, the second one the day after due to the fix. The latency went 30% below the level before the regression which is not expected and most likely can be attributed to the throughput adjustments.

dliappis (approved):
LGTM

PR #259, opened 2022-03-17
Use native SQL operation type and add paging queries (2nd attempt)

Luegg:
Second attempt for replacing the custom sql operation type with the new native one introduced in https://github.com/elastic/rally/pull/1438 (see https://github.com/elastic/rally-tracks/pull/242 for the 1st attempt).

This PR also adds new queries using the pages option. The new pagination queries cover all types of cursors.

The existing queries are unchanged and the impact on the results should be very small but we might see some decrease in latency. Mostly because the native operation type uses a streaming parser to extract the relevant parts of the response instead of parsing the whole response body.

I assessed that the queries succeed in reasonable time locally with both 1% and 100% of the documents.


michaelbaamonde (approved):
LGTM

PR #260, opened 2022-03-17
Dense vectors add searches before force merge

mayya-sharipova:
This gives us insights into search performance of
unmerged indices, which we expect to be common
use case for most users.

### DIFF
--- a/dense_vector/challenges/default.json
+++ b/dense_vector/challenges/default.json
@@ -45,6 +45,59 @@
       }
     },
     {
+      "name": "knn-search-10-50-before-force-merge",
+      "operation": "knn-search-10-50",
+      "warmup-iterations": 50,
+      "iterations": 100,
+      "clients": 1
+    },
+    {
+      "name": "knn-search-10-100-before-force-merge",
+      "operation": "knn-search-10-100",
+      "warmup-iterations": 50,
+      "iterations": 100,
+      "clients": 1
+    },
+    {
+      "name": "knn-search-100-1000-before-force-merge",
+      "operation": "knn-search-100-1000",
+      "warmup-iterations": 50,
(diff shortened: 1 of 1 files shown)


mayya-sharipova:
I've received an offline comment from @DJRickyB during the GH outage:

> this will make prior results incomparable. It's a relatively new track so I don't think that's a huge issue, but have you considered instead leaving these names alone and marking the new tasks `-before-force-merge` instead?

@DJRickyB, I've addressed your feedback in 5bf4559906a9c49b3f017ee832044e9cec432795

dense_vector/challenges/default.json
```diff
@@ -27,6 +27,51 @@
       "warmup-time-period": {{ bulk_warmup | default(40) | int }},
       "clients": {{bulk_indexing_clients | default(1)}}
     },
+    {
+      "name": "refresh-after-index",
+      "operation": "refresh"
+    },
+    {
+      "name": "wait-until-merges-finish-after-index",
```
jtibshirani, line 35: I wonder if we can just keep this one, and drop the next one `wait-until-merges-finish`? Since force merge is blocking, not sure we need to wait for merges to finish?

jtibshirani (approved):
This looks good to me too, thanks @mayya-sharipova.

dense_vector/challenges/default.json
mayya-sharipova: @jtibshirani Thanks for your comment.  In the past, we had a situation when merges were still happening in the background even after force merge is done. That's why now we rely on merge stats, details are [here](https://github.com/elastic/rally-tracks/commit/e2292ec2cb723e231bd8e3f4f8840088645a823a).

PR #261, opened 2022-03-17
Fix incorrect iterations in NYC Taxis

DJRickyB:
This change undoes an improperly resolved conflict from a prior PR. This should reduce typical runtimes of the default NYC Taxis challenge by about 15 minutes.


dliappis (approved):
LGTM thanks!

PR #238, opened 2022-02-17
Add new geoshape default track that runs with less data

iverase:
The current track for geoshape uses too much data and it is not fully curated, it is pretty hard to use. Therefore here I added a new default track that runs with less data and it is fully curated. It takes to run around 90 minutes and the output looks like:

```
|                                                         Metric |                          Task |       Value |   Unit |
|---------------------------------------------------------------:|------------------------------:|------------:|-------:|
|                     Cumulative indexing time of primary shards |                               |     61.7476 |    min |
|             Min cumulative indexing time across primary shards |                               |     16.8933 |    min |
|          Median cumulative indexing time across primary shards |                               |     19.2621 |    min |
|             Max cumulative indexing time across primary shards |                               |     25.5922 |    min |
|            Cumulative indexing throttle time of primary shards |                               |           0 |    min |
|    Min cumulative indexing throttle time across primary shards |                               |           0 |    min |
| Median cumulative indexing throttle time across primary shards |                               |           0 |    min |
|    Max cumulative indexing throttle time across primary shards |                               |           0 |    min |
|                        Cumulative merge time of primary shards |                               |     56.8158 |    min |
|                       Cumulative merge count of primary shards |                               |          85 |        |
|                Min cumulative merge time across primary shards |                               |     9.52783 |    min |
|             Median cumulative merge time across primary shards |                               |     18.7066 |    min |
|                Max cumulative merge time across primary shards |                               |     28.5814 |    min |
|               Cumulative merge throttle time of primary shards |                               |       22.86 |    min |
|       Min cumulative merge throttle time across primary shards |                               |      1.3177 |    min |
|    Median cumulative merge throttle time across primary shards |                               |     9.16597 |    min |
|       Max cumulative merge throttle time across primary shards |                               |     12.3763 |    min |
|                      Cumulative refresh time of primary shards |                               |    0.475633 |    min |
|                     Cumulative refresh count of primary shards |                               |          79 |        |
|              Min cumulative refresh time across primary shards |                               |     0.08635 |    min |
|           Median cumulative refresh time across primary shards |                               |    0.100633 |    min |
|              Max cumulative refresh time across primary shards |                               |     0.28865 |    min |
|                        Cumulative flush time of primary shards |                               |     2.21708 |    min |
|                       Cumulative flush count of primary shards |                               |          44 |        |
|                Min cumulative flush time across primary shards |                               |    0.393983 |    min |
|             Median cumulative flush time across primary shards |                               |    0.788533 |    min |
[...]

PR #262, opened 2022-03-18
Add new geoshape default track that runs with less data (#238)

iverase:
Makes this track to run under 1.5 hours.

backport #238


iverase:
@dliappis Does it needs to run any check?

dliappis (approved):
> @dliappis Does it needs to run any check?

LGTM. Normally if the backport/cherry pick is clean, we push straight to the branch

PR #264, opened 2022-03-22
Dense vectors add search before force merge

mayya-sharipova:
This is a follow up on PR #260 that was already approved, merged, 
and reverted, as we did not have `request_timeout` for refresh
operation and it was timing out.

This PR is dependant on https://github.com/elastic/rally/pull/1463

### DIFF
--- a/dense_vector/challenges/default.json
+++ b/dense_vector/challenges/default.json
@@ -45,6 +45,59 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


dliappis (approved):
LGTM, I've also tested it locally.

https://github.com/elastic/rally/pull/1463 has been merged already

PR #266, opened 2022-03-31
delete target-throughputs from nyc_taxis and http_logs

DJRickyB:
This change will make these benchmarks shorter and more platform-agnostic


pquentin (approved):
Thanks! I'm assuming you're leaving `default_no_target` and `default` untouched because we don't know yet why the former is slower?

DJRickyB:
Yes. It's a tentative unchange

PR #273, opened 2022-06-13
Simplify dense_vector track

mayya-sharipova:
- Remove "before-force-merge" operations, as we can get similar
insights from searches concurrent-with-indexing.
- Remove knn-search-10-50 (10-100 and 100-1000 searches give us
enough insights)


jtibshirani (approved):

DJRickyB (approved):
LGTM

PR #169, opened 2021-04-20
Use a new CDN backed bucket for standard tracks

dliappis:
Following the pattern we used with the EQL track (#147) we switch
storage of the corpora of standard tracks to a CDN backed Google Cloud
Storage bucket.

The previous bucket wasn't CDN backed so this should result in a
concrete improvement of the user experience.

PR #274, opened 2022-06-23
Switch to new bucket for offline download script

dliappis:
Since [^1] the standard Rally tracks use a CDN for corpora.
This commit modifies the offline downloader script to use the new
path.

[^1]: https://github.com/elastic/rally-tracks/pull/169

### DIFF
--- a/download.sh
+++ b/download.sh
@@ -8,7 +8,7 @@ set -u
 set -o pipefail
 
 readonly ROOT=".rally/benchmarks"
-readonly URL="http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora"
+readonly URL="http://rally-tracks.elastic.co"
 
 
 # see http://stackoverflow.com/a/246128
@@ -57,7 +57,7 @@ for f in ${FILES}; do
     TARGETS[${#TARGETS[*]}]="${TARGET_PATH}"
     if [ ! -f "${HOME}/${TARGET_PATH}" ]
     then
-        curl -o "${HOME}/${TARGET_PATH}" "${URL}/${TRACK}/${f}"
+        curl -o "${HOME}/${TARGET_PATH}" -L "${URL}/${TRACK}/${f}"
     fi
 done
 
--- a/download.sh
+++ b/download.sh
@@ -8,7 +8,7 @@ set -u
 set -o pipefail
 
 readonly ROOT=".rally/benchmarks"
-readonly URL="http://rally-tracks.elastic.co"
+readonly URL="https://rally-tracks.elastic.co"
 
 
 # see http://stackoverflow.com/a/246128


dliappis:
Relates https://discuss.elastic.co/t/not-able-to-download-track/305712/8

download.sh
```diff
@@ -8,7 +8,7 @@ set -u
 set -o pipefail
 
 readonly ROOT=".rally/benchmarks"
-readonly URL="http://benchmarks.elasticsearch.org.s3.amazonaws.com/corpora"
+readonly URL="http://rally-tracks.elastic.co"
```
pquentin, line 11: This only works with HTTPS for me:

```suggestion
readonly URL="https://rally-tracks.elastic.co"
```

download.sh
```diff
@@ -57,7 +57,7 @@ for f in ${FILES}; do
     TARGETS[${#TARGETS[*]}]="${TARGET_PATH}"
     if [ ! -f "${HOME}/${TARGET_PATH}" ]
     then
-        curl -o "${HOME}/${TARGET_PATH}" "${URL}/${TRACK}/${f}"
+        curl -o "${HOME}/${TARGET_PATH}" -L "${URL}/${TRACK}/${f}"
```
pquentin, line 60: I still don't understand why we'd need this but why not :shrug:

pquentin (changes requested):
Forgot to submit my review, sorry!

download.sh
dliappis: Good catch

download.sh
pquentin: If we want to be even more solid, what are you thoughts about this?

```suggestion
        curl -fsSL o "${HOME}/${TARGET_PATH}" "${URL}/${TRACK}/${f}"
```

fail, silent, show error, redirect

pquentin: Right it's going to be interactive in various situations.

pquentin (approved):
Thanks! LGTM.

dliappis:
Backported for completeness sake (however the [instructions](https://esrally.readthedocs.io/en/stable/offline.html?highlight=offline#using-tracks) only point to the master branch):

| branch | commit |
| ---- | ---- |
| 7.14 | https://github.com/elastic/rally-tracks/commit/ac264dbfd60685d3e21e7b482f13f6c2f235e6ff |
| 7.12 | https://github.com/elastic/rally-tracks/commit/7db9ba2b809a09ec20dc05edc127eb17b30ea2a1 |
| 7.2 | https://github.com/elastic/rally-tracks/commit/49ef13ab200ce8265b5e5f54e60d15139afaac99 |
| 7 | https://github.com/elastic/rally-tracks/commit/a38eb519f04e581de1f6d35a2a5e91775166d583 |
| 6 | https://github.com/elastic/rally-tracks/commit/2951f347fbf13e8deddb69af1c9107ce454bd599 |

PR #275, opened 2022-06-23
Modify tsdb track to allow ingesting into a data stream

martijnvg:
This change adds an `ingest_mode` track option to the tsdb track. If this track option is set to `data_stream`
then this track sets up a composable index templates that allows for ingesting into a tsdb data stream.

If the `ingest_mode`track param  isn't set or set to any other value than `data_stream` then it fall back to ingesting into a tsdb index, which is what the track is doing today. By default the `ingest_mode` track option is not specified so ingesting into a tsdb index is the default behaviour. 

Note that the mappings between `index.json` and `index-template.json` are identical. Only the settings
section differ slightly. In the case for tsdb data stream the following settings get generated upon data stream creation:

- `index.time_series.start_time` - based on current time and the `index.look_ahead_time` index setting
- `index.time_series.end_time` - based on current time and the `index.look_ahead_time` index setting
- `index.routing_path` (based from the mapping, essentially all fields with `time_series_dimension=true` attribute)

Another difference between the data stream and index ingest mode is that the `ingest mode=data_stream` also 
uses a pipeline. This to update the `@timestamp` field of documents to be close to current time. A painless script
is used to compute the number of days between `2021-04-28T17:18:23.410Z` (first timestamp in the data set) and current time. Then the pipeline adds that number of days to the `@timestamp` of documents, this makes the timestamp close to current time and maintains the original timestamp distribution. The data set roughly contains a days worth of metrics.

Also `dedupe.py` tool is added to the `_tools` directory. When ingesting into a tsdb data stream the `op_type=create` is required and so when a document has the same dimension field values and the same timestamp (dimension fields and timestamp are used to create the _id of a tsdb document) then a version conflict occurs (409). This tool removes the duplicates, so that no 409 are retuning when ingesting the data via the bulk api. Note that when indexing into a tsdb data stream the default `op_type=index` is used and then duplicates just get overwritten and Rally doesn't return an errors, so before it wasn't an issue from a benchmark perspective.


martijnvg:
> oh! I thought you'd make this --track-option on the tsdb track.

This is just how started out, because initially the benchmark looked very different (with ilm and all).
I can change this PR to embed the data stream specific bits into the tsdb track and add an track option to ingest
into a data stream.

nik9000:
> I can change this PR to embed the data stream specific bits into the tsdb track and add an track option to ingest
> into a data stream.

That'd be lovely!

martijnvg:
@nik9000 I've added a `ingest_mode` param to tsdb track and inlined the tsdb data stream track into tsdb track: 9c5077e24ccc83ceb40d316b03ae86352776731c

tsdb/_tools/dedupe.py
```diff
@@ -0,0 +1,106 @@
...
+    return dct
+
+def generate_event_key(parsed_line):
+    return parsed_line['kubernetes']['event']['involved_object']['uid']
+
+def generate_state_container_key(parsed_line):
+    key = parsed_line['kubernetes']['container']['name']
+    key += parsed_line['kubernetes']['pod']['name']
+    key += parsed_line['kubernetes']['node']['name']
+    container_id = safeget(parsed_line, 'kubernetes', 'container', 'id')
```
nik9000, line 21: I wonder if `parsed_line.get('kubernetes', []).get('container', []).get('id')` is more normal here. Someone from the perf team will want to review all of this anyway and they should know better than me.

tsdb/track.json
```diff
@@ -3,14 +3,42 @@
...
   "indices": [
     {
       "name": "tsdb",
       "body": "index.json"
     }
   ],
+  {%- endif %}
   "corpora": [
-    {%- if ingest_order is defined and ingest_order == "sorted" %}
+    {%- if ingest_mode is defined and ingest_mode == "data_stream" %}
```
nik9000, line 29: For those following along, we'll likely switch everything to the deduped version in a follow up.

nik9000 (commented):
This probably wants an update to the `README` with instruction on how to run the dedup process.

tsdb/_tools/dedupe.py
martijnvg: this is what I was looking for

martijnvg: pushed: c0e211ee305a15910b32aa2cae252f77dea9e900

martijnvg:
> This probably wants an update to the README with instruction on how to run the dedup process.

Added docs: fc67579b4b9206154747a4fec93d9bd6470b2e9f

nik9000 (approved):
LGTM. I'd wait for a review from one of the perf folks, but I like it.

tsdb/_tools/dedupe.py
```diff
@@ -0,0 +1,111 @@
...
+def generate_container_key(parsed_line):
+    return parsed_line['kubernetes']['container']['name'] + parsed_line['kubernetes']['pod']['name'] + parsed_line['kubernetes']['node']['name']
+
+def generate_volume_key(parsed_line):
+    return parsed_line['kubernetes']['volume']['name'] + parsed_line['kubernetes']['pod']['name'] + parsed_line['kubernetes']['node']['name']
+
+def generate_pod_key(parsed_line):
+    return parsed_line['kubernetes']['pod']['name'] + parsed_line['kubernetes']['node']['name']
+
+def generate_node_key(parsed_line):
```
michaelbaamonde, line 43: A minor thing, but this function and `generate_state_node_key` are identical. There's also several places where you could replace `parsed_line['kubernetes']['node']['name']` with `generate_state_node_key(parsed_line)` just to cut down on duplication if you'd like. Not a big deal, though.

michaelbaamonde (changes requested):
Thanks for putting this together @martijnvg and @nik9000! This all worked for me locally and generally looks good. If you could fix the typo and indentation in the README, we'll be good to merge. I did leave a minor suggestion regarding the dedupe script that you can take or leave. Not a big deal.

martijnvg:
Thanks @michaelbaamonde for reviewing! I made the suggested changes.

michaelbaamonde (approved):
LGTM, thanks @martijnvg!

PR #283, opened 2022-07-13
Test some tracks in CI

cavokz:


.github/workflows/tracks.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches: ["*"]
+
+jobs:
+  race:
+    name: Run `${{ matrix.track }}` on stack ${{ matrix.stack-version }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        stack-version: ["8.3.2"]
```
DJRickyB, line 16: this may be better as `latest` since the target is the master branch. it could be varied per target branch to match that branch's stack version

.github/workflows/tracks.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+
+      - name: Install dependencies
+        run: python3 -m pip install "https://github.com/DJRickyB/rally/archive/libs-fix.tar.gz"
```
DJRickyB, line 31: marking TODO change this, blocked on elastic/rally#1474

.github/workflows/tracks.yml
```diff
@@ -0,0 +1,39 @@
...
+      fail-fast: false
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
```
DJRickyB, line 28: Rally itself supports python `3.8`, `3.9`, and `3.10`. Should tracks be tested across at least the min and max supported versions as well? How does that get matrix-ified?

.github/workflows/tracks.yml
cavokz: Rally does not like `latest` as version. See https://github.com/cavokz/rally-tracks/runs/7325773069?check_suite_focus=true.

.github/workflows/tracks.yml
cavokz: With the latest commit I submitted, in trying to make the track/challenge/test-mode flexibility, I made difficult adding new dimensions to the matrix.

Github Actions are not very smart in this regards, either you go fully-generated combinations or you go fully-manual.

.github/workflows/tracks.yml
DJRickyB: sorry, that should be `--revision` flag now instead of `--distribution-version` if going that route: https://esrally.readthedocs.io/en/stable/command_line_reference.html?highlight=latest#revision

.github/workflows/tracks.yml
cavokz: I can deal with this later, when you have fixed your part. I'll notice quite quickly.

.github/workflows/tracks.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install dependencies
+        run: python3 -m pip install "https://github.com/DJRickyB/rally/archive/libs-fix.tar.gz"
+
+      - name: Run race
+        run: |
+          esrally race \
+            --distribution-version=${{ matrix.stack-version }} \
+            --track-path=${{ matrix.track }} \
+            --runtime-jdk=bundled \
+            --test-mode
```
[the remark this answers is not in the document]
cavokz: This is now handled by b437e84770149ed6a9dd28422f3ff43c708ca2c3.

PR #284, opened 2022-07-26
add a frequent-items challenge to the stack overflow track

hendrikmuhs:
add a challenge for `frequent_items`, a new aggregation added in https://github.com/elastic/elasticsearch/pull/83055

to be run with: `--challenge=frequent-items` and `--car="trial-license"`


so/challenges/default.json
```diff
@@ -50,4 +50,72 @@
...
+    },
+    {
+      "name": "frequent-items",
+      "description": "Checks the performance of the frequent item aggregation",
+      "default": false,
+      "schedule": [
+        {
+          "operation": "delete-index"
+        },
+                {
```
droberts195, line 62: nit: too much indenting

so/challenges/default.json
```diff
@@ -50,4 +50,72 @@
...
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {{index_settings | default({}) | tojson}}
+          }
+        },
+        {
+          "name": "check-cluster-health",
+          "operation": {
+            "operation-type": "cluster-health",
+            "index": "logs-*",
```
droberts195, line 72: Like with #285, should this be waiting for green on `so` rather than `logs-*`?

droberts195 (approved):
LGTM, if CI passes.

It looks like CI has been changed in the last few days so this PR needs master merging into it to pick up the new CI scripts.

pquentin:
@elasticmachine please update branch

pquentin:
@elasticmachine update branch

so/challenges/default.json
```diff
@@ -48,6 +101,21 @@
             "retry-until-success": true,
             "include-in-reporting": false
           }
+        },
+        {
+          "operation": "frequent_items_all_100",
+          "warmup-iterations": 3,
+          "iterations": 10
```
DJRickyB, line 108: Just a drive-by question, are these particularly long-running or intensive queries? You will only have 10 deciles if you only have  10 iterations, FYI, so we normally run >=100 ops in order to get a good sense of the distribution (unless it would cause the benchmark to run too long)

hendrikmuhs: yes,  `frequent_items` is quite heavy. On the other hand I don't want to simplify the query or reduce the data. Reducing iterations seems therefore reasonable, however we can discuss whether it is 10 or 20 or 50. What's an acceptable total runtime in your opinion?

DJRickyB: > What's an acceptable total runtime in your opinion?

I don't think there's a correct answer to this question, it's a matter of preference. How much variance is there in your measured service time? If there's a wide distribution you may want more iterations, and if not then you might be fine with 10.

hendrikmuhs: There isn't a wide distribution. For the purpose of e.g. finding performance regressions or to proof/disproof enhancements 10 is sufficient.

pquentin:
@hendrikmuhs Now that I merged #285, we have a conflict. Can you please take a look and fix them? Thanks and sorry for the long delays.

so/challenges/default.json
```diff
@@ -159,4 +159,72 @@
...
+          "operation": {
+            "operation-type": "create-index",
+            "settings": {{index_settings | default({}) | tojson}}
+          }
+        },
+        {
+          "name": "check-cluster-health",
+          "operation": {
+            "operation-type": "cluster-health",
+            "index": "logs-*",
```
DJRickyB, line 181: You've re-introduced the `logs-*` here during the merge fix

DJRickyB (approved):
LGTM

DJRickyB:
@elasticmachine run rally-tracks/it

PR #288, opened 2022-07-28
Fix listing of track `elastic/endpoint`

cavokz:


cavokz:
Related to https://github.com/elastic/rally/issues/1474.

PR #289, opened 2022-07-28
Add integration tests based on pytest-rally.

michaelbaamonde:
This PR introduces a suite of integration tests for rally-tracks using the (in-development) [pytest-rally](https://github.com/elastic/pytest-rally) plugin.

# Tests introduced

There are two types of tests: autogenerated tests for tracks that can be straightforwardly tested, and hand-written tests for more complex tracks that require custom parameters, various different scenarios, etc.

## Autogenerated

This [test class](https://github.com/michaelbaamonde/rally-tracks/blob/d940e02936aff46195b4942f926074ceec183903/it/test_all_tracks_and_challenges.py) relies on `pytest-rally`  to automatically generate a unique test per challenge per track. These tests run a race in test-mode and assert that the return code is 0.

We skip certain tracks, however. The `sql` track only contains one challenge, which does not support test-mode. We can work around this (see the track's [README](https://github.com/elastic/rally-tracks/blob/master/sql/README.md#testing)), but downloading the full corpus takes a while, so I'm just skipping for now as we iterate on this.

The `elastic/logs` and `elastic/security` tracks require custom parameters for testing, so we'll need to skip the autogenerated versions and instead hand-write tests.

## Hand-written

This [file](https://github.com/michaelbaamonde/rally-tracks/blob/d940e02936aff46195b4942f926074ceec183903/it/test_logs.py) translates the `elastic/logs` tests found in the [smoke test shell script](https://github.com/elastic/rally-tracks/blob/master/elastic/it/smoke-test.sh) into Python, using fixtures provided by `pytest-rally`. If we like this approach, we'll migrate other tests under [elastic/it](https://github.com/elastic/rally-tracks/tree/master/elastic/it) similarly.

# How to run them

Future PRs will introduce infrastructure for installing dependencies and running these tests in CI, but for now, the easiest way for reviewers to test locally is to:

1. In the venv you use for Rally development: `pip install git+ssh://git@github.com/elastic/pytest-rally`
2. `cd` into your local `rally-tracks` and check out this PR branch: `git fetch origin pull/289/head:pytest-rally && git checkout pytest-rally`
3. From the top level of `rally-tracks`, run `pytest it/ --log-cli-level=INFO`. The live log will show what's happening under the hood, most interestingly logging the `esrally race...` commands that the plugin generates for each test.
4. The above invocation will build ES from source. You can also run tests against a distribution version: `pytest it/ --log-cli-level=INFO --distribution-version=8.2.0`

## Notes

- The query-related tests in `test_logs.py` are rather slow on my machine. I'm curious if others see this, as well.


michaelbaamonde:
@elasticmachine update branch

DJRickyB (changes requested):
Can you include documentation on how to run (even just if it's mostly copy and pasted from this PR) to a doc in the source?

michaelbaamonde:
@DJRickyB I'm currently working on the automation components of this (installing dependencies, Makefile niceties, etc.) which will make running these tests pretty much turnkey and mostly self-explanatory. I'll just push to this PR branch when it's ready and give you a shout.

michaelbaamonde:
The latest commits introduce some net-new infrastructure into the repo. @DJRickyB and I are going to discuss synchronously tomorrow and walk through the workflow. Converting this to a draft in the meantime.

DJRickyB (approved):
This LGTM. I've played with each of the following
```
make venv-create
make install
make test
make shell
hatch -e unit run test
hatch -e it run test
```

and checked I didn't get anything extra in `git status` that I didn't start with. Excited to get started with this for all our PRs

PR #294, opened 2022-08-10
Many shards snapshots challenge for elastic/logs

dliappis:
This commit adds a new challenge for many-shards that benchmarks ES
performance while taking several (specified by the track parameter
`snapshot_counts`) snapshots.

The intention is to initially use it for tracking regressions by
visualizing the `service_time` of the `wait-for-snapshots` task.

This PR depends uses the Rally operation
`wait-current-snapshots-create` operation (introduced in
https://github.com/elastic/rally/pull/1542).

### DIFF
--- a/elastic/logs/challenges/many-shards-snapshots.json
+++ b/elastic/logs/challenges/many-shards-snapshots.json
@@ -36,7 +36,7 @@
       }
     },
     {# Issue <snapshot_counts> number of back to back snapshots and wait until all have completed #}
-      {% for snap_iter in range(snapshot_counts) %}
+      {% for snap_iter in range(snapshot_counts | default(100)) %}
       {
         "name": "create-snapshot-{{ snap_iter }}",
         "operation": {
--- a/elastic/logs/challenges/many-shards-snapshots.json
+++ b/elastic/logs/challenges/many-shards-snapshots.json
@@ -1,0 +1,65 @@
+{% import "rally.helpers" as rally %}
+{
+  "name": "many-shards-snapshots",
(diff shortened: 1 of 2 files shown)


elastic/logs/challenges/many-shards-snapshots.json
```diff
@@ -0,0 +1,65 @@
...
+        "param-source": "initial-indices-source",
+        "detailed-results": true,
+        "initial_indices_count": {{ p_initial_indices_count }},
+        "name": "auditbeatquantitative",
+        "timeout": "10m"
+      },
+      "clients": {{ p_bulk_indexing_clients }}
+    },
+    {
+      "name": "register-snapshot-repository",
```
pquentin, line 28: I noticed that no existing track uses the Rally's `delete-snapshot-repository` operation, any idea why? I guess it would be more robust for local execution?

I'm mostly asking for the challenge I'm working on, I don't think this should block this pull request!

pquentin (approved):
Thanks! LGTM.

elastic/logs/challenges/many-shards-snapshots.json
dliappis: That's a good question. Such an operation would make sense when >1 iterations of the same challenge run within the same use case. Even then, I am not sure that one needs to delete the repository; I was under the impression that if you re-register the same repository with the exact same parameters, it'll just verify the repository (we could test that). This is probably what happens when several sequential restorations happen -- in the form of several experiments including ES stop/starts -- in logging query benchmarks.

pquentin: Yeah I think it's only useful if you made a mistake when developing the track and want to start from a clean slate.

PR #295, opened 2022-08-10
Fix dependency issue for endpoint tests

DJRickyB:
The following two changes (along with a `make install`) resolved failing tests for me

### DIFF
--- a/elastic/Makefile
+++ b/elastic/Makefile
@@ -59,7 +59,7 @@ install: venv-create
 	# install pytest for tests
 	. $(VENV_ACTIVATE_FILE); pip3 install pytest==6.2.5 pytest-benchmark==3.2.2
 	# install dependencies for tests
-	. $(VENV_ACTIVATE_FILE); pip3 install geneve==0.0.3 pytest-asyncio==0.18.1 'https://github.com/elastic/package-assets/archive/main.tar.gz'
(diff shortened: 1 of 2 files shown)


elastic/Makefile
```diff
@@ -59,7 +59,7 @@ install: venv-create
 	# install pytest for tests
 	. $(VENV_ACTIVATE_FILE); pip3 install pytest==6.2.5 pytest-benchmark==3.2.2
 	# install dependencies for tests
-	. $(VENV_ACTIVATE_FILE); pip3 install geneve==0.0.3 pytest-asyncio==0.18.1
+	. $(VENV_ACTIVATE_FILE); pip3 install geneve==0.0.3 pytest-asyncio==0.18.1 'https://github.com/elastic/package-assets/archive/main.tar.gz'
```
pquentin, line 62: What are your thoughts on using `git+https://github.com/elastic/package-assets` instead? It fetches the package from git instead of what appears to be an obscure undocumented archive? This is basically how we install Rally below.

cc @cavokz

pquentin (approved):
Thanks! This fixes the test locally for me. We'll have to add them to CI. Left one question.

elastic/Makefile
DJRickyB: Can the version (branch/revision) be pinned with `git+https`?

DJRickyB: fixed in [6f82c67](https://github.com/elastic/rally-tracks/pull/295/commits/6f82c6781911d23286f880dfd9851e6c9e7ce2a5)

pquentin (approved):
Thanks! LGTM.

PR #296, opened 2022-08-11
dense_vector: store refresh-after-index metrics in rally-results

pquentin:
index-append can appear deceptively fast because the HNSW graph can be
built during refresh, so we want refresh times too.

### DIFF
--- a/dense_vector/challenges/default.json
+++ b/dense_vector/challenges/default.json
@@ -77,7 +77,8 @@
       "name": "refresh-after-update",
       "operation": {
         "operation-type": "refresh",
-        "request-timeout": 1000
+        "request-timeout": 1000,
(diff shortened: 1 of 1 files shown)


dense_vector/challenges/default.json
```diff
@@ -36,7 +36,8 @@
       "name": "refresh-after-index",
       "operation": {
         "operation-type": "refresh",
-        "request-timeout": 1000
+        "request-timeout": 1000,
+        "include-in-reporting": true
```
mayya-sharipova, line 40: @pquentin It would be useful also do this `include-in-reporting` for `refresh-after-update`.

mayya-sharipova (approved):
@pquentin Thanks a lot!

PR #298, opened 2022-08-11
Add support for mounting searchable snapshots

pquentin:
This is copied from https://github.com/elastic/rally-eventdata-track/blob/master/eventdata/runners/mount_searchable_snapshot_runner.py but:

 * this uses a simple function instead of a class
 * stops trying to figure out the response format as Elasticsearch 8.x
   now uses the 7.x format.

This will be used for testing the frozen tier on elastic/logs.


pquentin:
This will probably evolve as I'm using it, and may even end up in Rally. Closing for now.

PR #281, opened 2022-07-12
Add `elastic/endpoint`

cavokz:

PR #299, opened 2022-08-11
don't glob recursively

DJRickyB:
As a result of recent PR #281 we aimed to simplify file globbing in the ILM param source. Unfortunately, we did *not* want to recursively scan for ILM objects, since the many-shards ones are separated from the standard ones via nested directories. This change reverts to the non-recursive method for seeking ILM objects.


elastic/tests/runners/resources/ilm/nested/dont-read-me.json
```diff
@@ -0,0 +1,3 @@
+{
+  "policy": {}
```
DJRickyB, line 2: I didn't have to change the test itself, but adding this file cause the existing ilm test to fail, then succeed after adding the above fix :)

DJRickyB:
I'm going to ignore `index-alert-source-events` failures for now

michaelbaamonde (approved):
Makes sense. LGTM.

PR #300, opened 2022-08-18
Update synthetic source config

nik9000:
This flips `text` fields from having a sub-keyword field to `stored`
fields and it re-enables `ignore_above` on all `keyword` fields now that
synthetic source supports that.


nik9000:
I'm using this to test the performance of https://github.com/elastic/elasticsearch/pull/89466 - once that's merged we can merge this.

DJRickyB (approved):
LGTM from a correctness standpoint

nik9000:
It will change "what we're doing" in the nightlies for tsdb, but it'll change it to something a little more realistic.

nik9000:
@DJRickyB What should I do about backporting this? It can only run on 8.5+.

PR #301, opened 2022-08-22
Add Jenkins jobs

michaelbaamonde:
This PR adds Jenkins job definitions. The structure is similar to what we do in [Rally](https://github.com/elastic/rally/pull/1561) and builds on https://github.com/elastic/rally-tracks/pull/289, which introduced integration testing infrastructure into this repo.

Here's a screenshot from my local testing:

| rally-tracks-ci-screenshot |
|:--:| 
| *I tested using `ci` as the default branch, but in production, it will be `master`, so the job names will reflect that.* |

A couple of notes:

- This is intended to replace the GitHub workflows that we've been using, so I've deleted them. Several of these workflows actually should have been failing this whole time (e.g. those that test `elastic/logs`), but the race invocations don't specify `--on-error=abort`, which is an oversight on our part. Integration tests using `pytest-rally` abort on error by default, so we'll catch issues before merging PRs.
- I've [added a few integration tests](https://github.com/elastic/rally-tracks/commit/8d4f4d9aa3ca3b511c610b52aa48a5fc0e32d29a) for the `elastic/security` track. A future PR will finish converting the shell scripts in `elastic/it` to Python in this fashion, but I wanted to have make sure we had some coverage before removing the GH workflows.


.ci/build.sh
```diff
@@ -0,0 +1,144 @@
...
+
+# fail this script immediately if any command fails with a non-zero exit code
+set -e
+# fail on pipeline errors, e.g. when grepping
+set -o pipefail
+# fail on any unset environment variables
+set -u
+
+# based on https://gist.github.com/sj26/88e1c6584397bb7c13bd11108a579746?permalink_comment_id=4155247#gistcomment-4155247
+function retry {
```
dliappis, line 28: As a future enhancement I've been aiming to replace this helpers with native python code in a future iteration (at least in the repo where we first started using them). Maybe a TODO is worth to remind us about refactoring, or at least DRYing up the common bash functions somehow.

dliappis (commented):
Thanks for this massive improvement!

I've only managed to take look at a few files, leaving a little nit for now, will continue the review tomorrow

.ci/build.sh
```diff
@@ -0,0 +1,144 @@
...
+function install {
+  java_common
+  python_common
+  install_python_prereq
+  make install
+}
+
+function precommit {
+  install
+  make test
```
DJRickyB, line 116: please also invoke the brand new `pre-commit` make target here

DJRickyB:
```suggestion
  make test
  make pre-commit
```

.ci/build.sh
michaelbaamonde: Yeah, I agree that we should come up with a better, reusable approach in the future. Comment added.

.ci/jobs/pull-request-it.yml
```diff
@@ -0,0 +1,27 @@
...
+      - github-pull-request:
+          org-list:
+            - elastic
+          allow-whitelist-orgs-as-admins: true
+          trigger-phrase: '.*run\W+rally-tracks/it.*'
+          github-hooks: true
+          status-context: "rally-tracks/it"
+          cancel-builds-on-update: true
+          black-list-labels:
+            - '>test-mute'
```
alexsapran, line 22: drive-by comment, I was looking for a way to disable tests for draft PR or PR that I raised and not intended to be merged. This setting will not trigger any PR with the label `>test-mute` so maybe we want to add that label as well.

michaelbaamonde: Sure, we can do that.

.ci/jobs/pull-request-precommit.yml
```diff
@@ -0,0 +1,27 @@
+---
```
dliappis, line 1: This job and the previous one `pull-request-it` look almost identical. It might be premature optimization at this stage, but we could consider turning them to templates, as the only things that seems to change is `precommit`/`it` and the `description`.

dliappis (approved):
I didn't test this locally but did a pass. LGTM.

PR #302, opened 2022-08-23
Support frozen tier logging benchmarks

pquentin:
This does two things:

 * add a runner that can mount partial indices
 * add a challenge calling that runner

While the runner should eventually be in Rally and I think I should have targeted the master branch, I would be happy to get some intial feedback here.

### DIFF
--- a/elastic/logs/challenges/logging-querying.json
+++ b/elastic/logs/challenges/logging-querying.json
@@ -22,14 +22,14 @@
         "clients": {{ p_bulk_indexing_clients }},
         "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
       },
+      {
+        "name": "compression-stats",
+        "operation": {
+          "operation-type": "compression-statistics",
+          "param-source": "create-datastream-source"
+        }
+      },
     {% endif %}
-    {
-      "name": "compression-stats",
-      "operation": {
-        "operation-type": "compression-statistics",
-        "param-source": "create-datastream-source"
-      }
-    },
     {
       "name": "logging-queries",
       "parallel": {
--- a/elastic/tests/runners/snapshot_test.py
+++ b/elastic/tests/runners/snapshot_test.py
@@ -29,7 +29,7 @@ async def test_mount_snapshot(es):
         {
             "snapshots": [
                 {
-                    "snapshot": "eventdata-snapshot",
+                    "snapshot": "logging-snapshot",
                     "uuid": "mWJnRABaSh-gdHF3-pexbw",
                     "indices": [
                         "elasticlogs-2018-05-03",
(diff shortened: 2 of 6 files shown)


elastic/tests/runners/snapshot_test.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+@mock.patch("elasticsearch.Elasticsearch")
+@pytest.mark.asyncio
+async def test_mount_snapshot(es):
+    es.snapshot.get.return_value = as_future(
+        {
+            "snapshots": [
+                {
+                    "snapshot": "eventdata-snapshot",
```
dliappis, line 32: this workload should be put to rest, maybe we rename to something else?

elastic/logs/challenges/logging-querying.json
```diff
@@ -22,14 +22,14 @@
...
       },
+      {
+        "name": "compression-stats",
+        "operation": {
+          "operation-type": "compression-statistics",
+          "param-source": "create-datastream-source"
+        }
+      },
     {% endif %}
-    {
```
dliappis, line 26: by moving this inside the conditional, we are sure it won't break https://github.com/elastic/elasticsearch-sizing/blob/master/logging/run_books/index_compression.md, right?

dliappis (commented):
Did a very quick initial pass, the approach looks sane. Left a few small comments.

elastic/logs/challenges/logging-querying.json
inqueue: LGTM on the first pass. I'd want a test to ensure this concern is a non-issue.

pquentin: This change does not break that use case because index compression stats rely on indexing. I'll look into adding a test.

pquentin: I actually have no idea how to test that, this is another reason to prefer having logic outside of the tracks.

elastic/tests/runners/snapshot_test.py
pquentin: Addressed in https://github.com/elastic/rally-tracks/pull/302/commits/b7600440fcd38ad5fb5c001ce1e861eb17bf30b0

elastic/logs/challenges/logging-querying.json
dliappis: Did a little bit of digging, I think the index compression methodology won't get broken with this change because it uses a different challenge: https://github.com/elastic/rally-tracks/blob/7354a0307ae67463adabdde4fc1719f132f2ce80/elastic/logs/challenges/logging-disk-usage.json#L92

dliappis (approved):
LGTM (but we should be merging to master first and then backport to this branch)

pquentin:
Since https://github.com/elastic/rally-tracks/pull/306 was merged, I just merged this one as a backport.

PR #308, opened 2022-08-26
Finish black migration

pquentin:
We need to install black in the developer virtual environment to allow
editors to use it. And there are various places where the formatting
could be improved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,8 @@ export HATCHLING_VERSION = "1.6.0"
 export PIP_VERSION = "22.2"
 export WHEEL_VERSION = "0.37.1"
 export PRE_COMMIT_VERSION = "2.20.0"
+export BLACK_VERSION = "22.3.0"
+export ISORT_VERSION = "5.8.0"
 VIRTUAL_ENV ?= .venv
 VENV_ACTIVATE_FILE = $(VIRTUAL_ENV)/bin/activate
 VENV_ACTIVATE = . $(VENV_ACTIVATE_FILE)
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -54,8 +56,8 @@ check-venv:
 	printf $(VE_MISSING_HELP); \
 	fi
 
-install: venv-create
-	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade hatch==$(HATCH_VERSION) hatchling==$(HATCHLING_VERSION) pip==$(PIP_VERSION) wheel==$(WHEEL_VERSION) pre-commit==$(PRE_COMMIT_VERSION)
+install:
+	. $(VENV_ACTIVATE_FILE); $(PIP_WRAPPER) install --upgrade hatch==$(HATCH_VERSION) hatchling==$(HATCHLING_VERSION) pip==$(PIP_VERSION) wheel==$(WHEEL_VERSION) pre-commit==$(PRE_COMMIT_VERSION) black==$(BLACK_VERSION) isort==$(ISORT_VERSION)
```
cavokz, line 60: Shall we update `.pre-commit-config.yaml` so to use `black` and `isort` as installed here instead of having them independently managed by `pre-hook`?

pquentin: We could do that but then why have pre-commit at all? Given our use of pylint in Rally pre-commit does not help much anyway. I'm considering reverting https://github.com/elastic/rally/pull/1546 for this reason.

cavokz: I've not any strong opinion about whether to use pre-commit or not, I just replicated what was done in Rally. Feel free to remove it also from here, if you have any second thought.

I simply don't think is good to use different versions of the tools here and there. Honestly, I think we should just have them as develop dependencies in pyproject.toml (where I expect them to be) and remove them from the Makefile as well.

pquentin: I added a new commit to store the version in a single location. (We should move to pyproject.toml but it's orthogonal)

cavokz: Thanks.

cavokz (approved):

PR #313, opened 2022-09-01
Use dynamic package assets in `elastic/logs` track

michaelbaamonde:
Draft for now. I'll be doing some benchmarking with this vs. what's on `master` as well as some cleanups. We'll also need to align on branching, as there are some complexities re: ES version compatibility.


michaelbaamonde:
Closing this PoC PR, as we've decided to go with an alternative approach.

PR #315, opened 2022-09-06
Synthetic _source support in so_vectors

nik9000:
This modifies the `so_vectors` track to support synthetic _source and to test fetch performance on docs that are hot in the disk cache.

### DIFF
--- a/so_vector/index.json
+++ b/so_vector/index.json
@@ -5,7 +5,7 @@
   },
   "mappings": {
     "_source": {
-      "excludes": ["titleVector"]
+      "mode": {{ source_mode | default("stored") | tojson }}
     },
     "properties": {
       "userId": {
@@ -22,6 +22,7 @@
       },
       "title": {
         "type": "text"
+        {% if source_mode == "synthetic" %},"store": true{% endif %}
       },
       "titleVector": {
         "type": "dense_vector",
@@ -34,6 +35,7 @@
       },
       "body": {
         "type": "text"
+        {% if source_mode == "synthetic" %},"store": true{% endif %}
       }
     }
   }
--- a/so_vector/operations/default.json
+++ b/so_vector/operations/default.json
@@ -1,4 +1,23 @@
 {
+  "name": "default",
+  "operation-type": "search",
+  "body": {
+    "query": {
+      "match_all": {}
+    }
+  }
+},
+{
+  "name": "default_1k",
+  "operation-type": "search",
+  "body": {
+    "query": {
+      "match_all": {}
+    },
+    "size": 1000
+  }
+},
+{
   "name": "knn-search-10-50-acceptedAnswerId",
   "operation-type": "raw-request",
   "path": "/vectors/_knn_search",
--- a/so_vector/operations/default.json
+++ b/so_vector/operations/default.json
@@ -1,5 +1,5 @@
 {
-  "name": "default",
+  "name": "fetch_10",
   "operation-type": "search",
   "body": {
     "query": {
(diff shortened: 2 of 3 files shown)


nik9000:
Requires https://github.com/elastic/elasticsearch/pull/89840

nik9000:
Also pairs well with https://github.com/elastic/rally/pull/1573

so_vector/index.json
```diff
@@ -5,7 +5,7 @@
   },
   "mappings": {
     "_source": {
-      "excludes": ["titleVector"]
```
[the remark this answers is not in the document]
mayya-sharipova: Unrelated to this PR, where to find documentation on different source modes and what they mean?
I could not find details in https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html#synthetic-source

so_vector/index.json
```diff
@@ -22,6 +22,7 @@
       },
       "title": {
         "type": "text"
+        {% if source_mode == "synthetic" %},"store": true{% endif %}
```
mayya-sharipova, line 25: Does it mean that synthetic source is not supported for text fields?

nik9000: It *is*, but only if you `store` them or have a `keyword` sub field. Depending on your goals with the `text` field you might want to do one or the other and I didn't want to assume for anyone.

so_vector/index.json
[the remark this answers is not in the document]
nik9000: @romseygeek did you document the other values for `mode` anywhere? The other values a `stored` which is our old behavior and `disabled` which is the same as `enabled: false` for `_source`.

Now! I've realized I haven't really looked into `excludes` and `includes` and have added it to a brainstorm session tomorrow. I don't know what it does now. And I'm reasonably sure we shouldn't support it with `mode: synthetic`.

so_vector/operations/default.json
```diff
@@ -1,3 +1,22 @@
...
+      "match_all": {}
+    }
+  }
+},
+{
+  "name": "fetch_1000",
+  "operation-type": "search",
+  "body": {
+    "query": {
+      "match_all": {}
```
mayya-sharipova, line 15: `match_all` will retrieve first 1000  consecutive docs based on Lucene doc IDs.  Should we instead try to fetch randomly located docs by for example using `_knn_search`?

nik9000: We use a pretty similar operation on the other places that support synthetic source. I'd *love* to have a consistent way to fetch random IDs - probably in addition to this. I'd like to be able to compare fetching hot disk blocks to fetching random ones. I know we have a random score, but I've a bit worried the random score will dominate the time.

jtibshirani:
@nik9000 thanks for all your work on synthetic source for vectors!

I would prefer to keep this track as-is to best represent what our users are doing. In general, I think text search users will drop the vectors from the _source using a simple `excludes`. There is not a strong use case for retrieving the vector values during a search -- users want to avoid returning these large JSON payloads. And many text search users have a different "source of truth" that they can reindex from, so it is okay for them to modify _source in this way.

nik9000:
Hey! Just so I know precisely what you need- do you want none of the
changes I've proposed or are any parts worth it?

I think it's possible to keep the source modification when not running with
synthetic source enabled, for example.

jtibshirani:
For now I think we should keep the track simple and not make any of these changes. It has enough fancy new features for one track (approximate kNN with filtering) :)

I don't think users will want to use synthetic source instead of just disabling the vector in _source. And we are not interested in raw fetch performance for vectors.

PR #317, opened 2022-09-13
Remove the downsample challenge from the tsdb track

salvatore-campagna:
This challenge requires a new operation which is still not merged. 
The new operation (downsample) is part of a set of changes to Rally. 
The PR that adds support for the new operation in Rally uncovered
an issue with the delete index operation not completing correctly. 
As a result, we temporarily remove the new downsample challenge 
from the tsdb track. Once the new operation is merged we will add 
back this challenge.


DJRickyB (approved):
LGTM thank you

PR #318, opened 2022-09-13
CI fixes

michaelbaamonde:
This gets CI jobs running and passing.

- `make install` now ensures that the venv is created first
- The security tests are formatted correctly
- We're seeing the same bug that caused https://github.com/elastic/rally-tracks/pull/317 in the security tests. I marked the failing one as `xfail` until we're able to figure out what's causing this.


DJRickyB (approved):
LGTM

PR #320, opened 2022-09-14
Re-enable security test

DJRickyB:
Now that elastic/rally#1580 is merged, this should be fixed.

### DIFF
--- a/it/test_security.py
+++ b/it/test_security.py
@@ -25,7 +25,6 @@ class TestSecurity:
         ret = rally.race(track="elastic/security", challenge="security-indexing", track_params={"number_of_replicas": "0"})
         assert ret == 0
 
-    @pytest.mark.xfail(reason="index deletion bug under investigation")
     def test_security_indexing_querying(self, es_cluster, rally):
         ret = rally.race(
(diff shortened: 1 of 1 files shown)


pquentin (approved):
Thanks! It does pass now. Should we also define `xfail_strict=true` to have CI fail if such xfail tests pass?

DJRickyB:
> Thanks! It does pass now. Should we also define `xfail_strict=true` to have CI fail if such xfail tests pass?

My immediate reaction is "oooo", as I wasn't aware of this option

But I think this would potentially not work well for us because the fix to a given test failure (as it was here) could be in `rally`, meaning we'd technically have a CI deadlock for an xfail test being marked in `rally-tracks` master vs fixed in a branch of `rally`. Am I thinking about this correctly? I can explain this further if I'm not making sense

pquentin:
Right, but we can always bypass the CI checks and merge anyway. At this point it's a balance: are we going to break more things by bypassing checks sometimes or by missing xfail annotations that are in fact passing? I think you're right that bypassing the checks would be more dangerous, I would not trust myself here!

PR #327, opened 2022-09-26
Synchronize all pushes to 'master' branch with '8.5' branch

pquentin:
As discussed offline, this should reduce or backporting burden. I'm using the same action that Elasticsearch used to sync from master to main. It's not tested, we may have to merge now and debug later.


DJRickyB (approved):
> we may have to merge now and debug later.
Let's do it! LGTM

pquentin:
@DJRickyB So I think my pull request worked, but it complains that 8.5 and master have diverging history, because only some commits got backported. https://github.com/elastic/rally-tracks/actions/runs/3195702505/jobs/5216686865

Is it OK to force push 8.5 with the contents of master?

DJRickyB:
> Is it OK to force push 8.5 with the contents of master?

Yes, definitely

PR #328, opened 2022-09-29
Add filters aggregation to noaa track

martijnvg:
This adds four `filters` aggregation tasks to default challenge.
Also another challenge has been added that only runs `filters` aggregations and two `filter` aggregations that are equivalent to two other `filters` aggregations that are also defined (to see performance difference).


nik9000 (commented):
I'm quite happy to have the extra filter performance numbers. I'm not great at reviewing these though. it's probably going to need double checking by someone on the perf team.

martijnvg:
I changed this PR , so that the filter aggregations aren't ran during the `aggs` challenge, but only via the newly added challenge. The reason is that we need to go over all tasks that are currently executed during the nightly run and decide what is important to keep track of. At the moment it is a tedious task to investigate a performance degradation.

noaa/challenges/default.json
```diff
@@ -226,4 +226,68 @@
...
+           {%- endif %}
+        },
+        {
+          "name": "refresh-after-force-merge",
+          "operation": "refresh",
+          "clients": 1
+        },
+        {% include 'filters-aggs-tasks.json' %}
+        {
+          "operation": "filter-agg-with-high-cardinality-filter",
```
dliappis, line 281: is there a reason why the two operations `filter-agg-with-high-cardinality-filter` and `filter-per-10-degrees-interval` don't live within the included `filters-aggs-tasks.json`?

dliappis (commented):
Thanks for raising this new challenge. Left a few minor questions.

noaa/challenges/default.json
martijnvg: These two operations test run the `filter` aggregation and the aggregations in the `filters-aggs-tasks.json` file run the `filters` aggregation. The refactoring in elastic/elasticsearch#90318 changes the implementation of `filter` aggregation to use the implementation of the `filters` aggregation. And `filter` aggregations defined here exist just to double check we got the expected performance improvement and these `filter` aggregation don't need to run highly with all the other aggregations.

Initially the idea was to run `filters` aggregation as part of the `aggs` challenge, but we should first take a look at all the other operations that run as part of the `aggs` challenge. As the number of operations make investigating performance regressions more difficult.

noaa/challenges/default.json
```diff
@@ -226,4 +226,68 @@
         },
         {% include 'aggs-tasks.json' %}
       ]
+    },
+    {
+      "name": "filter-aggs",
+      "description": "Checks the performance of many aggregations",
```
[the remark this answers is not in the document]
martijnvg: I've updated the description: 3e3f3a44e780e7dcf4f5b4c455d58fbff2d12b73

dliappis (approved):
LGTM

dliappis:
@martijnvg I will merge this now, but we need backports. How far back do you want it backported? Here are the current branches that this repo has: https://github.com/elastic/rally-tracks/branches

nik9000:
The filter agg should be able to go as far back as you care to try and go.

PR #330, opened 2022-10-04
[Dense Vector] Move to _search API and add random query vector support

DJRickyB:
This PR moves KNN searches from `raw-request` to `search` operations, and adds functionality to allow to use random query vectors per iteration instead of needing to provide a static query vector for a given task.


dense_vector/track.py
```diff
@@ -0,0 +1,31 @@
+import random
+
+
+def knn_param_source(track, params, **kwargs):
+    # choose a suitable index: if there is only one defined for this track
+    # choose that one, but let the user always override index
+    random.seed(31)
```
jtibshirani, line 7: Does this mean that the same sequence of random vectors will be produced across race runs? I think that's okay, just wanted to check.

jtibshirani (commented):
This makes sense to me. It'd be great to update the `so_vector` track as well to use the new parameter source.

dense_vector/track.py
```diff
@@ -0,0 +1,31 @@
...
+    else:
+        default_index = "_all"
+
+    index_name = params.get("index", default_index)
+
+    return {
+        "body": {
+            "knn": {
+                "field": "vector",
+                "query_vector": params.get("query-vector", [random.uniform(-1.0, 1.0) for _ in range(96)]),
```
mayya-sharipova, line 19: Just to clarify, we will still have an option to provide a specific `query-vector`?
One thing we wanted to add to the track is to calculate recall metric, and this depends on the query to be specific.

jtibshirani: @mayya-sharipova that's right, you're still able to specify a specific query vector. It only randomly generates one if the `query-vector` parameter is missing. The current PR still uses specific queries, and doesn't use randomization yet.

dense_vector/track.py
DJRickyB: yes

DJRickyB: commented to that effect

DJRickyB:
> This makes sense to me. It'd be great to update the so_vector track as well to use the new parameter source.

`so_vector` will require a little more effort as we'd have to consider the additional parts of the query body. Probably not terribly difficult but I'd have to know if you wanted it done as part of this PR

DJRickyB:
Opened https://github.com/elastic/rally-tracks/issues/331 to work on so_vector

dense_vector/track.py
```diff
@@ -0,0 +1,32 @@
...
```
jtibshirani, line 20: Sorry @DJRickyB , I missed an important consideration here. The vector similarity we're using requires that all vectors have unit length. So we'll need to normalize these random vectors. Here's an example using NumPy, but it's a simple formula that we could also code by hand: https://stackoverflow.com/questions/21030391/how-to-normalize-a-numpy-array-to-a-unit-vector. Currently the searches will fail if we try to use a random query vector.

Alternatively, would it be possible to check in a "query vector" file of ~1000 vectors and use that to create all queries? We have such a file available, and the vectors are already normalized, and more realistic than random vectors.

PR #332, opened 2022-10-13
Use list of queries for dense_vector track

jtibshirani:
This commit updates the dense_vector track to use a list of unique query
vectors during the search operations, instead of repeating a single query
vector. This helps ensure the timings are realistic. It's also a step towards
reporting recall in the benchmarks, since we'll be able to average over a set
of realistic query vectors.


jtibshirani:
One thing I'm not sure about -- is it fine to check in `queries.json`, which has 2000 query vectors and is a bit large?

dense_vector/track.py
```diff
@@ -1,32 +1,60 @@
...
+        result = {"index": self._index_name, "cache": self._params.get("cache", False), "size": self._params.get("k", 10)}
+
+        if self._exact_scan:
+            result["body"] = {
+                "query": {
+                    "script_score": {
+                        "query": {"match_all": {}},
+                        "script": {
+                            "source": "cosineSimilarity(params.query, 'vector') + 1.0",
+                            "params": {"query": self._queries[self._iters]},
```
DJRickyB, line 39: nit, but this would allow arbitrarily many invocations if you wanted:
```suggestion
                            "params": {"query": self._queries[self._iters % 2000]},
```

DJRickyB (approved):
LGTM, left one IMO optional suggestion

DJRickyB:
> One thing I'm not sure about -- is it fine to check in queries.json, which has 2000 query vectors and is a bit large?

I think this is fine, unless you have a way to just procedurally generate them during `init` (during `params` would cause runtime overhead on the query executions)

mayya-sharipova (approved):
@jtibshirani Thanks, that's a great idea, and indeed a step towards reporting recall

DJRickyB:
> @jtibshirani Just for my education, how this is going to work? We take first 100 queries from queries.json for warmup? And then take the first 1000 queries from queries.json for real measurement?

The same instance of the parameter source is used for the whole task. So in this configuration the first 100 queries will count as warmup, then queries 101-1100 will be measured normally.

jtibshirani:
> I think this is fine, unless you have a way to just procedurally generate them during init (during params would cause runtime overhead on the query executions)

There's not a way to procedurally generate them, since they're taken from a dataset of query vectors (published alongside the document vector dataset).

dense_vector/track.py
jtibshirani: It seems clearest to me to fail instead of silently repeating queries, since the goal of the query dataset is to avoid repeating queries.

PR #333, opened 2022-10-14
Update CCS challenge to support multiple clusters

b-deam:
A common architecture for geographically dispersed Logging/o11y use cases is to use the Cross Cluster Search functionality to fan out searches from a central (local) cluster to many 'remote' clusters.  

This PR modifies the existing challenge to better model this common architecture by adding support for an arbitrary number of remote clusters.

The modified remote cluster runners now have the following behaviour:
- `configure-remote-clusters` takes a `local-cluster` parameter, whose identifier is the name of the respective cluster in [Rally's `target-hosts` parameter](https://esrally.readthedocs.io/en/stable/command_line_reference.html#id1). This `local-cluster` is then used as the central cluster from which all other remaining clusters specified in `target-hosts` should be connected to.
- `configure-ccr` takes a `source-cluster` parameter, whose identifier is the name of the cluster from which data should be replicated [via the Follow Index API](https://www.elastic.co/guide/en/elasticsearch/reference/current/ccr-put-follow.html) to all other remaining clusters specified in `target-hosts`

Notes:
- I added [a naive `field_caps` task](https://github.com/elastic/rally-tracks/pull/333/commits/71caba01d4cf89339a644edc8d24fe6cc967e272) that hits the field caps API across the ‘local’ and all remote clusters via `/logs-*,remote*:logs-*/_field_caps?fields=*`. It executes roughly every ~5s on with a [poisson distribution](https://esrally.readthedocs.io/en/latest/track.html#choosing-a-schedule), attempting to mimic the parallel workflow queries.
- I also added an [additional, multi-cluster runner](https://github.com/elastic/rally-tracks/compare/master...b-deam:rally-tracks:many-cluster-support?expand=1#diff-f1a5a6af750cc38dd0f60379bec719d6cdda9a449c7000e4b9459dc464918dd8) to delete data streams from _all_ clusters in `target-hosts`


elastic/shared/runners/remote_cluster.py
```diff
@@ -1,143 +1,179 @@
+import asyncio
+import copy
 import math
 import time
 
 from elasticsearch import ElasticsearchException
 from esrally.driver.runner import Runner
 
 """
-Runners for benchmarking multi-clusters. Once these runners are stable, we can move them to Rally.
```
DJRickyB, line 8: Just curious, as this is no longer in the comment. Do you feel like they're not suitable for Rally?

> Once these runners are stable, we can move them to Rally.

elastic/shared/runners/datastream.py
```diff
@@ -50,6 +50,33 @@ async def create(es, params):
...
+        super().__init__()
+
+    async def __call__(self, multi_es, params):
+        """
+        Deletes specified datastreams from all clusters
+        """
+        ops = 0
+        for cluster_name, cluster_client in multi_es.items():
+            data_stream = mandatory(params, "data-stream", "delete-datastream")
+            logger = logging.getLogger(__name__)
```
DJRickyB, line 66: Just to keep it lean, can we move this to an instance variable in `__init__`?

elastic/shared/runners/datastream.py
```diff
@@ -50,6 +50,33 @@ async def create(es, params):
...
+
+    def __init__(self):
+        super().__init__()
+
+    async def __call__(self, multi_es, params):
+        """
+        Deletes specified datastreams from all clusters
+        """
+        ops = 0
+        for cluster_name, cluster_client in multi_es.items():
```
DJRickyB, line 64: given we always invoke `delete_data_stream` (and we raise on failure), isn't `ops` always `len(multi_es.items())`? Alternatively, if you wanted a slicker way to do this, you could omit the increment `ops +=1` below and instead do: 
```suggestion
        for ops, (cluster_name, cluster_client) in enumerate(multi_es.items()):
```

elastic/shared/runners/datastream.py
```diff
@@ -50,6 +50,33 @@ async def create(es, params):
...
+            data_stream = mandatory(params, "data-stream", "delete-datastream")
+            logger = logging.getLogger(__name__)
+            try:
+                logger.info("Deleting data stream: [%s] in cluster [%s]", data_stream, cluster_name)
+                await cluster_client.indices.delete_data_stream(name=data_stream)
+            except elasticsearch.ElasticsearchException as e:
+                msg = f"Failed to delete datastream [{data_stream}]; [{e}]"
+                raise BaseException(msg)
```
DJRickyB, line 72: trying to remember, is there something about pickle/thespian/serialization that means we shouldn't throw with this?
```suggestion
                raise BaseException(msg, e)
```

elastic/shared/runners/datastream.py
DJRickyB: actually, if you do that, there's no longer a reason to import from `esrally.driver.runner` and we can instead duck type this runner class

DJRickyB: double-actually, I now see the `mandatory` function invocation you're borrowing from `Runner`. I still think duck-typing is worthwhile (and therefore not using `mandatory`) since we only use this runner here, and wouldn't anticipate a scenario where `data-stream` isn't defined anyway

elastic/shared/runners/remote_cluster.py
b-deam: IMHO, these runners are specific enough to the implementation of the challenge that they don't make sense to implement as core runners. Additionally, we do actually indicate that [_all_ multi-clusters runners should be custom](https://esrally.readthedocs.io/en/latest/command_line_reference.html#id1):
> <h1> target-hosts </h1>
> Rally can also create client connections for multiple Elasticsearch clusters. This is only useful if you want to create [custom runners](https://esrally.readthedocs.io/en/latest/adding_tracks.html#adding-tracks-custom-runners) that execute operations against multiple clusters, for example for [cross cluster search](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-cross-cluster-search.html) or cross cluster replication.

elastic/shared/runners/datastream.py
b-deam: Not that I'm aware of, it looks like we do properly catch these via https://github.com/elastic/rally/commit/19d4cfa21c494aac8c68eda5ce0ce52c623dc3ce

elastic/shared/runners/datastream.py
b-deam: Addressed in https://github.com/elastic/rally-tracks/pull/333/commits/73c2bf289d88ddb3fa35c7f092dd7cffdbf59fc8

elastic/shared/runners/datastream.py
b-deam: Nice suggestion! Addressed in https://github.com/elastic/rally-tracks/pull/333/commits/73c2bf289d88ddb3fa35c7f092dd7cffdbf59fc8

elastic/shared/runners/remote_cluster.py
```diff
@@ -1,143 +1,183 @@
...
-    async def __call__(self, multi_es, params):
-        end_request_timeout = time.process_time() + params.get("request-timeout", 7200)
-        required_licenses = ["trial", "platinum", "enterprise"]
-
-        def request_timeout():
-            return math.ceil(end_request_timeout - time.process_time())
-
-        remote_es = multi_es[params["remote-cluster"]]
-        local_es = multi_es[params["local-cluster"]]
+    def request_timeout(self, timeout):
```
DJRickyB, line 109: do we need this function? It seems like it always returns `timeout` unless `time.process_time()` takes > 1s to run

elastic/shared/runners/remote_cluster.py
```diff
@@ -1,143 +1,183 @@
...
+        following_license = await following_cluster_client.license.get()
+        source_license = await source_cluster_client.license.get()
+        source_license_type = source_license.get("license", {}).get("type")
+        following_license_type = following_license.get("license", {}).get("type")
 
-        if local_license_type not in required_licenses or remote_license_type not in required_licenses:
+        if source_license_type not in required_licenses or following_license_type not in required_licenses:
             raise BaseException(
-                f"Cannot use license type(s) [{local_license_type}, {remote_license_type}] "
+                f"Cannot use license type(s) [{source_license_type}, {following_license_type}] "
```
DJRickyB, line 124: In actuality here, one or the other could be permitted, so I think this needs to be split out. It would also be helpful to spit out the cluster name, even though I don't think it's likely we'll use this track in a scenario with a mixture of follower cluster licenses

DJRickyB (changes requested):

elastic/shared/runners/remote_cluster.py
b-deam: To be honest this was copied over from the original implementation, I agree it doesn't make sense to keep. Simplified in https://github.com/elastic/rally-tracks/pull/333/commits/9263e4a726ad28c47fdf52d253fa165f300403a7.

elastic/shared/runners/remote_cluster.py
b-deam: 👍 - addressed in https://github.com/elastic/rally-tracks/pull/333/commits/d30cf62b32be11beb4b1dc3d9d0de16062b6c83e

elastic/shared/runners/remote_cluster.py
DJRickyB: OK. I was gonna say, if we need the runner to be slower or more complicated I can come up with some pretty funny stuff to do

DJRickyB (approved):
LGTM

PR #334, opened 2022-10-17
Use list of queries for dense_vector track

jtibshirani:
This commit updates the dense_vector track to use a list of unique query
vectors during the search operations, instead of repeating a single query
vector. This helps ensure the timings are realistic. It's also a step towards
reporting recall in the benchmarks, since we'll be able to average over a set
of realistic query vectors.

Backport for #332.


DJRickyB:
Hi @jtibshirani, please feel free to backport at-will in this repository. We aim to automatically sync to the latest versioned branch (8.5 currently), then leave it to the committer to push to prior branches as they see fit.

jtibshirani:
@DJRickyB that makes sense, I just got caught up in something else and forgot to merge this! Doing that now.

PR #335, opened 2022-10-19
Remove legacy application-logs corpora

michaelbaamonde:
This PR removes the `application-logs` corpora from the `elastic/logs` track. The intention of the track going forward is to facilitate benchmarking of integrations (i.e. packages).  `application-logs` does not correspond to any official integration, so we'll remove it as a step in this direction.

This will likely have an effect on nightly benchmarks, so we'll plan on annotating once this is merged and takes effect.


michaelbaamonde:
Removing references to `application-logs*` in [this test](https://github.com/elastic/rally-tracks/pull/335/commits/2deff2b73deb1a3ef5e00b5d357d4e7f06d02ce4) doesn't seem worth it, nor does it really affect what's actually being tested.

DJRickyB (approved):
This LGTM, thanks

DJRickyB:
still approved, pending IT

Issue #337, opened 2022-10-24
Split the many-shards configurations to its own track


DJRickyB:
The Many Shards challenges in the `elastic/logs` tracks are beginning to bleed into the more core functionality of the track itself. It would be good to split these configurations to their own track for cleanliness' sake.

PR #303, opened 2022-08-24
[many-shards] Remove challenges and setup tasks from elastic/logs

dliappis:
In this commit we take care of some tech debt in the elastic/logs track
related to the many-shards* challenges.

Specifically, we remove the `base` and `full` challenges that are
unused and simplify the randomization of the base path of the snapshot
repository, which until now was done in a very indirect and difficult
to follow way, by including a Jinja2 file.

PR #338, opened 2022-10-24
Make base_path static

DJRickyB:
In #303 we added a safety net for running concurrent many shards snapshot challenges to the same bucket. Unfortunately the override applied for all `base_path`, including critically the non-many-shards challenges, rendering the base_path unconfigurable. I think we should still do this, but wait until #337 in order to do this as cleanly as possible.

This PR removes the base_path override by commenting with reference to the upstream issue we're dependent on.


michaelbaamonde (approved):
LGTM

PR #343, opened 2022-11-07
Separate many-shards snapshot repo settings to its own dict

DJRickyB:
In a similar spirit to #338, but working. Currently, all snapshot base paths in any `elastic/logs` challenge are set to `many-shards-{epoch}`, not just the many shards challenge. This change causes only the settings for that challenge to be mutated for concurrency, as the existing comment notes.


dliappis (approved):
LGTM thank you

PR #348, opened 2022-11-22
Add a time series aggregation to tsdb track

martijnvg:


tsdb/operations/default.json
```diff
@@ -86,4 +90,43 @@
...
+                },
+                "aggs": {
+                  "available_memory": {
+                    "min": {
+                      "field": "kubernetes.node.memory.available.bytes"
+                    }
+                  }
+                }
+              },
+              "min_available_memory": {
```
salvatore-campagna, line 123: I am not sure why we are also  including this pipeline aggregation. Do we expect to see changes in performance if we introduce some optimization for time_series aggregator?

salvatore-campagna:
LGTM apart understanding the reason why we need the min_bucket pipeline aggregation.

tsdb/operations/default.json
martijnvg: I added this because I think a pipeline aggregator is part of the usage pattern with time series aggregation.
All the time series buckets is an information overflow, getting min, max or derivative from all time series buckets will be useful informations to plot in a graph.

martijnvg: Or do you think we should just focus on the time series agg itself in the benchmark? So removing the pipeline agg and maybe even the date histogram here? So it is just the time_series agg?

martijnvg:
Closing in favour of #351

PR #349, opened 2022-11-22
Add detailed results to CCS queries 

b-deam:
With the merging of https://github.com/elastic/rally/pull/1618 and https://github.com/elastic/rally/pull/1619 we add the collection of shard meta-data for each search request with the intention of allowing us to visualise the number of shards hit and skipped by each query, an important metric for CCS behaviour.


b-deam:
👋  @DJRickyB - If you have some spare cycles this week would you mind reviewing?

PR #351, opened 2022-12-06
Include 'time_series' aggregation tests

### LINKED ISSUE #350: Extend the TSDB challenge to include time_series aggregations
We would like to run performance test for the `time_series` aggregation so that we are able to:
1. spot possible regressions with future changes
2. spot improvements we are planning to introduce (like pushing down filters from parent aggregations to child aggregations)

The typical use case is to have a `(auto)date_histogram` with a nested `time_series` aggregation or a `terms` aggregation with a nested `time_series` aggregation. At the moment we add the additional operations to the existing `append-no-conflicts` challenge to avoid reindexing data while running another challenge.

salvatore-campagna:
Include Rally tests for `time_series` aggregations.

Resolves #350

### DIFF
--- a/tsdb/operations/default.json
+++ b/tsdb/operations/default.json
@@ -115,6 +115,11 @@
                 }
               }
             }
+          },
+          "min_total_memory_usage_by_date": {
+            "min_bucket": {
+              "buckets_path": "ts>total_memory_usage_by_date"
+            }
           }
         }
       }
@@ -148,6 +153,11 @@
                   "field": "kubernetes.container.memory.usage.bytes"
                 }
               }
+            }
+          },
+          "min_total_memory_usage_by_date": {
+            "min_bucket": {
+              "buckets_path": "ts>total_memory_usage_by_date"
             }
           }
         }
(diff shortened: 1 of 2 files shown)


martijnvg (commented):
This looks good. I do think we need to add: https://github.com/elastic/rally-tracks/pull/348/files#diff-fd03625f346296b8349bb933b96aca0e9bd6d6144a954667c44f25e246b67e79R7

Otherwise we run into maximum number of allowed bucket limit while running the tsdb track.

tsdb/operations/default.json
```diff
@@ -1,89 +1,291 @@
...
+    "aggs": {
+      "by_date": {
+        "date_histogram": {
+          "field": "@timestamp",
+          "fixed_interval": "1h"
         },
-        "size": 1000
+        "aggs": {
+          "ts": {
+            "time_series": {
```
martijnvg, line 108: I think we should also add a pipeline aggregator here. Since the next change, we're trying to do is pushing down the pipeline aggregator computation to time series aggregator on data node. With that change the response of the time_series agg would change too, it will not return time series buckets but the result of the associated pipeline aggregations.

I think a sum pipeline aggregator is ok here.

martijnvg (approved):
I think the max bucket limit needs to be increased here, LGTM otherwise.

PR #354, opened 2022-12-12
ESQL: Add grouping by keyword fields operator

dnhatn:
This PR targets `esql` branch.


dnhatn:
`rally-tracks/it` fails because the `_esql` endpoint is not available yet.

PR #357, opened 2022-12-19
Add sort by field with and without can match challenge to http-logs and geonames tracks

javanna:
We recently found a regression that affected searches sorted by a keyword field (https://github.com/elastic/elasticsearch/pull/92026).

Given that we had no benchmarks for sorting by keyword, this commit adds the relevant operations to the http-logs and geonames tracks. We will want to also add similar challenges to the many-shards benchmarks, as the regressions we found can be seen with more than a couple of shards. This commit adds also specific challenges to verify the effect of https://github.com/elastic/elasticsearch/pull/51852 when a search is sorted by numeric or timestamp.

### DIFF
--- a/http_logs/challenges/common/default-schedule.json
+++ b/http_logs/challenges/common/default-schedule.json
@@ -173,6 +173,42 @@
           "target-throughput": 0.5
         },
         {
+          "operation": "desc_sort_timestamp_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "desc_sort_timestamp_no_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "sort_keyword_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "sort_keyword_no_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "sort_numeric_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "sort_numeric_no_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
           "name": "force-merge-1-seg",
           "operation": {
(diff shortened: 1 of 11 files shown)


http_logs/challenges/common/default-schedule.json
```diff
@@ -172,6 +172,42 @@
...
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
+        {
+          "operation": "sort_numeric_no_can_match_shortcut",
+          "warmup-iterations": 200,
+          "iterations": 100,
+          "target-throughput": 2
+        },
```
javanna, line 210: In order to make these part of many-shards (as these are particularly significant when run against many shards), I am not sure whether I need to add anything to some of the existing files under `elastic/logs/challenges`. Also, is it common to have some of these challenges not part of the ordinary http-logs benchmarks but only of the many-shards ones? That would be good too but I am not sure if it's an existing pattern .

javanna: I am not sure how to parse that :) I am adding my challenges to existing files. Does it mean that I also have to create an additional separate file for many-shards?

ebadyano: I think, you can create a file with all the new operations and schedules in one file and then include it similar to filed-caps. So under tasks create a new file sort_by_field.json for example and add all the new operations:
```
{
  "name": "sort_country_code_can_match_shortcut",
  "operation": {
      "operation-type": "search",
      "body": {
        "track_total_hits": false,
        "query": {
          "match": {
            "timezone": "America"
          }
        },
        "sort" : [
          {"country_code.raw" : "asc"}
        ]
      }
  },
  "warmup-iterations": 200,
  "iterations": 100,
  "clients": 2
},
...

and then include it to the 
```
{% include "tasks/sort_by_field.json" %}
```
The only ting we don't seem to be using target throughput and instead driving it as much as we can. So maybe don't specify the target throughout and instead specify the number of clients you want to run the queries?

javanna: Thanks, I've fallen silent because I am trying to gather context. I am catching up with Armin on how many shards benchmarks are run, and it seems that we currently have no search charts for the many shards challenge, which I was expecting I can add to. I will update my PR once I figured out how to proceed.

javanna:
@ebadyano I updated the description of the PR. It seems like more work is needed to add search challenges to the existing many shards benchmarks. I will look into that as a follow-up. For now, adding these bnenchmarks to geonames and http-logs is enough.

javanna:
@ebadyano could you review this please? It is hopefully ready.

javanna:
no worries, thanks @ebadyano

PR #362, opened 2023-01-05
Use `null` to reset cluster settings in CCS challenges

b-deam:
`-1` works for some cluster settings, but not others. `null` is a safer option.


dliappis (approved):
LGTM

b-deam:
@elasticmachine run rally-tracks/it

PR #364, opened 2023-01-13
Add a split corpora to TSDB to allow sequential indexing

pquentin:
As detailed here: https://github.com/elastic/rally/issues/1650

Can be tested with `esrally race --distribution-version=8.6.0 --track-path=$HOME/src/rally-tracks/tsdb --track-params="corpus:split,bulk_indexing_clients:16"`

### DIFF
--- a/tsdb/README.md
+++ b/tsdb/README.md
@@ -155,6 +155,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `index_mode` (default: time_series): Whether to make a standard index (`standard`) or time series index (`time_series`)
 * `codec` (default: default): The codec to use compressing the index. `default` uses more space and less cpu. `best_compression` uses less space and more cpu.
 * `ingest_mode` (default: index) Should be `data_stream` to benchmark ingesting into a tsdb data stream.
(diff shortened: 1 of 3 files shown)


tsdb/README.md
```diff
@@ -155,6 +155,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `index_mode` (default: time_series): Whether to make a standard index (`standard`) or time series index (`time_series`)
 * `codec` (default: default): The codec to use compressing the index. `default` uses more space and less cpu. `best_compression` uses less space and more cpu.
 * `ingest_mode` (default: index) Should be `data_stream` to benchmark ingesting into a tsdb data stream.
+* `corpus` (default: full) Should be `split` to use a corpus split in 16 to be used with 16 indexing clients and index mostly in @timestamp order.
```
martijnvg, line 158: Maybe name this `split16`? 

I also regularly run this on a different environment as well. And I think the default number of indexing clients (8) is good there. I was thinking of adding a variant for split (`split8`) in this scenario as well in a followup change.

pquentin: Thanks, this is a good idea. Hopefully this is only temporary until Rally learns how to do this itself though.

martijnvg: > Hopefully this is only temporary until Rally learns how to do this itself though.

👍 agreed

PR #365, opened 2023-01-17
[ESQL] Add more tests to the ESQL challenge

csoulios:
Add more tests to the ESQL challenge for the nyc taxis dataset. 

As a baseline for measuring performance, we are using the current aggregations framework, so that we can compare performance between ESQL and the existing functionality.

The following benchmarks are added: 

#### Filtered avg `passenger_count`
This query can be compared with the already existing non-filtered avg `passenger_count` 

```
from nyc_taxis | where total_amount > 1 and rate_code_id=="2" | stats avg(passenger_count), count(passenger_count)
```

#### Average `total_amount` grouped by `integer` field `passenger_count`
```
from nyc_taxis | stats avg(total_amount) by passenger_count | sort passenger_count
```

#### Average `total_amount` grouped by a low cardinality`keyword` field `rate_code_id`
```
from nyc_taxis | stats avg(total_amount) by rate_code_id | sort rate_code_id
```

#### Retrieve resultes sorted by timestamp field
```
from nyc_taxis | sort pickup_datetime desc | project pickup_datetime, dropoff_datetime, trip_distance| limit 1000
```

### Different partitioning methods

Each benchmark is executed with all combinations of different partitioning methods and consequently a different parallelization strategies:

- Current aggs framework through the `_search` endpoint. Query runs on a single thread as the index has only on 1 shard
- The `_esql` endpoint using partitioning by shard. This also runs on a single thread so we can compare performance with the existing aggs
- The `_esql` endpoint partitioned by segment. 
- The `_esql` endpoint using doc partitioning
- The `_esql` endpoint using shard partitioning, enabling task parallelization

### DIFF
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -828,7 +828,6 @@
           "warmup-iterations": 10,
           "iterations": 50
         },
-
         {
           "operation": "avg_amount_group_by_keyword_aggregation",
           "clients": 1,
@@ -859,7 +858,6 @@
           "warmup-iterations": 10,
           "iterations": 50
         }, 
-
         {
           "operation": "sort_by_ts_query",
           "clients": 1,
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -669,8 +669,8 @@
       ]
     },
     {
-      "name": "compute-engine",
-      "description": "Checks the performance of compute engine",
+      "name": "esql",
+      "description": "Performance benchmarks for the ESQL compute engine",
       "default": false,
       "schedule": [
         {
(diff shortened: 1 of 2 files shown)


nik9000:
Neat! It looks like there's a test error. I'm not really an expert there. But it otherwise does look good.

ChrisHegarty (approved):
LGTM

csoulios:
> Neat! It looks like there's a test error. I'm not really an expert there. But it otherwise does look good.

Test fails with the following error: 
```
16:07:06 [ERROR] Cannot race. Error in load generator [0]
16:07:06 	Cannot run task [avg_passenger_count_esql_shard_partitioning]: Request returned an error. Error type: transport, Description: Incorrect HTTP method for uri [/_esql] and method [POST], allowed: [HEAD, PUT, DELETE, GET] ({'error': 'Incorrect HTTP method for uri [/_esql] and method [POST], allowed: [HEAD, PUT, DELETE, GET]', 'status': 405})
```

Most probably the tests validates queries against the `main` branch of Elasticsearch, which does not support the `_esql` endpoint. (See https://elastic.slack.com/archives/C0JGTNK0S/p1674138873592169)

I will go ahead and merge this PR.

PR #367, opened 2023-01-20
Added six geo benchmarks, three geo_grid and three grid aggregations

craigtaverner:
Each of the set of three covers:

* geohash - tile the plane with geohash
* geotile - tile the plane with MVT-like tiles
* geohex - tile the plan with H3 hexagons (and pentagons)


iverase (commented):
LGTM, it would be nice to have some performance test for this operations which are currently missing.

PR #370, opened 2023-01-25
tsdb: Switch split16 to be a single file

pquentin:
That way, it will be split in 16, but each part will start from beginning to end.

### DIFF
--- a/tsdb/track.json
+++ b/tsdb/track.json
@@ -32,84 +32,9 @@
       {%- if corpus is defined and corpus == "split16" %}
         "documents": [
           {
-            "source-file": "documents-split-0.json.bz2",
-            "document-count": 7289607,
-            "uncompressed-bytes": 8252385790
-          },
-          {
-            "source-file": "documents-split-1.json.bz2",
-            "document-count": 7289607,
-            "uncompressed-bytes": 8252715218
-          },
-          {
-            "source-file": "documents-split-2.json.bz2",
-            "document-count": 7289606,
-            "uncompressed-bytes": 8252482193
-          },
-          {
-            "source-file": "documents-split-3.json.bz2",
-            "document-count": 7289606,
-            "uncompressed-bytes": 8253066083
-          },
-          {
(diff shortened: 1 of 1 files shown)


tsdb/track.json
```diff
@@ -32,92 +32,17 @@
...
-          {
-            "source-file": "documents-split-14.json.bz2",
-            "document-count": 7289606,
-            "uncompressed-bytes": 8252927958
-          },
-          {
-            "source-file": "documents-split-15.json.bz2",
-            "document-count": 7289606,
-            "uncompressed-bytes": 8252803345
+            "source-file": "documents-split16.json.bz2",
```
martijnvg, line 35: Question: how does Rally know how to split this file into 16 parts? (and each part having the correct date range)

martijnvg (approved):
Thanks @pquentin! I left a question for my own understanding.
LGTM

tsdb/track.json
pquentin: Splitting the file in 16 is the default behavior of Rally, provided we configure it to use 16 bulk indexing clients. I rearranged the original file to make sure that each of the 16 splits was starting from beginning to end. Given the previous split16 work, all I had to do was:

> cat documents-split-0.json documents-split-1.json documents-split-2.json documents-split-3.json documents-split-4.json documents-split-5.json documents-split-6.json documents-split-7.json documents-split-8.json documents-split-9.json documents-split-10.json documents-split-11.json documents-split-12.json documents-split-13.json documents-split-14.json documents-split-15.json > documents-split16.json

Does that make sense?

martijnvg: I see, and each of the `documents-split-N.json` files has an equal size, so that Rally knows exactly how to split the `documents-split16.json` file?

pquentin: Yes, each bulk indexing client will split and mmap the file independently based on its id (from 0 to 15) and the number of lines (which is why we have to provide document-count).

martijnvg: 👍

PR #371, opened 2023-01-25
Exclude sort by field operations when runtime fields are benchmarked in http_logs

javanna:
With #357 I introduced new challenges to benchmark sort by field with and without can_match enabled. These break the http_logs track when it's run with `runtime_fields:true` parameter because the fields I am sorting on are not mapped.

With this PR we don't execute such operations when in runtime_fields mode.


inqueue (approved):
LGTM, thanks!

PR #373, opened 2023-02-02
Add new track for tsdb based on k8s integration

martijnvg:
This track indexes generated k8s pod and k8s container data based on the k8s integration and then test performance of searches that are based on k8s integration visualisations. See README for more details.


tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,89 @@
+## TSDB k8s query track
+
+This track main goal is keeping track of the performance of common k8s integration search requests.
```
inqueue, line 3:
```suggestion
The main goal of the TSDB k8s query track is to measure the performance of common k8s integration search requests.
```

tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,89 @@
...
+* Average memory per container.
+* Average node cpu usage per container.
+* Average node memory usage per container.
+* Unique deployment count.
+* Percentile cpu usage per container.
+* Status per pod.
+* Transmitted network usage per pod.
+
+The search request is sourced from Kibana Lens in the k8s integration. For each of this search request a template is defined in operations.
+From each template three search requests are generated. First with last 15 minutes filter, then last 2 hours filter and finally with a last 24 hour interval. In case a `date_histogram` is used in the query template then each of this variation uses a different interval. A fixed interval of 30 seconds, 1 minute and 30 minutes repectively.
```
inqueue, line 21:
```suggestion
From each template, three search requests are generated. First with the last 15 minutes filter, then the last 2 hours filter, and finally with a last 24-hour interval. In case a `date_histogram` is used in the query template then each of these variations uses a different interval. A fixed interval of 30 seconds, 1 minute, and 30 minutes respectively.
```

tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+The k8s visualizations that run these queries don't run very often or under a high query load.
+Often these visualizations are loaded and then some time later re-loaded. This triggers the shards of the k8s pod and container data streams to go search-idle. However indexing always happen in the background. When shards become search-active again, a refresh need to occur as part of the search request. This track is designed to emulate this runtime behaviour.
```
inqueue, line 24:
```suggestion
Often these visualizations are loaded and then sometime later re-loaded. This triggers the shards of the k8s pod and container data streams to go search-idle. However, indexing always happens in the background. When shards become search-active again, a refresh needs to occur as part of the search request. This track is designed to emulate this runtime behaviour.
```

tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+This is done by concurrently indexing while the searches are ran. By lowering the `index.search.idle.after` setting from 30s (which is the default) to 1s. And force fully reducing the query load, so that one search gets executed every 3 seconds.
+
+When the data sets are updated. The `end_time` constant and `time_intervals` dictonary in `operations/default.json` file must be updated.
```
inqueue, line 28:
```suggestion
When the data sets are updated, the `end_time` constant and `time_intervals` dictionary in the `operations/default.json` file must be updated.
```

inqueue (changes requested):
This is looking good, thank you. I left some README suggestions and have not executed the track yet. 

Can you add track parameters for `number_of_shards` and `number_of_replicas`, if it makes sense to do so?

martijnvg:
Thanks for taking a look @inqueue. I updated the track.

> Can you add track parameters for number_of_shards and number_of_replicas, if it makes sense to do so?

I think it makes sense and I added the track params.

tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,91 @@
...
+From each template, three search requests are generated. First with the last 15 minutes filter, then the last 2 hours filter, and finally with a last 24-hour interval. In case a `date_histogram` is used in the query template then each of these variations uses a different interval. A fixed interval of 30 seconds, 1 minute, and 30 minutes respectively.
+
+The k8s visualizations that run these queries don't run very often or under a high query load.
+Often these visualizations are loaded and then sometime later re-loaded. This triggers the shards of the k8s pod and container data streams to go search-idle. However, indexing always happens in the background. When shards become search-active again, a refresh needs to occur as part of the search request. This track is designed to emulate this runtime behaviour.
+
+This is done by concurrently indexing while the searches are ran. By lowering the `index.search.idle.after` setting from 30s (which is the default) to 1s. And force fully reducing the query load, so that one search gets executed every 3 seconds.
```
dliappis, line 26: These indexing operations are the ones prefixed as `touch-` right (referenced from https://github.com/elastic/rally-tracks/pull/373/files#diff-18fad8cb6e0990bf9839cad4594f45f8cfde9d2a0a8cce9495614f74c5ab670cR1070-R1085)

For example I can see the following during an ongoing run

```
cpu_usage_per_pod_15_minutes,touch-pod-1-15_minutes
touch-pod-1-2_hours,cpu_usage_per_pod_2_hours
touch-pod-1-24_hours,cpu_usage_per_pod_24_hours
touch-pod-2-15_minutes,memory_usage_per_pod_15_minutes
memory_usage_per_pod_2_hours,touch-pod-2-2_hours
```

showing that `touch-...` is run in parallel.

I think we could clarify a bit more that these touch tasks do; they are special in the sense that `bulk-size` is defined to `1`.

We could also reword this paragraph; for example reading

> This is done by concurrently indexing while the searches are ran. By lowering the `index.search.idle.after` setting from 30s (which is the default) to 1s. And force fully reducing the query load, so that one search gets executed every 3 seconds.

I am not sure what is exactly meant here (maybe full stops should be removed?)

dliappis (commented):
This is a great and awesome amount of work, thank you!

I have been able to run this benchmark (it's ongoing, will report when it succeeds).

Left a suggestion to clarify a little about what the `touch-...` tasks are doing and also clarify some sentences in the readme.

tsdb_k8s_queries/README.md
martijnvg: @dliappis I've updated the readme. Is that paragraph now clearer?

tsdb_k8s_queries/README.md
```diff
@@ -0,0 +1,91 @@
...
+Often these visualizations are loaded and then sometime later re-loaded. This triggers the shards of the k8s pod and container data streams to go search-idle. However, indexing always happens in the background. When shards become search-active again, a refresh needs to occur as part of the search request. This track is designed to emulate this runtime behaviour.
+
+The way that this is emulated is by executing search and bulk operations concurrently. For each of the mentioned visualization there is a search operation and a corresponding bulk operation with a `touch-*` prefix as name. Both of the search and bulk operation are run concurrently using a `parallel` task. The bulk operations perform minimal indexing (1 index client and one document per bulk request) and reindexes the same curpus with a different `@timestamp` field using an ingest processor. This is sufficient so that the refresh isn't a noop. Also the data streams are configured with a `index.search.idle.after` of `1s`, so that shards of the backing indices become search-idle after 1 second of the last search request. Additionally, the search operations are configured to only execute a search every 4 seconds.
```
dliappis, line 26:
```suggestion
The way that this is emulated is by executing search and bulk operations concurrently. For each of the mentioned visualization there is a search operation and a corresponding bulk operation with a `touch-*` prefix as name. Both of the search and bulk operation are run concurrently using a `parallel` task. The bulk operations perform minimal indexing (1 index client and one document per bulk request) and reindexes the same corpus with a different `@timestamp` field using an ingest processor. This is sufficient so that the refresh isn't a noop. Also the data streams are configured with a `index.search.idle.after` of `1s`, so that shards of the backing indices become search-idle after 1 second of the last search request. Additionally, the search operations are configured to only execute a search every 4 seconds.
```

tsdb_k8s_queries/operations/default.json
```diff
@@ -0,0 +1,1085 @@
+{% set end_time = "2023-05-16T21:00:00.000Z" %}
+{% set time_intervals = {"15_minutes": ["30s", "2023-05-16T20:45:00.000Z"], "2_hours": ["1m", "2023-05-16T19:00:00.000Z"], "24_hours": ["30m", "2023-05-15T21:00:00.000Z"]} %}
```
dliappis, line 2: The above two seem to be used also in `challenges/default.json`; does it make more sense to move them, as well as the variables from `challenges/default.json`:

```
{% set search_iterations = 20 %} 
{% set search_warmup_iterations = 50 %}
{% set target_interval = 4 %}
```

to a centralized place, e.g. at the top of `track.json` (right after `{% import "rally.helpers" as rally with context %}`) for better discoverability? If you choose to do that, remember to update the references in `README.md`.

dliappis (approved):
LGTM

Left two comments, one for a minor typo, and another for a suggestion to move all the variables in `track.json` for better discoverability. Up to you if you want to implement the suggestion.

tsdb_k8s_queries/README.md
dliappis: @martijnvg it looks great, thank you.

tsdb_k8s_queries/operations/default.json
martijnvg: This makes sense to me. I will update this PR.

inqueue (approved):
LGTM, thanks!

tsdb_k8s_queries/operations/default.json
dliappis: > This makes sense to me. I will update this PR.

LGTM also after https://github.com/elastic/rally-tracks/pull/373/commits/5ea2206e02119b7f2f02a7fc803634082a99aec6

PR #375, opened 2023-02-02
tsdb: set default source mode to synthetic

pquentin:
Now that Elasticsearch [enforces synthetic source mode for TSDB](https://github.com/elastic/elasticsearch/pull/93380), let's allow all challenges to work with default values, which meant setting the default source mode to `synthetic`. In future work, we want to rely on Elasticsearch to set the default source mode.

Tested with:

 * esrally race --revision=9babcc9bb9a277174d43fa5c583268f642a83983 --track-path=$HOME/src/rally-tracks/tsdb --test-mode --on-error=abort --kill-running-processes
 * esrally race --revision=9babcc9bb9a277174d43fa5c583268f642a83983 --track-path=$HOME/src/rally-tracks/tsdb --test-mode --on-error=abort --kill-running-processes --track-params=index_mode:standard,source_mode:synthetic
 * esrally race --revision=9babcc9bb9a277174d43fa5c583268f642a83983 --track-path=$HOME/src/rally-tracks/tsdb --test-mode --on-error=abort --kill-running-processes --track-params=index_mode:standard,source_mode:stored

And I checked that the index settings were as expected. This should fix the rally-tracks compatibility tests.

### DIFF
--- a/tsdb/README.md
+++ b/tsdb/README.md
@@ -151,7 +151,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `number_of_replicas` (default: 0)
 * `number_of_shards` (default: 1)
 * `force_merge_max_num_segments` (default: unset): An integer specifying the max amount of segments the force-merge operation should use.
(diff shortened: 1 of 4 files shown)


ebadyano (approved):
lgtm

PR #376, opened 2023-02-02
Correct append-sorted-no-conflicts description

inqueue:
Correct the challenge description for `append-sorted-no-conflicts` to be clear the challenge leads to ___lower___ indexing throughput, not ___larger___. Lower indexing throughput with index sorting enabled is a known and [documented](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules-index-sorting.html) behavior, and it can also be observed in the nightlies.

image


ebadyano (approved):
Nice, thank you! lgtm

PR #380, opened 2023-02-13
Refactor elastic/logs runners to suit forthcoming client upgrade

b-deam:
In https://github.com/elastic/rally/pull/1669 we'll upgrade the ES client from 7.x to 8.x. As part of this upgrade some of the runners need to explicitly pass keyword arguments (unnamed args not supported), and some of the exception handling has changed.

This commit refactors the exception handling (luckily only one spot),
and converts the runners to use named arguments. Luckily it appears
there's only one place the method signature has changed, meaning 
all but one of the runner changes are BWC.

### DIFF
--- a/elastic/shared/runners/ilm.py
+++ b/elastic/shared/runners/ilm.py
@@ -35,7 +35,12 @@ async def create_ilm(es, params):
             name = os.path.splitext(os.path.basename(p))[0]
             with open(p, "r") as f:
                 policy = {"policy": json.load(f)["policy"]}  # only `policy` is allowed
-            await es.ilm.put_lifecycle(name, body=policy)
+            try:
+                await es.ilm.put_lifecycle(name=name, policy=policy)
+            # method signature changed in es client >8.x
+            except TypeError:
+                await es.ilm.put_lifecycle(name, body=policy)
+
             policy_num += 1
 
     return policy_num, "ops"
--- a/elastic/shared/runners/ilm.py
+++ b/elastic/shared/runners/ilm.py
@@ -35,10 +35,6 @@ async def create_ilm(es, params):
             name = os.path.splitext(os.path.basename(p))[0]
             with open(p, "r") as f:
                 policy = {"policy": json.load(f)["policy"]}  # only `policy` is allowed
-            try:
-                await es.ilm.put_lifecycle(name=name, policy=policy)
-            # method signature changed in es client >8.x
-            except TypeError:
                 await es.ilm.put_lifecycle(name, body=policy)
 
             policy_num += 1
--- a/elastic/shared/runners/ilm.py
+++ b/elastic/shared/runners/ilm.py
@@ -35,7 +35,7 @@ async def create_ilm(es, params):
             name = os.path.splitext(os.path.basename(p))[0]
             with open(p, "r") as f:
                 policy = {"policy": json.load(f)["policy"]}  # only `policy` is allowed
-                await es.ilm.put_lifecycle(name, body=policy)
+            await es.ilm.put_lifecycle(name, body=policy)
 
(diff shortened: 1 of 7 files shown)


elastic/shared/runners/remote_cluster.py
```diff
@@ -162,9 +173,9 @@ async def _follow_indices(self, params, source_cluster_name, source_cluster_clie
                 await following_cluster_client.ccr.follow(
                     index=index, wait_for_active_shards="1", body=follow_body, request_timeout=request_timeout
                 )
-            except ElasticsearchException as e:
+            except (ApiError, TransportError, ElasticsearchException) as e:
                 msg = f"Failed to follow index [{index}] from [{source_cluster_name}] on [{following_cluster_name}]; [{e}]"
-                raise BaseException(msg)
+                raise e(msg)
```
pquentin, line 178: I realize that this was working this way before this pull request, but by raising a new exception we're losing the traceback from the previous exception here. The best solution would be to use [add_note](https://docs.python.org/3/tutorial/errors.html#enriching-exceptions-with-notes) but it's only available starting with Python 3.11. The next best thing is to remove `; [{e}]` from `msg` and use [exception chaining](https://docs.python.org/3/tutorial/errors.html#exception-chaining):

```suggestion
                raise Exception(msg) from e
```

That way we keep the context from the previous exception while still raising our own exception. Now we have to think about the exception we want to raise. `BaseException` was a bad choice in my opinion as it prevents Ctrl-C from having any effect. `Exception` works but is very broad. Then there's  `RallyError` or one of its subclasses.

elastic/shared/runners/ilm.py
```diff
@@ -35,7 +35,12 @@ async def create_ilm(es, params):
             name = os.path.splitext(os.path.basename(p))[0]
             with open(p, "r") as f:
                 policy = {"policy": json.load(f)["policy"]}  # only `policy` is allowed
-            await es.ilm.put_lifecycle(name, body=policy)
+            try:
+                await es.ilm.put_lifecycle(name=name, policy=policy)
+            # method signature changed in es client >8.x
+            except TypeError:
+                await es.ilm.put_lifecycle(name, body=policy)
+
```
pquentin, line 43: Something I don't like about those try/except (including the one I suggested below!) is that we don't have a clear end-of-life policy for tracks. For how long do we have to support using Rally 2.7.0? On what branches?

One reason I'm saying this is that we can get avoid try/excepts completely here:

 * This specific change could be handled in Rally in the same way that we changed the signature of perform_request in https://github.com/elastic/rally/pull/1512
 * We could just catch `Exception` in remote_cluster.py. With exception chaining even if there's a TypeError we'll see it so making sure we only catch exceptions from the Python client becomes less important.

pquentin (changes requested):

elastic/shared/runners/ilm.py
b-deam: Agreed - I added the pre-requisite [BWC logic to Rally as part of the 8.x upgrade](https://github.com/elastic/rally/pull/1669/commits/b62d79362ee713f587ffee618234ac2475d10806) and I've reverted the track/ILM runner specific BWC logic in https://github.com/elastic/rally-tracks/pull/380/commits/52d036a1fdad81a3172a342640a03a6b23852e60

elastic/shared/runners/remote_cluster.py
b-deam: Thanks for the suggestions! I think I've covered all bases in https://github.com/elastic/rally-tracks/pull/380/commits/e1ba2306babf0ca7cc5091433b8252fafceeeb74.

pquentin (approved):
Thanks! LGTM

PR #382, opened 2023-03-08
ESQL: Add multi_terms aggs

csoulios:
Add benchmarks for mutli_terms aggs using both `multi_terms` agg and `composite` agg.

### `multi_terms` agg  

```
GET /nyc_taxis/_search
{
    "size": 0,
    "query": {
        "range": {
            "dropoff_datetime": {
                "gte": "2015-01-01 00:00:00",
                "lt": "2015-02-01 00:00:00"
            }
        }
    },
    "aggs": {
        "trip_type_payment_type_vendor_id": {
            "multi_terms": {
                "terms": [
                    {
                        "field": "trip_type"
                    },
                    {
                        "field": "payment_type"
                    },
                    {
                        "field": "vendor_id"
                    }
                ]
            }
        }
    }
}
```

### `composite` agg

```
GET /nyc_taxis/_search
{
    "size": 0,
    "query": {
        "range": {
            "dropoff_datetime": {
                "gte": "2015-01-01 00:00:00",
                "lt": "2015-02-01 00:00:00"
            }
        }
    },
    "aggs": {
        "vedor_payment": {
            "composite": {
                "sources": [
                    {
                        "trip_type": {
                            "terms": {
                                "field": "trip_type",
                                "order": "asc"
                            }
                        }
                    },
                    {
                        "payment_type": {
                            "terms": {
                                "field": "payment_type",
                                "order": "asc"
                            }
                        }
                    },
                    {
                        "vendor_id": {
                            "terms": {
                                "field": "vendor_id",
                                "order": "asc"
                            }
                        }
                    }
                ]
            }
        }
    }
}
```

### ESQL query

```
POST /_esql
{
    "query": "from nyc_taxis | where dropoff_datetime < \"2015-02-01 00:00:00\" AND dropoff_datetime >= \"2015-01-01 00:00:00\" | stats c = count(dropoff_datetime) by trip_type, payment_type, vendor_id | sort c desc",
    "pragma" : {
        "data_partitioning" : "doc"
    }
}
```

### Results

```
|                                        50th percentile latency |                                                multi_terms-keyword |   1703.33        |     ms |
|                                        90th percentile latency |                                                multi_terms-keyword |   1725.54        |     ms |
|                                       100th percentile latency |                                                multi_terms-keyword |   1735.71        |     ms |

|                                        50th percentile latency |                                            composite_terms-keyword |    518.589       |     ms |
|                                        90th percentile latency |                                            composite_terms-keyword |    526.252       |     ms |
[...]

PR #384, opened 2023-03-10
Disable failing esql challenge

pquentin:
It cannot work as it uses the _esql endpoint which is not public.


csoulios:
It looks like this PR was closed when I merged #382.

Thanks for opening #385 again

PR #386, opened 2023-03-18
Is the rally-tracks/it Jenkins job still triggered?

pquentin:
We've noticed in some pull requests that the rally-tracks/it Jenkins job still ran despite having been removed in https://github.com/elastic/rally-tracks/commit/f2f81810ee2dd94d024e9e2c706c963f7211fada and despite https://elasticsearch-ci.elastic.co/view/rally-tracks/ being gone. Let's see if this is still the case.


pquentin:
Looks like we're good. I think the issue was that the esql branch still had the Jenkins declaration?

PR #389, opened 2023-03-27
Revert "http_logs: use suitable date formater for @timestamp (#388)"

b-deam:
This reverts commit cffc8641055233485b9d1c9a773b1303c7c5313f due to cryptic nightly failures.

cc @inqueue

PR #388, opened 2023-03-20
http_logs: use only epoch_second format for @timestamp

inqueue:
This PR updates the index and runtime mapping for the HTTP Logs track. Set the `@timestamp` date format to `epoch_second` as it is in the corpus. For runtime fields, use the generic date format parser `strict_date_optional_time` to align with the runtime fields-specific corpus.

#### Sample

```json
{"@timestamp": 894034934, "clientip":"211.62.0.0", "request": "GET /images/dburton.jpg HTTP/1.0", "status": 200, "size": 12009}
```

PR #390, opened 2023-03-30
http_logs: set the date formatter for @timestamp

inqueue:
This PR reverts https://github.com/elastic/rally-tracks/pull/389, a revert of https://github.com/elastic/rally-tracks/pull/388, so it is a revert of a revert. In addition, it fixes the target index mapping for ISO 8601 when setting the `ingest_pipeline` track parameter to `grok`. Here is how it ends up:

* The track will use only ISO 8601 format via `strict_date_optional_time` for `@timestamp` for runtime fields or `ingest_pipeline: grok`. Both runtime fields and the ingest pipeline configurations will use the unparsed corpora which contain ISO 8601 formatted dates.
* The track will use epoch seconds (`epoch_second`) all other times; that is when executing against the parsed corpora.

### DIFF
--- a/http_logs/index.json
+++ b/http_logs/index.json
@@ -11,7 +11,11 @@
     },
     "properties": {
       "@timestamp": {
+        {%- if ingest_pipeline is defined and ingest_pipeline == "grok" %}
+        "format": "strict_date_optional_time",
+        {%- else %}
         "format": "epoch_second",
+        {%- endif %}
         "type": "date"
       },
       "message": {
--- a/http_logs/index.json
+++ b/http_logs/index.json
@@ -11,7 +11,7 @@
     },
     "properties": {
       "@timestamp": {
-        "format": "strict_date_optional_time||epoch_second",
+        "format": "epoch_second",
         "type": "date"
       },
       "message": {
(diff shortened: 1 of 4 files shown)


http_logs/index.json
```diff
@@ -11,7 +11,11 @@
     },
     "properties": {
       "@timestamp": {
-        "format": "strict_date_optional_time||epoch_second",
+        {%- if ingest_pipeline is defined and ingest_pipeline == "grok" %}
```
dliappis, line 14: shouldn't this conditional also include and `or runtime_fields is defined` as done in https://github.com/elastic/rally-tracks/blob/b1479ad8b466058f6c7348b9a1e875eba7a6447f/http_logs/track.json#L48?

dliappis (commented):
Left a question.

http_logs/index.json
inqueue: The [conditional](https://github.com/elastic/rally-tracks/blob/b1479ad8b466058f6c7348b9a1e875eba7a6447f/http_logs/track.json#L48) you referenced in track.json defines the path for which index mapping to use. The runtime-fields configuration does not use the `index.json` mapping and therefore is not included in the conditional.

dliappis (approved):
LGTM

http_logs/index.json
dliappis: Makes sense, thanks.

PR #395, opened 2023-04-17
First implementation of the behavioral_analytics track.

afoucret:


behavioral_analytics/track.py
```diff
@@ -0,0 +1,193 @@
...
+import logging
+import math
+import numbers
+import operator
+import re
+
+from esrally.track.params import *
+
+
+class EventBatchDataReader:
```
pquentin, line 11: Can you please add a comment mentioning that most of this was copy/pasted from Rally?

pquentin (changes requested):
This looks great! I have left a few comments.

behavioral_analytics/track.json
```diff
@@ -0,0 +1,29 @@
...
+    }
+  ],
+  "corpora": [
+    {
+      "name": "behavioral_analytics",
+      "base-url": "https://rally-tracks.elastic.co/behavioral_analytics",
+      "documents": [
+        {
+          "source-file": "events.json.bz2",
+          "document-count": 24514512
```
pquentin, line 18: Can you please add uncompressed-bytes too?

PR #403, opened 2023-06-02
Configurable number of replicas for dense_vector track

javanna:
This is important as running benchmarks against serverless requires at least 1 replica, hence this parameter needs to be at least configurable.


inqueue (approved):
Thanks!

Would you mind just sticking 

```
* `number_of_replicas` (default: 0)
```

in the Parameters section at the end of the README?

Otherwise, LGTM. No further review is needed.

PR #404, opened 2023-06-06
Dense vector to use polling mode for force merge call

javanna:
I experienced cases where force merge takes a long time and causes timeout issues, this should improve the situation.


dliappis (approved):
LGTM

PR #405, opened 2023-06-08
Weighted terms benchmarks for ELSER

davidkyle:
The Elasticsearch [ELSER](https://www.elastic.co/guide/en/machine-learning/master/ml-nlp-elser.html) model generates weighted tokens for use in the new [text_expansion](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-text-expansion-query.html) query. The goal of this track is to benchmark the performance of the query without the cost of evaluating ELSER. Of particular interest is the relative performance the compound weighted terms query used in `text_expansion` when Block Max WAND is enabled or disabled. In addition a `match` query is run against the same datasets to establish a baseline. 

The more tokens the ELSER model creates the more term in the compound weighted query. Another goals is to understand how increasing the number or tokens/terms affects latency. To test this queries with 10, 20, 30, 40 & 50 terms are used. This idea is extended to the match query but replace terms with words.

The dataset is 1 million documents from the [MS Marco](https://microsoft.github.io/msmarco/) passage retrieval collection, the body text has been tokenised using the ELSER model with the tokens written to the field `ml.tokens`.  

The compressed JSON for 1 million documents is 743MB.


elser/README.md
```diff
@@ -0,0 +1,133 @@
...
+The benchmark considers queries with different numbers of terms. Again these queries are created from the inference 
+response, the following script creates the operations in `operations/weighted-token-queries.json`. The first argument to the script 
+is a file containing the inference response. 
+
+```
+#!/bin/sh
+
+if [ -z $1 ]
+then 
+	echo "Missing input file: usgae ./gen_queries.sh input_json";
```
jimczi, line 40:
```suggestion
	echo "Missing input file: usage ./gen_queries.sh input_json"; 
```

Should we published the script in a separate file? There are other examples where the script (generally in python) is provided as a separate file.

elser/README.md
```diff
@@ -0,0 +1,133 @@
...
+done 
+
+```
+
+### Preparing the Dataset
+<!-- TODO Ingest MS Marco details. -->
+
+
+Once restored the MS Marco passage documents need to be run through ELSER to generate the tokens.
+This is done by reindexing through an ingest pipeline containing the inference processor.
```
jimczi, line 71: Not sure I understand this part, are you saying that the index is generated from a snapshot? Can we index the raw documents instead and add a version where the weighted terms are merged with the original documents?

elser/operations/weighted-token-queries.json
```diff
@@ -0,0 +1,2511 @@
+{
```
jimczi, line 1: I think this is a possible approach but I wonder if we should add a version or replace this one with a live one that generates the weighted terms at query time. That's not possible for the ingestion since it would take too much time on a single machine to index the full dataset but we should be able to infer this cost at query time. We can make it an option of the track as well. Anyway just ideas for follow ups and how we could use this benchmark to test the end to end experience.

elser/challenges/default.json
```diff
@@ -0,0 +1,179 @@
+{
```
jimczi, line 1: That's good default but I think we can have multiple challenges for the same track in a follow up. This one focus on ELSER but we can have a different one that focus more on BM25 and maybe another one that perform the BM25/ELSER rank?

elser/README.md
```diff
@@ -0,0 +1,133 @@
+## ELSER Track
```
jimczi, line 1: I wonder if we should focus on ELSER only in this track. I'd prefer to name it msmarco-passage and add multiple challenges to report (ELSER, BM25, dense vector, hybrid search, ...)? We can start with ELSER in this PR but leave room for others in follow ups.

elser/README.md
```diff
@@ -0,0 +1,133 @@
...
+Max WAND optimisation is controlled by the `track_total_hits` query parameter (set to `false` 
+to _enable_ Block Max WAND and `true` to _disable_).
+
+The `text_expansion` query rewrites to a compound weighted term query using the token weights
+retuned by evaluating the [ELSER](https://www.elastic.co/guide/en/machine-learning/master/ml-nlp-elser.html) model. For benchmarking, evaluating the NLP model is not necessary nor is it desired as the 
+purpose is to test the generated query. The query terms are pre-generated by the ELSER model
+and hard coded into the searches. 
+
+
+The BM25 match queries are present to establish a benchmark for comparrison with the
```
jimczi, line 14:
```suggestion
The BM25 match queries are present to establish a benchmark for comparison with the
```

elser/operations/bm25-queries.json
```diff
@@ -0,0 +1,120 @@
+{
```
jimczi, line 1: I wonder if we can use the 30k queries instead of a subset? We can generate the query for the queries file directly (randomly or not). Then we'll be able to report the recall using the annotations from the test dataset?

jimczi (commented):
Looks great, thanks @davidkyle 
I left some comments regarding the focus for the track and some for possible follow ups.
Regarding the document size, I think we should try to go with the full dataset (8M) from the start. That shouldn't be an issue for query and disk size, only for the ELSER inference at ingest time. The approach here is to provide them from an offline run so that should be fine too?

elser/challenges/default.json
jimczi: So I think we should rename the track and call it msmarco-passage with different challenges.

elser/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Preparing the Dataset
+<!-- TODO Ingest MS Marco details. -->
+
+
+Once restored the MS Marco passage documents need to be run through ELSER to generate the tokens.
+This is done by reindexing through an ingest pipeline containing the inference processor.
+
+
+Firt, put the pipeline
```
kderusso, line 74:
```suggestion
First, put the pipeline
```

elser/track.json
```diff
@@ -0,0 +1,31 @@
...
+      "name": "msmarco-passage-collection-elser",
+      "documents": [
+        {
+          "target-index": "msmarco-passage-collection-elser",
+          "source-file": "msmarco-passage-collection-elser-documents.json.bz2",
+          "document-count": 100000,
+          "compressed-bytes": 78074178,
+          "uncompressed-bytes": 254501902
+        }
+      ]
```
pquentin, line 22: Can you please add support for `--test-mode`? https://esrally.readthedocs.io/en/stable/adding_tracks.html#adding-support-for-test-mode

We use this in the CI to make sure your track does not have errors.

elser/README.md
davidkyle: I am trying describing the process of creating the test corpus with ELSER tokens. I've refined the README, hope it is clearer now.

elser/operations/weighted-token-queries.json
davidkyle: Generating the tokens at query time requires some orchestration as the ELSER model first needs to be downloaded and deployed then we have to call out to the _infer API. The tokens won't change for the current version of ELSER, hard coding the queries is good for now.

I agree we can be more sophisticated: perhaps pre-generate the tokens for thousands of queries, save those is a JSON file  and generate the weighted term query dynamically from Python. 

The AppEx QA team are testing ELSER end to end and benchmarking how long it takes to ingest their document set. We have good coverage in AppEx and the tooling they use is better suited end to end testing if we wanted to extend that coverage.

davidkyle:
I am in the process of ELSER-ising the msmarco passage dataset, I would appreciate some guidance to the size of the corpus.

The first 100,000 documents bzip compress to a 74MB file. Let's assume 1 million docs would therefore be 740MB, is that too big.

The full msmarco passage dataset is 8.8 million docs which could be a 6512MB bzip file. Is that at all practical?  The Stack overflow vector track uses 2 million docs which seems like a reasonable number

jimczi:
> The full msmarco passage dataset is 8.8 million docs which could be a 6512MB bzip file. Is that at all practical? The Stack overflow vector track uses 2 million docs which seems like a reasonable number

That's almost 8x the size of the dataset without ELSER. Since we're focusing on performance I guess that's ok to start with 1M and we can revise later. My assumption is that we could use the full 8M to check the recall but this is maybe too ambitious considering the expansion size at the moment.

elser/operations/bm25-queries.json
davidkyle: I've added a Python parameter source class to produce the queries from a JSON file containing the ELSER tokens. The file only contains 1 set of tokens for now (so it is always the same query) but it will be trivially easy to add more in future.

davidkyle:
@pquentin the `rally-tracks-compat` CI run has failed a few times due to cancellation after exceeding the 6 hour time limit. There are no timestamps in the log file, can you help me understand what is taking so long please?

pquentin:
@davidkyle You can see the timestamps like this:

image

However, I'm not sure what the issue is. I can run the tsdb_k8s_queries track in test mode on my laptop in a few minutes, not sure why it appears to block. I've added commits that will let us inspect the Rally logs (and fail this faster).

pquentin:
The Rally log does not help us as much as I had hoped: https://github.com/elastic/rally-tracks/pull/415#issuecomment-1597175139

msmarco-passage-elser/README.md
```diff
@@ -0,0 +1,120 @@
...
+            "results_field": "tokens"
+          }
+        }
+      }
+    }
+  ]
+}
+```
+
+Create the destination index with the rank_features mapping for the `ml.tokens` field.
```
benwtrent, line 49:
```suggestion
Create the destination index with the `rank_features` mapping for the `ml.tokens` field.
```

msmarco-passage-elser/README.md
```diff
@@ -0,0 +1,120 @@
...
+```
+PUT msmarco-passage-collection-elser
+{
+  "mappings": {
+    "properties": {
```
benwtrent, line 54:
```suggestion
    "_source": { "exclude": ["ml.tokens"]},
    "properties": {
```

msmarco-passage-elser/msmarco-passage-collection-elser.json
```diff
@@ -0,0 +1,42 @@
+{
+    "mappings": {
+        "properties": {
```
benwtrent, line 3:
```suggestion
        "_source": {"exclude": ["ml.tokens"]},
        "properties": {
```

benwtrent (approved):
I think we should disable source. But I think this is a good first pass on this track.

davidkyle:
The CI failure was because the index setting `number_of_replicas` was accidentally set to 1. This caused the  `tsdb_k8s_queries` track to timeout waiting for the cluster to go green which could never happen as the single node cannot satisfy replica shards leaving the cluster in a yellow state with unassigned shards.

pquentin:
Great detective work! Do you understand why this affected tsdb_k8s_queries instead of your track? Also, did you see Ben's review?

Issue #331, opened 2022-10-05
Move so_vector KNN operations to use _search endpoint


DJRickyB:
We should move the remaining KNN operations (those in the `so_vector` track) to use `search` operations. The only open question I have is if we should add a parameter source that supports randomness (like #330) or just convert the operations as they are.

CC @mayya-sharipova @jtibshirani

jtibshirani:
I think adding a randomization option would also be helpful (as long as it's not too tricky?) That way, we could test random vectors alongside filters or text queries.

PR #410, opened 2023-06-13
Update `so_vector` track to use _search and clean up the track

benwtrent:
This cleans up our `so_vector` track to make it more inline with `dense_vector`

I did not start doing random vector input or anything more than our current single vector. My focus here was tackling tech-debt, not adding functionality. 

I moved the query to its own file so that the track is easier to read.

closes: https://github.com/elastic/rally-tracks/issues/331


inqueue (approved):
LGTM, thank you

Tested track execution using `--test-mode` against an ESS deployment. All operations completed successfully without errors.

PR #413, opened 2023-06-15
Remove unnecessary wait-until-merges-finish

ebadyano:
wait-until-merges-finish is called before any indexing and force merge is called.


inqueue (approved):
LGTM, thanks!

PR #415, opened 2023-06-19
Fix small mistake in tsdb_k8s_queries benchmark.

martijnvg:
Changed data_stream.dataset filter in `status_per_pod_*` queries from `kubernetes.state_pod` to `kubernetes.pod`.

The `kubernetes.state_pod` value doesn't exist in the two data streams and therefor the this query currently never yields any result. This change should fix that.


pquentin (approved):
Thanks! LGTM. (I have not checked that the new value exists but I trust you.)

I'm very surprised that this passes btw.

PR #418, opened 2023-06-21
updating with new track with 10k pods

gizas:
Updating the specific rally with new corpora.

The new corpus includes 10k pods 

```bash
cat doc-ds-metrics-kubernetes.pod.json | jq '.kubernetes.pod.name' | sort -u > pod.txt

cat pod.txt| wc -l
   10000
```

### DIFF
--- a/tsdb_k8s_queries/track.json
+++ b/tsdb_k8s_queries/track.json
@@ -35,25 +35,25 @@
   "corpora": [
     {
       "name": "k8s-pod",
-      "base-url": "https://storage.googleapis.com/cloudnative-rally/rally2",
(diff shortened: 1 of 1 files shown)


martijnvg (approved):
LGTM

I ran the benchmark locally in test mode and in its entirety via esbench. 
No failures and I also checked whether some of the queries produce logical responses.

Note this will affect the the performance results, since there are less pods.
This means previous nightly performance results will no longer be comparable.
This is ok, because 8M+ pods and 16 containers is a little bit unrealistic :)
and supported 10k pods was a real performance goal. (along, 5K, 20K, 50K and 100k)

PR #419, opened 2023-06-21
[dense-vector] Load queries only once in memory

dliappis:
With this commit we use only one in-memory structure to store the dense_vector track queries (4MiB).

### DIFF
--- a/dense_vector/track.py
+++ b/dense_vector/track.py
@@ -22,6 +22,7 @@ class KnnParamSource:
             for line in file:
                 self._queries.append(json.loads(line))
         self._iters = 0
+        self.infinite = True
 
     def partition(self, partition_index, total_partitions):
         return self
@@ -70,6 +71,7 @@ class KnnRecallParamSource:
         self._cache = params.get("cache", False)
         self._params = params
         self._queries = []
+        self.infinite = True
 
         cwd = os.path.dirname(__file__)
         with open(os.path.join(cwd, "queries.json"), "r") as file:
--- a/dense_vector/track.py
+++ b/dense_vector/track.py
@@ -15,13 +15,13 @@ class KnnParamSource:
         self._cache = params.get("cache", False)
         self._exact_scan = params.get("exact", False)
         self._params = params
+        self._queries = []
 
         cwd = os.path.dirname(__file__)
         with open(os.path.join(cwd, "queries.json"), "r") as file:
-            lines = file.readlines()
-        self._queries = [json.loads(line) for line in lines]
+            for line in file:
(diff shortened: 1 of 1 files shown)


dense_vector/track.py
```diff
@@ -69,12 +69,13 @@ def __init__(self, track, params, **kwargs):
...
+        self._queries = []
 
         cwd = os.path.dirname(__file__)
         with open(os.path.join(cwd, "queries.json"), "r") as file:
-            lines = file.readlines()
-        self._queries = [json.loads(line) for line in lines]
-        self.infinite = True
+            # TODO take a random set of 20 queries instead
+            for line in file:
+                self._queries.append(json.loads(line))
```
benwtrent, line 78:
```suggestion
                self._queries.append(json.loads(line))
        self._queries = random.sample(self._queries, 20)
```
I *think* that is the way to do it. While we load the whole bunch of them in, its still pretty cheap IMO. Unless you want to do something like 

`random.sample(range(line_count), 20)` and then only append at each of those line indices.

However, I don't know how to get a line count from a file via python :)

benwtrent (commented):
so_vector also reads in queries, though for that race its only one (for now). Would be good to future proof it.

dense_vector/track.py
jdconrad: That's fine with me. I used the amount that was already used for the rest of the track. I do think 20 may see a lot of noise, though for recall numbers, I'd be more inclined to go with 200. We will have to see.

dliappis: > I _think_ that is the way to do it. While we load the whole bunch of them in, its still pretty cheap IMO. Unless you want to do something like
> 
> `random.sample(range(line_count), 20)` and then only append at each of those line indices.
> 
> However, I don't know how to get a line count from a file via python :)

Thanks @benwtrent and @jdconrad !

I was wondering, should we limit the scope of this PR to just address the memory waste and then address the TODO item in a follow up? 
My reasoning is that perhaps we could probably not need to read the entire file anyway, and as Jack mentioned we may want to tinker with the numbers (plus it'll probably create noise in the nightly charts, so we'd better experiment first).
With that in mind, I'd be leaning towards merging this tiny improvement first and then address the TODO in a follow up WDYT?

benwtrent: It isn't an emergency. @dliappis I am fine with this :)

dliappis: Thanks @benwtrent ! Could you LGTM this please, so that I can merge it?

benwtrent (approved):

PR #420, opened 2023-06-22
Make touch-* bulk operations heavier in tsdb k8s queries track

martijnvg:
The `touch-*` bulk operations are specifically used for indexing while searching, 
which allows us to catch performance issues we otherwise wouldn't see.

This change makes these tasks heavier so that more documents get indexed before
a search-idle shard gets refreshed due to a search. This should give more realistic refresh behaviour.

Before a low number of documents would be indexed between searches (around 5), with this
change thousands of documents with multiple clients should be indexed between searches.
There is usually 4 seconds between each search, this remains unchanged.

### DIFF
--- a/tsdb_k8s_queries/challenges/default.json
+++ b/tsdb_k8s_queries/challenges/default.json
@@ -93,7 +93,7 @@
               {
                 "name": "touch-pod-1-{{name}}",
                 "operation": "touch-pod-index",
-                "clients": {{bulk_indexing_clients | default(8)}}
+                "clients": {{concurrent_bulk_indexing_clients | default(6)}}
               },
               {
                 "operation": "cpu_usage_per_pod_{{name}}",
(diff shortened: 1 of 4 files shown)


martijnvg:
I test this change to the track out locally in test mode and via esbench.
Heavier concurrent indexing effect service time of the searches, 
but shows a more expensive refresh that we have seen in production too.
This we need to see in this track too as it affects the service time 
of searches too (as part of searching a search idle shard refresh occurs too).

tsdb_k8s_queries/challenges/default.json
```diff
@@ -24,6 +24,11 @@
                           "field": "@timestamp",
                           "value": {{'"{{_ingest.timestamp}}"'}}
                       }
+                  },
+                  {
+                    "script": {
+                      "source": "Random r  = new Random();ctx['kubernetes']['pod']['uid'] = 'value-' + r.nextLong();"
```
martijnvg, line 30: This is needed otherwise we endup indexing documents with same id too often.

PR #421, opened 2023-06-22
[dense_vector] Allow configurable index settings

dliappis:
This commit adds configurable index settings via the track parameter `index_settings`. This something typically supported by most standard tracks which was missing from `dense_vector`.


romseygeek (approved):
LGTM

PR #424, opened 2023-06-28
Use keyword arguments for Elasticsearch API methods

ebadyano:
Positional arguments can't be used with Elasticsearch API methods. Fixed a call where a keword was missing.


b-deam (approved):
I _knew_ I'd miss something - thanks for fixing this. 

Surprised this only came up now, but it looks like we only use the `create-datastream` runner in the `logging-snapshot-restore` challenge. 

LGTM.

PR #425, opened 2023-06-28
Add k8s_metrics track to benchmark refreshes during indexing

inqueue:
Introduce a new ~`serverless_k8s`~ `k8s_metrics` track designated (mostly) for Elastic Serverless deployments using generated Kubernetes pod metrics data. The track includes three documented challenges:

1. `append-no-conflicts-metrics-index-with-refresh` (default)
2. `append-no-conflicts-metrics-index-only`
3. `append-no-conflicts-metrics-with-fast-refresh`

#### Invocation Example

```sh
esrally race --track=k8s_metrics --pipeline=benchmark-only --client-options=$CLIENT_OPTS --telemetry=node-stats,shard-stats,data-stream-stats --telemetry-params="node-stats:node-stats-include-indices" --target-hosts=$HOST_IP --challenge=$CHALLENGE --kill-running-processes
```

The track is ultimately meant to benchmark search-after-write performance on serverless, though it does not perform search operations as it is just focused on benchmarking indexing throughput and how it is impacted by refresh. Search tasks can, and probably should, be added in a later iteration.

Closes ES-5547


serverless_k8s/README.md
```diff
@@ -0,0 +1,60 @@
...
+### `append-no-conflicts-metrics-with-fast-refresh`
+
+Index a metrics document corpus while indexing a small Kibana corpus to a separate fast refresh index. This challange simulates indexing to a smaller index with fast refresh enabled while concurrently bulk indexing to a larger data stream.
+
+#### Parameters
+
+* `bulk_indexing_clients` (default: `8`)
+* `ingest_percentage` (default: `100`)
+* `fast_refresh_clients` (default: `1`)
+* `fast_refresh_indexing_interval` (default: `15`)
```
Tim-Brooks, line 57: Is this indexing every 15 seconds?

That seems fine. But just checking.

serverless_k8s/README.md
```diff
@@ -0,0 +1,60 @@
+# Serverless Kubernetes Track
+Serverless Kubernetes is a track intended for benchmarking serverless Elasticsearch using a Kubernetes pod metrics dataset. A small corpus of Kibana object definitions is included for indexing with fast refresh enabled.
+
+## Track Parameters
+
+| Setting | Default | Description |
+| --- | --- | --- |
+| `bulk_indexing_clients` | `8` | The number of bulk indexing clients. |
```
Tim-Brooks, line 8: Is this what you normally run for tests vs. stateless?

Just curious because we are introducing quite a bit of "wait" time on the indexing pathway (waiting for object store flush) and I am wondering if that is a visible blocker with just 8 clients.

Tim-Brooks:
This looks nice. Do the bulk requests at the "small kibana like" index either force refresh or use wait_for on refresh?

If they don't, without search requests, search idle optimizations will kick in and that index won't actually refresh.

serverless_k8s/README.md
b-deam: Using a fixed number of clients simulates a closed system whereby the throughput is bound by the service time of each of the requests, but the value of 8 is just a default. 

We often increase or decrease it depending on the intention behind the benchmark, and if 8 isn't enough to 'induce' the behaviour/scenarios we want to benchmark then we can increase it at runtime (or the default here).

dliappis (commented):
A few initial small comments

serverless_k8s/README.md
inqueue: > Is this indexing every 15 seconds?
> 
> That seems fine. But just checking.

Correct, it is indexing to the fast refresh index every 15s. These parameters can be changed at runtime.

The fast refresh corpus contains just 21 documents. Would it make more sense to spread the indexing of these documents to 30s or 60s and end the operation once the fast refresh docs are finished? The current behavior is to not end the operation until the heavy indexing corpus has been fully indexed.

inqueue:
> This looks nice. Do the bulk requests at the "small kibana like" index either force refresh or use wait_for on refresh?
> 
> If they don't, without search requests, search idle optimizations will kick in and that index won't actually refresh.

The operation uses Rally's bulk runner which currently does not support `?refresh`. We should be able to change that in the very near term (https://github.com/inqueue/rally/commit/41b4c30a6041da574e741b23380707953a1eec38 is a POC).

inqueue:
This PR now depends on https://github.com/elastic/rally/pull/1743 and is ready for another review. Please see the [README](https://github.com/elastic/rally-tracks/blob/63609b30d6eb08f845ea7c0d0848f8b1a3c7454d/serverless_k8s/README.md) for challenge and parameter details.

serverless_k8s/README.md
```diff
@@ -0,0 +1,73 @@
+# Serverless Kubernetes Track
+Serverless Kubernetes is a track intended for benchmarking serverless Elasticsearch using a Kubernetes pod metrics dataset. A small corpus of Kibana object definitions is included for indexing with fast refresh enabled.
+
```
b-deam, line 3: I'm not sure how I feel about the name, will this ever be ran against stateful ES? Is it enough to just name it something related to Search After Write, or similar?

serverless_k8s/templates/rallybenchmark.json
```diff
@@ -0,0 +1,314 @@
+{
+  "index_patterns": [
+    "rallybenchmark"
```
b-deam, line 3: Should we change this name to reflect the benchmark name?

serverless_k8s/templates/metrics-kubernetes.pod.json
```diff
@@ -0,0 +1,38 @@
...
+      },
+      "managed_by": "fleet",
+      "managed": true
+    },
+    "data_stream": {
+      "hidden": false,
+      "allow_custom_routing": false
+    }
+  }
+}
```
b-deam, line 38: Nit: EOF.

b-deam (approved):
Only left a couple of high level comments, otherwise LGTM.

b-deam (approved):
LGTM 👍

PR #426, opened 2023-06-29
Remove data-url property

pquentin:
I believe this single line caused lot of people (me included) to think we were still using S3.


ebadyano (approved):
Nice catch! LGTM thanks!

dliappis (approved):
LGTM

PR #429, opened 2023-07-03
First version of the wikipedia dataset + creating the track.

afoucret:


wikipedia/track.py
```diff
@@ -0,0 +1,108 @@
...
+        csv_reader = csv.reader(queries_file)
+        next(csv_reader)
+        queries_with_probabilities = list(tuple(line) for line in csv_reader)
+
+        queries = [query for query, _ in queries_with_probabilities]
+        probabilities = [float(probability) for _, probability in queries_with_probabilities]
+
+        for query in random.choices(queries, weights=probabilities, k=k):
+            # remove special chars from the query + lowercase
+            yield re.sub("[^0-9a-zA-Z]+", " ", query).lower()
```
dliappis, line 28: It might make sense to compile the regexp once outside of the function even like `regexp=re.compile("[^0-9a-zA-Z]+")` and just `yield regexp.sub(" ", query).lower()` here from a performance view. (Some evidence of performance benefits [here](https://medium.com/geoblinktech/so-a-few-months-ago-i-had-to-search-the-quickest-way-to-apply-a-regular-expression-to-a-huge-c0883f8d4e4f)).

wikipedia/track.py
```diff
@@ -0,0 +1,108 @@
...
+def query_iterator(k: int) -> Iterator[str]:
+    with open(QUERIES_FILENAME) as queries_file:
+        csv_reader = csv.reader(queries_file)
+        next(csv_reader)
+        queries_with_probabilities = list(tuple(line) for line in csv_reader)
+
+        queries = [query for query, _ in queries_with_probabilities]
+        probabilities = [float(probability) for _, probability in queries_with_probabilities]
+
+        for query in random.choices(queries, weights=probabilities, k=k):
```
dliappis, line 26: I am concerned about repetability and consistency of results without predefining [random.seed](https://docs.python.org/3/library/random.html#random.seed).
In other tracks we have typically allowed [overriding the seed](https://github.com/elastic/rally-eventdata-track/blob/41cf1f94d955b7ac75454c0440b0b6ec610dee61/eventdata/parameter_sources/elasticlogs_bulk_source.py#L103).
Depending on what we need to achieve with this track, and whether it's important to have repeatable results with the same hardware (e.g. for nightly benchmarks) I'd recommend setting a default seed that can be overriden with a track parameter.

wikipedia/track.json
```diff
@@ -0,0 +1,34 @@
...
+      "body": "wikipedia-{{index_mapping_type | default("minimal")}}-mapping.json"
+    }
+  ],
+  "corpora": [
+    {
+      "name": "wikipedia",
+      "base-url": "https://rally-tracks.elastic.co/wikipedia",
+      "documents": [
+        {
+          "source-file": "documents.json.bz2",
```
dliappis, line 20: I've noticed that decompression with `pbzip2` only uses one cpu core. This leads to an unnecessary amount of time spent decompressing it on a powerful loaddriver.
Take a look at https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=bzip2#sharing-your-track-with-others on how to compress it with `pbzip2` and thus reduce decompression time.

wikipedia/README.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Generating the documents dataset
+
+To regenerate the dataset from scratch, first download and unzip an archive
+of Wikipedia dumps from [this link](https://dumps.wikimedia.org/enwiki/latest/enwiki-latest-pages-articles-multistream.xml.bz2) (~21GB).
+
+Then run this command:
+
+```bash
+python _tools/parse_documents.py <path_to_xml_file> | bzip2 --best > documents.json.bz2
```
dliappis, line 25:
```suggestion
python _tools/parse_documents.py <path_to_xml_file> | pbzip2 -9 -k -m2000 > documents.json.bz2
```

dliappis: see https://github.com/elastic/rally-tracks/pull/429/files#r1294578331 and https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=bzip2#sharing-your-track-with-others for more details

wikipedia/track.json
dliappis: @afoucret / @saikatsarkar056 I've uploaded a `pbzip2` compressed version of the corpora under the same bucket (it's called `rally-tracks/wikipedia/documents.json.pbzip2`), so feel free to swap those two (and update the compressed-bytes to `19064764394`).

wikipedia/track.py
afoucret: I did a change a while ago to compile the regexp only once in a global constant.

wikipedia/track.py
afoucret: Done. Thank you for the feedback.

wikipedia/track.py
```diff
@@ -0,0 +1,108 @@
...
+
+class QueryParamSource(QueryIteratorParamSource):
+    def __init__(self, track, params, **kwargs):
+        super().__init__(track, params, **kwargs)
+        self._index_name = params.get("index", track.indices[0].name if len(track.indices) == 1 else "_all")
+        self._cache = params.get("cache", True)
+
+    def params(self):
+        result = {
+            "body": {"query": {"query_string": {"query": next(self._queries_iterator), "default_field": self._params["search-fields"]}}},
```
jimczi, line 96: Let's use a multi-match query to avoid the reserved keywords (and, or, ...)?

jimczi (approved):
LGTM

wikipedia/track.py
afoucret: I will do this in a follow-up. We still have few run to do and I would prefer to keep the same query for now to get comparable results.

PR #432, opened 2023-07-10
Calculate the probability distribution of wikipedia title clickstream

saikatsarkar056:


saikatsarkar056:
@afoucret Here is my initial attempt to calculate the probability distribution of most frequently clicked wikipedia title for June, 2023. Please share your thoughts/feedbacks so that I can improve this work. Thank you for help.

We can discuss the following questions.

1. Should we consider the number of occurrences of the (referrer, resource) pair while calculating the probability distribution?
2. Should we consider last 3 or 6 months and combine them in the prob distribution calculation?

saikatsarkar056:
@afoucret This PR is ready for review. Can you please provide some feedbacks to this PR? Thank you.

wikipedia/_tools/parse_clicks.py
```diff
@@ -0,0 +1,106 @@
...
+
+# Add the console handler to the logger
+logger.addHandler(console_handler)
+
+
+class ClickStreamDist:
+    def __init__(self, year=2023, month=6, lang="en"):
+        self.filename = f"clickstream-{lang}wiki-{year}-{month:02d}.tsv.gz"
+        self.url = f"https://dumps.wikimedia.org/other/clickstream/{year:d}-{month:02d}/{self.filename}"
+        self.clickstream_output_file = "clickstream.tsv.gz"
```
afoucret, line 31: Did you consider using `~/.rally/benchmarks/data/wikipedia/` to store the downloaded file so the source directory is not polluted with data file?

I would also prefer to keep the original file name so we can have several dataset downloaded.

wikipedia/_tools/parse_clicks.py
```diff
@@ -0,0 +1,106 @@
...
+        writer.writerow(["Word", "Probability"])
+        writer.writerows(sorted_prob[:10000])
+
+    def clean_up(self):
+        logger.info("Cleaning-up...")
+
+        os.remove(self.clickstream_output_file)
+
+
+click_stream = ClickStreamDist(year=2023, month=6, lang="en")
```
afoucret, line 103: Do we want to use params from the CLI instead of hardcoded values?

wikipedia/_tools/parse_clicks.py
```diff
@@ -0,0 +1,106 @@
...
+        sorted_prob = sorted(word_prob.items(), key=lambda x: x[1], reverse=True)
+        return sorted_prob
+
+    def dump_probability_distribution(self, sorted_prob):
+        logger.info("Dumping probability distribution...")
+
+        # Dump the first 10000 entries to a different file
+        writer = csv.writer(sys.stdout)
+        writer.writerow(["Word", "Probability"])
+        writer.writerows(sorted_prob[:10000])
```
afoucret, line 95: I think the truncate is done too late. The sum of all probabilities is not equals to 1 anymore.

Instead, the queries list should be truncated before the probabilities are computed.

Your algorithm is:
- Compute probabilities
- Filter out value we do not want (Main Page)
- Sort by probabilities
- Keep the 10000 first result

What I think the algorithm should be:
- Filter out value we do not want (Main Page)
- Sort by count and keep only the first 10000 results
- Compute the probabilities of the latter step

afoucret (changes requested):

wikipedia/_tools/parse_clicks.py
saikatsarkar056: Updated

wikipedia/_tools/parse_clicks.py
saikatsarkar056: Updated

wikipedia/_tools/parse_clicks.py
saikatsarkar056: Updated

saikatsarkar056:
@afoucret Can you please review this PR again? Thank you.

afoucret:
I merged your code into my own branch #429.

You can start from my branch for your next task.

PR #434, opened 2023-07-11
Fix template issues in the k8s_metrics track

inqueue:
This PR addresses template loading issues in the k8_metrics track as neither component nor composable templates are loading in the current state.

* 9ec41a02d1eeb738e8ec68698c7590123ad2c07c corrects the Jinja template to address the failed loading of component templates.
* 4a2852b487bcd380c9f1cf1ca9a388e57697b56d specifies loading of a single composable template to get around an issue created by https://github.com/elastic/elasticsearch/pull/97417. It means the track challenge `append-no-conflicts-metrics-with-fast-refresh` will not work with `fast_refresh` enabled until a solution can be achieved.

### DIFF
--- a/k8s_metrics/challenges/default.json
+++ b/k8s_metrics/challenges/default.json
@@ -31,7 +31,8 @@
     {
       "name": "create-track-composable-templates",
       "operation": {
-        "operation-type": "create-composable-template"
+        "operation-type": "create-composable-template",
+        "template": "metrics-kubernetes.pod-template"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


ebadyano (approved):
LGTM, 

A side note: Does it at all makes sense to add `it` tests? we are skipping it for auto generated, but maybe it could work with something similar like we do for elastic/logs? not sure if it will work on a self-hosted env..at least to check if templates were loaded and things like that..

inqueue:
> LGTM,
> 
> A side note: Does it at all makes sense to add `it` tests? we are skipping it for auto generated, but maybe it could work with something similar like we do for elastic/logs? not sure if it will work on a self-hosted env..at least to check if templates were loaded and things like that..

Thank you, @ebadyano, very good point. I would like enable `it` tests in the near future.

PR #436, opened 2023-07-13
k8s_metrics track updates for fast refresh

inqueue:
This commit makes a few adjustments to the k8s_metrics track for benchmarking search-after-write:

* Use a system index instead of a data stream for the fast refresh due to https://github.com/elastic/elasticsearch/pull/97417.
* Increase the corpus size of the fast refresh index to extend the benchmark runtime.
* Add test mode support for `append-no-conflicts-metrics-with-fast-refresh`.
* Switch to using target throughput instead of target interval for fast refresh indexing.
* Update README.md to reflect the changes.


ebadyano (approved):
LGTM

PR #437, opened 2023-07-18
[Search Perf Benchmark][Rally track] Benchmark _search BM25

saikatsarkar056:
Issue: https://github.com/elastic/enterprise-search-team/issues/5079


wikipedia/track.json
```diff
@@ -0,0 +1,34 @@
...
+    {
+      "name": "wikipedia",
+      "base-url": "https://rally-tracks.elastic.co/wikipedia",
+      "documents": [
+        {
+          "source-file": "documents-1k.json.bz2",
+          "document-count": 1000,
+          "compressed-bytes": 10257016,
+          "uncompressed-bytes": 37065115
+        }
```
afoucret, line 24: I reverted this change that should not have been pushed.
I think you should have used esrally with the `--test-mode` argument instead of making this change.
This way it would have automatically downloaded the `documents-1k.json` file instead of the full.

afoucret (commented):
I have merged your code into #429 and did some refactoring on it (https://github.com/elastic/rally-tracks/pull/429/commits/e3550e4266f91bb889d5b269ae0d2769d9f1a261).

The main goal of the refactoring is to reduce code duplication between the BM25 and the search application param source.

afoucret:
Closing since the code has landed into #429

wikipedia/track.json
saikatsarkar056: That makes sense.

PR #438, opened 2023-07-24
Benchmark ELSER

maxhniebergall:
New rally track for benchmarking ELSER throughput with `allocations` and `threads per allocation` as primary parameters.

Running this track requires a deployment with an ML node with at least 4GB of memory. This track creates an index, downloads the ELSER model, deploys the ELSER model, creates a pipeline, and bulk ingests documents from ~the MSMARCO dataset~ a custom dataset derived from the BERT vocabulary.


msmarco-passage-elser-throughput/challenges/default.json
```diff
@@ -0,0 +1,108 @@
...
+                },
+                "inference_config": {
+                  "text_expansion": {
+                    "results_field": "tokens"
+                  }
+                }
+              }
+            }
+          ],
+          "on_failure": [
```
davidkyle, line 48: In general we recommend using `on_failure` as best practice but in this case it would be best to allow the error to propagate back to the caller and fail the bulk upload. If the model deployment was down for some reason the benchmark would run suspiciously fast and the cause can only be found be searching for an error in the index.

msmarco-passage-elser-throughput/track.py
```diff
@@ -0,0 +1,141 @@
...
+    while wait_time_per_cycle_seconds * try_count < max_wait_time_seconds:
+        try:
+            response = await es.ml.get_trained_models(model_id=elser_model_id, include="definition_status")
+            if is_model_fully_defined(response):
+                return True
+        except NotFoundError:
+            print("\nwaiting... try count:", try_count, end="")
+            await asyncio.sleep(wait_time_per_cycle_seconds)
+            try_count += 1
+    print()
```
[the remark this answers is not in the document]
maxhniebergall: Yes, in combination with the starting new line and the end="" in the print above, this empty print is needed to keep everything on its own line

msmarco-passage-elser-throughput/challenges/default.json
maxhniebergall: What you are saying about bulk running suspiciously fast in its failure mode is definitely true, but I tried removing `on_failure`, and it doesn't return an error. Do you know if there is something else I need to do to get it to throw an exception?

maxhniebergall: either way, the final report shows the failure rate on bulk as 100% when its failing, so I don't think its a big deal either way.

msmarco-passage-elser-throughput/track.py
```diff
@@ -0,0 +1,141 @@
...
+    number_of_allocations = params["number_of_allocations"]
+    threads_per_allocation = params["threads_per_allocation"]
+    queue_capacity = params["queue_capacity"]
+    try:
+        await es.ml.start_trained_model_deployment(
+            model_id=elser_model_id,
+            wait_for="fully_allocated",
+            number_of_allocations=number_of_allocations,
+            threads_per_allocation=threads_per_allocation,
+            queue_capacity=queue_capacity,
```
davidkyle, line 107: Model deployments have cache for inference results, if that cache is being hit then that is faster than inference. Can you please disable the cache just to be sure it is not affecting the results.

`cache_size='0b'`

https://www.elastic.co/guide/en/elasticsearch/reference/current/start-trained-model-deployment.html#start-trained-model-deployment-query-params

davidkyle (approved):
LGTM

Nice!

elser-speedtest/track.py
```diff
@@ -0,0 +1,152 @@
...
+
+async def create_elser_model(es, params):
+    await get_xpack_capabilities(es)
+    if await put_elser(es) == False:
+        return False
+    if await poll_for_elser_completion(es) == False:
+        return False
+
+    return True
+
```
davidkyle, line 147:
```suggestion


```

droberts195:
> This seems to indicate that the error is in a different rally track (nyc_taxis).

Presumably it's because there's only 1 ML node available and ELSER carries on running after the new track, using up all the capacity.

The solution is probably to make the new track stop ELSER as its last action so that the next track has some ML capacity available.

maxhniebergall:
@elasticmachine run buildkite/rally-tracks-it

maxhniebergall:
the most recent failure in _CI / rally-tracks-compat (pull_request)_ has the same error @jonathan-buttner found in [this investigation](https://github.com/elastic/ml-team/issues/1015#issuecomment-1692296334) (under _Findings_ expand _Licensing and other errors_)

`
Cannot run task [put elser model]: Request returned an error. Error type: api, Description: circuit_breaking_exception ({'error': {'root_cause': [{'type': 'circuit_breaking_exception', 'reason': '[parent] Data too large, data for [<http_request>] would be [4144368272/3.8gb], which is larger than the limit of [4080218931/3.7gb], real usage: [4144368272/3.8gb], new bytes reserved: [0/0b], usages [inflight_requests=0/0b, request=0/0b, fielddata=0/0b, eql_sequence=0/0b, model_inference=0/0b]', 'bytes_wanted': 4144368272, 'bytes_limit': 4080218931, 'durability': 'TRANSIENT'}], 'type': 'circuit_breaking_exception', 'reason': '[parent] Data too large, data for [<http_request>] would be [4144368272/3.8gb], which is larger than the limit of [4080218931/3.7gb], real usage: [4144368272/3.8gb], new bytes reserved: [0/0b], usages [inflight_requests=0/0b, request=0/0b, fielddata=0/0b, eql_sequence=0/0b, model_inference=0/0b]', 'bytes_wanted': 4144368272, 'bytes_limit': 4080218931, 'durability': 'TRANSIENT'}, 'status': 429})
`

maxhniebergall:
Got the same error again. I think we will have to wait until that investigation is resolved.

davidkyle:
One comment about the name `ELSER Speed Test Track` it's not clear what is being tested, inference speed or ingest speed. This is an ingest test so maybe rename it `ELSER Ingest Speed Test Track

PR #439, opened 2023-07-26
Add max_num_segments parameter for force_merge operation to all tracks

romseygeek:
Not all tracks currently allow configuring the maximum number of segments for
a force merge operation.  Merging to a single segment can be very expensive and
is not always necessary for a benchmark.  This commit makes the number of 
segments targetted by a force merge configurable in all tracks in the repo.

### DIFF
--- a/dense_vector/challenges/default.json
+++ b/dense_vector/challenges/default.json
@@ -86,7 +86,7 @@
       "operation": {
         "operation-type": "force-merge",
         "mode" : "polling",
-        "max-num-segments": 1,
+        "max-num-segments": {{ max_num_segments | default(1) }},
         "request-timeout": 7200,
         "include-in-reporting": true
       }
--- a/geopointshape/challenges/default.json
+++ b/geopointshape/challenges/default.json
@@ -37,7 +37,7 @@
         {
           "operation": {
             "operation-type": "force-merge",
-            "max-num-segments": 1,
+            "max-num-segments": {{ max_num_segments | default(1) }},
             "request-timeout": 7200
           }
         },
--- a/nested/challenges/default.json
+++ b/nested/challenges/default.json
@@ -38,6 +38,7 @@
         {
           "operation": {
             "operation-type": "force-merge",
+            "max-num-segments": {{ max_num_segments | default(1) }},
             "request-timeout": 7200
           }
         },
--- a/percolator/challenges/default.json
+++ b/percolator/challenges/default.json
@@ -38,6 +38,7 @@
         {
           "operation": {
             "operation-type": "force-merge",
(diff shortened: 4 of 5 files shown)


romseygeek:
The compat test fail appears to be a time out in the security track, which this PR does not touch.

dliappis (commented):
Thanks Alan!

Could you add documentation in the corresponding README files (e.g. for `dense_vector` [here](https://github.com/elastic/rally-tracks/tree/master/dense_vector#parameters)) about this new parameter?

romseygeek:
> Could you add documentation in the corresponding README files (e.g. for dense_vector [here](https://github.com/elastic/rally-tracks/tree/master/dense_vector#parameters)) about this new parameter?

done!

romseygeek:
Tests timed out again, I don't think that's related to my change though?

javanna (dismissed):

romseygeek:
I note that #438 is failing with the same error, which suggests to me that the timeout isn't a problem with this PR but an already existing issue?

dense_vector/README.md
```diff
@@ -29,4 +29,5 @@ This track accepts the following parameters with Rally 0.8.0+ using `--track-par
 * `bulk_size` (default: 5000)
 * `bulk_indexing_clients` (default: 1): Number of clients that issue bulk indexing requests.
 * `index_settings`: A list of index settings. Index settings defined elsewhere (e.g. `number_of_replicas`) need to be overridden explicitly.
+* `max_num_segments`: The number of segments to target when doing a force merge (default: -1)
```
inqueue, line 32: Should the default value be `1` instead of `-1`, to avoid impacting the current stateful nightly benchmarks? Otherwise, we will need to set the parameter in the nightlies to avoid impacting benchmarking results.

inqueue (changes requested):
Thanks! I left a comment.

dense_vector/README.md
romseygeek: The current parameter in the nightlies is unset, which is the same as `-1`.  This has the same effect as passing `1` for the volume of data we use in nightly benchmarks.  It might be worth changing nightlies to pass `1` explicitly here, because if for some reason one of the benchmarks ends up with more than 5Gb of data in it you may end up with more than one segment after force merging.

inqueue: Right, thank you. This looks good for pmc, percolator, and nested. dense_vector and geopointshape both have `max-num-segments: 1`. Can we use `1` for those defaults?

romseygeek: done!

inqueue (approved):
LGTM, thank you!

PR #441, opened 2023-07-27
k8s_metrics: add fast refresh indexing and search challenges

inqueue:
Add two challenges to the k8s_metrics track for search-after-write benchmarking: `fast-refresh-index-only` and `fast-refresh-index-with-search`.

* Add fast-refresh-index-only
* Add fast-refresh-index-with-search
* Use warmup-time-period
* Use target-interval
* Add queries

ES-6044 relates


ebadyano (approved):
LGTM

PR #442, opened 2023-07-31
Add ES logs to CI workflow artifact in GitHub

gbanasiak:
See https://github.com/actions/upload-artifact#upload-using-multiple-paths-and-exclusions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,5 +99,7 @@ jobs:
         if: always()
         with:
           name: rally-tracks-compat-logs
-          path: /home/runner/.rally/logs/
+          path: |
+            /home/runner/.rally/logs/
+            /home/runner/.rally/benchmarks/races/
           if-no-files-found: error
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -99,5 +99,7 @@ jobs:
         if: always()
         with:
           name: rally-tracks-compat-logs
-          path: /home/runner/.rally/logs/
+          path: |
+            /home/runner/.rally/logs/
+            /home/runner/.rally/benchmarks/races/
```
gareth-ellis, line 104: What about            
```suggestion
            /home/runner/.rally/benchmarks/races/*/*/*/*/rally-benchmark.log
```

That way we don't include the race.json files

gareth-ellis (approved):
LGTM

.github/workflows/ci.yml
gbanasiak: Addressed in https://github.com/elastic/rally-tracks/pull/442/commits/270eedc4b817273ec30a6b1ebec097ea0c1d407c.

PR #443, opened 2023-08-02
k8s_metrics: update fast-refresh indexing parameters

inqueue:
* Fast refresh parameter update for search after write.
* Update fast-refresh docs with new parameter values.


ebadyano (approved):
lgtm

PR #444, opened 2023-08-04
Cohere vector benchmarks

TattdCodeMonkey:
Built the `cohere_vector` rally track which will be used by the Search team to benchmark vector search with ~30M vectors. We will also be adding tracks for smaller data sets to get a comparison of performance at different dataset sizes.

Note:
 - The cohere vectors had to be normalized to unit-length to be ingested
 - I don't believe we support the correct similarity algorithm for cohere, so the benchmark is not expected to produce relevant results, just a performance test for now


cohere_vector/operations/default.json
```diff
@@ -0,0 +1,40 @@
...
+  "param-source": "knn-param-source",
+  "k": 100,
+  "num-candidates": 1000
+},
+{
+  "name": "script-score-query",
+  "operation-type": "search",
+  "param-source": "knn-param-source",
+  "k": 10,
+  "exact": true
```
jimczi, line 39: I don't think we need this one, it's super slow and we already know that it shouldn't be used on datasets of this size

jimczi (approved):
lgtm

PR #446, opened 2023-08-14
Add dummy buildkite pipeline

gareth-ellis:
Create empty buildkite workflow

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -43,7 +43,7 @@ spec:
     apiVersion: buildkite.elastic.dev/v1
     kind: Pipeline
     metadata:
-      name: IT
+      name: Rally Tracks - IT
     spec:
       pipeline_file: .buildkite/it/pipeline.yml
       repository: elastic/rally-tracks
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,55 @@
+# Declare a Backstage Component that represents rally-tracks
+---
+# yaml-language-server: $schema=https://json.schemastore.org/catalog-info.json
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rally-tracks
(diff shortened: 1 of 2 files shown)


catalog-info.yaml
```diff
@@ -0,0 +1,55 @@
...
+spec:
+  type: buildkite-pipeline
+  owner: group:es-perf
+  system: buildkite
+
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: IT
```
pquentin, line 46: The name will show up here: https://buildkite.com/elastic?team=es-perf so it would be nice to be consistent. Maybe "Rally Tracks - IT". That way the URL slug will be "rally-tracks-it" as already declared in the links section.

catalog-info.yaml
```diff
@@ -0,0 +1,55 @@
...
+    spec:
+      pipeline_file: .buildkite/it/pipeline.yml
+      provider_settings:
+        trigger_mode: none
```
pquentin, line 50: This means that no build at all will be triggered, see https://buildkite.com/docs/apis/rest-api/pipelines#provider-settings-properties for more details. You want to remove those two lines to trigger on pull requests and pushes to the default branch.

```suggestion
```

pquentin (commented):
Thanks! Can you please sure that this passes the Backstage entity validator? https://backstage.elastic.dev/entity-validation

catalog-info.yaml
gareth-ellis: thanks - renamed

pquentin:
Also can you please rename the pull request since it's not the initial commit of this repository. Maybe "Add dummy Buildkite IT pipeline"?

gareth-ellis:
The catalog-info.yaml file passes the validation on https://backstage.elastic.dev/entity-validation

pquentin (approved):
Thanks for iterating! LGTM.

PR #448, opened 2023-08-14
Add initial autoscaling benchmark

inqueue:
This commit adds a workload to exercise the performance and functionality of Serverless autoscaling using the nyc_taxis track. The challenge executes one or more iterations; i.e., the "step", starting with a "base" number of clients. There are three primary track parameters:

* `as_bulk_clients_base` (default: `8`): The base number of clients at the start of the benchmark.
* `as_bulk_step_count` (default: `3`): The number of times to execute bulk operations. Increasing the step increases the number of bulk clients for each iteration of the step.
* `as_reduce_factor` (default: `3`): An integer indicating how often to insert a reduced bulk operation into the loop. 

At the minimum, step and reduce should be equal to give the benchmark a chance to run at least one reduce iteration. For a heavier benchmark, configure the step count to 6, leaving the other parameters unchanged.

### DIFF
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -939,4 +939,12 @@
       "schedule": [
         {{ rally.collect(parts="common/update-aggs-only-schedule.json") }}
       ]
+    },
+    {
+      "name": "autoscale",
+      "description": "Bulk index with a varying number of clients",
+      "default": false,
+      "schedule": [
+        {{ rally.collect(parts="common/autoscale-schedule.json") }}
+      ]
     }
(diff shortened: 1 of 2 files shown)


nyc_taxis/challenges/default.json
```diff
@@ -939,4 +939,12 @@
       "schedule": [
         {{ rally.collect(parts="common/update-aggs-only-schedule.json") }}
       ]
+    },
```
ebadyano, line 942: Thank you for putting it to a separate file! It feels like nyc_taxis just keeps on growing :) it's less messy that way! I wonder if eventually we want to separate some other challenges out as well..

ebadyano (approved):
lgtm

PR #449, opened 2023-08-15
Support DLM/ILM conditional for auditbeat templates

b-deam:
https://github.com/elastic/rally-tracks/pull/406 introduced DLM support for the `elastic/logs` track, but the auditbeat templates still have ILM policies defined, which will fail when ran against serverless Elasticsearch.


pquentin (approved):
Thanks! LGTM.

It would have been nice to turn this into a single composable template as in #406 to make future changes easier, such as using build_flavor to determine the correct settings automatically, but I suppose it's not possible given the lifecycle name differs in ILM.

PR #455, opened 2023-09-08
Cohere bench improvements.

afoucret:


cohere_vector/challenges/default.json
```diff
@@ -44,16 +44,38 @@
...
+    },
+    {
+      "parallel": {
+        "tasks": [
+          {
+            "name": "parallel-documents-indexing-bulk",
+            "operation": "parallel-documents-indexing",
+            "clients": {{ parallel_indexing_bulk_clients | default(1) | int }},
+            "time-period": {{ parallel_indexing_time_period | default(1800) | int }},
+            "target-throughput": {{ parallel_indexing_bulk_target_throughput | default(1) | int }}
```
jimczi, line 68: I guess we also need the bulk size to derive the throughput in terms of docs per second?

cohere_vector/challenges/default.json
```diff
@@ -44,16 +44,38 @@
...
+      "iterations": {{ standalone_search_iterations | default(10000) | int }}
     },
     {
-      "name": "knn-search-100-1000",
+      "name": "standalone-search-knn-100-1000",
       "operation": "knn-search-100-1000",
       "warmup-iterations": 100,
-      "iterations": 1000
+      "clients": {{ standalone_search_clients | default(1) | int }},
+      "iterations": {{ standalone_search_iterations | default(10000) | int }}
```
jimczi, line 58: Can we also another task to test with 8 clients? Same for knn-10-100.

cohere_vector/index-vectors-only-mapping.json
```diff
@@ -0,0 +1,23 @@
+{
+  "settings": {
+    "index": {
+      "number_of_shards": {{number_of_shards | default(1)}},
+      "number_of_replicas": {{number_of_replicas | default(0)}}
```
jimczi, line 5: Should we add the preloading option too:

```
"store.preload": ["vec", "vex", "vem"]
```

jimczi (approved):
I left some comments on the missing bits that we discussed but the change looks great.

cohere_vector/challenges/default.json
afoucret: We now have two flavor of the knn searches`*single-client`and `*-multiple-clients` For multiple clients, I kept the number of clients configurable with `8` as a default.

cohere_vector/challenges/default.json
afoucret: This is right.

PR #458, opened 2023-09-21
Improve the wikipedia track.

afoucret:


wikipedia/challenges/default.json
```diff
@@ -18,7 +18,7 @@
     {
       "name": "index-documents",
       "operation": "index-documents",
-      "warmup-time-period": {{ bulk_warmup | default(40) | int }},
+      "warmup-time-period": {{ bulk_warmup | default(40) | int}},
```
afoucret, line 21:
```suggestion
      "warmup-time-period": {{bulk_warmup | default(40) | int}},
```

Remove space

wikipedia/track.py
```diff
@@ -73,16 +74,20 @@ def __init__(self, track, params, **kwargs):
...
+                "path": f"{SEARCH_APPLICATION_ROOT_ENDPOINT}/{self.search_application_params.name}/_search",
+                "body": {
+                    "params": {
+                        "query_string": query,
+                    },
                 },
-            },
-        }
+            }
+        except StopIteration:
```
saikatsarkar056, line 88: **Non-blocking question**: Do we see any exception for this part? Does this happen due to long running of the benchmark?

saikatsarkar056 (approved):

saikatsarkar056:
The change looks good to me. I have some non-blocking questions.

PR #460, opened 2023-09-28
Update elastic/logs

gareth-ellis:
Whilst preparing the synthetic test for serverless, I found a number of additional parameters that were useful, so they are included in this PR. 
My aim was to try and keep the previous behaviour, but allow more overrides if necessary

### DIFF
--- a/elastic/logs/track.json
+++ b/elastic/logs/track.json
@@ -8,9 +8,12 @@
 {% set num_corpus = 18 %}
 {% set p_bulk_indexing_clients = (bulk_indexing_clients | default(8)) %}
 {% set p_search_query_clients = (search_query_clients | default(1)) %}
+{% set p_runtime_indexing_clients = (runtime_indexing_clients | default(p_bulk_indexing_clients)) %}
+{% set p_ingest_percentage = (ingest_percentage | default(100)) %}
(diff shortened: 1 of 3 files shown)


elastic/logs/track.json
```diff
@@ -7,9 +7,13 @@
 {% set p_num_query_workflows = p_query_workflows | length %}
 {% set num_corpus = 18 %}
 {% set p_bulk_indexing_clients = (bulk_indexing_clients | default(8)) %}
+{% set p_search_query_clients = (search_query_clients | default(1)) %}
```
inqueue, line 10: What do you think about naming the parameter `search_clients` to be consistent with the [API](https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/examples.html#ex-search)?

inqueue (approved):
LGTM. We might consider documenting the new parameters at some point. I am good to leave it out for now.

gareth-ellis:
I think this is good to merge - the defaults are the same as before, so if the new parameters aren't supplied, then the previous parameters will be used (e.g bulk_indexing_clients is used if runtime_indexing_clients isn't provided)

inqueue (approved):
Thank you for documenting the new parameters. LGTMx2

PR #464, opened 2023-10-04
Add a new tsdb track to test _tsid hashing

salvatore-campagna:
Include a new TSDB track and dataset. The plan is to use it to test _tsid hashing and maybe impact of compression on dimension fields.


salvatore-campagna:
The CI tests are failing because the template and to be more precise the mappings include more than 21 dimension fields (`"time_series_dimension": "true"). This is a feature that Elasticsearch still does not support and it is the whole [urpose for having this dataset. The (contender) Elasticsearch implementation I have [here](https://github.com/elastic/elasticsearch/pull/98023) actually supports more than 21 dimensions. I will need this track dataset to evaluate impact of such feature.

tsdb_synth/README.md
```diff
@@ -0,0 +1,182 @@
...
+            "type": "double",
+            "time_series_metric": "gauge",
+        },
+        "key": {
+            "type": "keyword"
+        },
+        "val": {
+            "type": "integer"
+        },
+        "d_ufseibmjnkguynqfyclevvweebkoxqgvspbbngkvtrsegqtwuyel_001": {"type": "keyword", "time_series_dimension": "true"},
```
salvatore-campagna, line 41: I will probbaly replace the "true" value here with a Jinja "statement" which uses "false" as default and that we can override by means of `track.params`.

salvatore-campagna: Or maybe I can change the maximum number of allowed dimensions.

salvatore-campagna:
We can't just change the max number of dimensions because indexing would fail anyway due to too large values for dimensions and for the `_tsid` field.

tsdb_synth/track.json
```diff
@@ -0,0 +1,33 @@
...
+  "indices": [
+    {
+      "name": "tsdb_synth",
+      "body": "index.json"
+    }
+  ],
+  "corpora": [
+    {
+      "name": "tsdb_synth",
+      "base-url": "https://storage.googleapis.com/rally-tracks/tsdb_synth/",
```
gareth-ellis, line 14: Could this be https://rally-tracks.elastic.co/tsdb_synth/ ? They are the same place at the moment, but https://rally-tracks.elastic.co/tsdb_synth/ goes through the cdn

salvatore-campagna: Ok...so going through theCDN was probably the cause of the issue about the file not being there I guess...just a result of the delay in distributing the file across the CDN probably.

PR #466, opened 2023-10-06
Added some ESQL queries to `elastic/logs`

craigtaverner:
This dataset is of interest to ESQL particularly as we're targeting observability use cases. Currently ESQL is not mature enough to replace the workflows themselves, but can be used in the discover dashboard, and the queries chosen reflect possible usage in that dashboard, as well as investigating the impact of multiple grouping keys on similar aggregations.

Currently the intention is to include all new ESQL queries within the `logging-querying` challenge, since that generates and benchmarks the kind of data we're interested in. When running locally, using track.params that are based on the nightlies configuration, but scaled down by a factor of ten, we see the following:

* Data generation takes about 30min
* Normal logging-querying takes about 60 minutes
* New ESQL queries take about 4 minutes

So these queries do not impact the total benchmark run time by much at all.


elastic/logs/challenges/esql.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "esql",
```
craigtaverner, line 2: This challenge is currently only used for easier testing on an already populated dataset. The plan is to delete this before merging.

If, however, we want to keep ESQL in a separate challenge, a lot more work is required:

* Copy over the index setup and bulk indexing code from `logging-querying`
*. Copy the nightlies config for `logging-querying` to a new `elastic-logs-esql` config
* Simplify the above two setup for anything not really required for `esql`

And finally, such a step would mean duplicating the benchmark activity of generating the data, which is currently a very time-consuming and expensive step. It takes many hours, and the `eqsl` queries take only minutes to run. It is much more resource efficient to just run after `logging-querying` on the same challenge.

gbanasiak: I like adding `esql` at the end of `logging-querying`.

gbanasiak (approved):
LGTM.

The `<start_date, end_date>` interval was increased to work around https://github.com/elastic/elasticsearch/issues/100438. Once the issue gets addressed we can revert this increase to reduce test duration.

PR #468, opened 2023-10-11
Added ELSER V2 challenges for the elser-ingest-speedtest track

maxhniebergall:
New benchmarks for the new models introduced in 8.11.


davidkyle (approved):
LGTM

PR #469, opened 2023-10-13
Added four enrich indices for ENRICH ESQL benchmarks on nyc_taxis track

craigtaverner:
We benchmark a query that does not enrich (control) against four, each of which enriches 1 to 4 fields, to see the impact of enrich on ESQL performance.


gbanasiak (commented):
This looks good overall. I've left some minor comments inline. It wouldn't hurt to verify cluster health post `index-enrich-sources` operation. With default index settings (1 replica), cluster should be at least `yellow`.

gbanasiak:
@craigtaverner Please configure 0 replicas in all the new 4 indices. I hope this will solve current CI issue.

nyc_taxis/challenges/default.json
```diff
@@ -737,6 +749,103 @@
...
+          "operation": {
+            "operation-type": "cluster-health",
+            "request-params": {
+              "wait_for_status": "{{cluster_health | default('yellow')}}",
+              "wait_for_no_relocating_shards": "true"
+            },
+            "retry-until-success": true
+          },
+          "tags": ["setup"]
+        },
```
gbanasiak, line 767: Thank you for changing the number of replicas. We can now switch to `green` here.

gbanasiak: Trying to address via https://github.com/elastic/rally-tracks/pull/469/commits/0bc6526736ac770997ca6e22e53a14e8718c0701.

gbanasiak (approved):
LGTM. Thank you for your patience and all the iterations.

PR #470, opened 2023-10-13
Do not skip ESQL challenge in IT tests

gbanasiak:
With ESQL being part of `main` brach in Elasticsearch we should no longer skip `esql` challenge in IT tests.


gareth-ellis (approved):
LGTM

PR #471, opened 2023-10-16
Added group0 query for simple esql stats

craigtaverner:
This is a natural additional useful query to benchmark, a control query for the other stats-by grouping queries.


inqueue (approved):
LGTM

PR #472, opened 2023-10-17
Semantic search benchmark

jimczi:
This PR adds a new semantic search benchmark track, which leverages the dataset available at [https://microsoft.github.io/msmarco/Datasets#passage-ranking-dataset](microsoft/MSMARCO-Passage-Ranking). 
The primary goal of this track is to evaluate the performance of different semantic search strategies at scale without the need to assess ELSER or any inference model. 
To achieve this, we have expanded the dataset to encompass the 8.8 million passages and a thousand queries, enriching them with weighted tokens generated from the ELSER v2 model.


msmarco-passage-ranking/README.md
```diff
@@ -0,0 +1,292 @@
...
+### Example Query
+
+Queries are structured within a JSON array, where each individual object signifies a unique 'query' and its corresponding expansion achieved through ELSER v2, which is stored pre-computed in the 'query_expansion' field.:
+
+<details>
+  <summary><i>Example query object</i></summary>
+
+```json
+{
+    "query": ")what was the immediate impact of the success of the manhattan project?",
```
davidkyle, line 180: Is that a stray `)` at the beginning of the query?

davidkyle (approved):
LGTM

This supersedes the ELSER v1 track [msmarco-passage-elser](https://github.com/elastic/rally-tracks/tree/master/msmarco-passage-elser), to avoid confusion I think the v1 track can be removed now.

PR #465, opened 2023-10-06
Unblock geonames for public serverless

gbanasiak:
Prepares `geonames` track for public serverless use by:
* skipping index settings which are not available to serverless users,
* introducing an optional post-ingest sleep in place of a force merge followed by index stats loop.

Remarks:
* An alternative to `"index.requests.cache.enable": false` is to enable `request_cache=true` parameter in the queries which is proposed in https://github.com/elastic/rally/pull/1789.
* The `build_flavor != "serverless" or serverless_operator == true` expression is safe in older Rally versions where `build_flavor` and `serverless_operator` variables are undefined (unlike `build_flavor != "serverless" or serverless_operator`) . In such case the expression is evaluated as `true`.
* The tags in `{# ... #}` comments are meant to allow us an automated edit/removal in the future.

Test:

<details>

```
% esrally race --track-path=../rally-tracks/geonames --target-hosts=${ES_HOST}:443 --pipeline=benchmark-only --client-options="use_ssl:true,api_key:${ES_API_KEY}" --on-error=abort --test-mode --track-params="post_ingest_sleep:true,post_ingest_sleep_duration:5"

    ____        ____
   / __ \____ _/ / /_  __
  / /_/ / __ `/ / / / / /
 / _, _/ /_/ / / / /_/ /
/_/ |_|\__,_/_/_/\__, /
                /____/

[INFO] Race id is [48b2a64e-537f-453f-bf57-2be99b9e3b5b]
[INFO] Detected Elasticsearch Serverless mode with operator=[False].
[INFO] Excluding [check-cluster-health], [force-merge], [wait-until-merges-finish], [index-stats], [node-stats] as challenge [append-no-conflicts] is run on serverless.
[INFO] Racing on track [geonames], challenge [append-no-conflicts] and car ['external'] with version [oss].

Running delete-index                                                           [100% done]
Running create-index                                                           [100% done]
Running index-append                                                           [100% done]
Running refresh-after-index                                                    [100% done]
Running refresh-after-force-merge                                              [100% done]
Running post-ingest-sleep                                                      [100% done]
Running default                                                                [100% done]
Running term                                                                   [100% done]
Running phrase                                                                 [100% done]
Running country_agg_uncached                                                   [100% done]
Running country_agg_cached                                                     [100% done]
Running scroll                                                                 [100% done]
Running expression                                                             [100% done]
Running painless_static                                                        [100% done]
Running painless_dynamic                                                       [100% done]
Running decay_geo_gauss_function_score                                         [100% done]
[...]

PR #482, opened 2023-10-19
Unblock vector tracks for public serverless

gbanasiak:

Remarks:
* In `so_vector` an optional post-ingest pause is not introduced because original schedule in `index-and-search` challange executes search immediately after index/refresh. However, the second part of the schedule repeats all the searches after force merge to a single segment which is not available to public users. This part of the schedule gets skipped entirely.
* In `dense_vector` the pause is introduced _multiple_ times.


gareth-ellis (approved):
LGTM

ebadyano (approved):
lgtm, thank you!

PR #484, opened 2023-10-19
Simplify ESQL specific operations using new ESQL runner

craigtaverner:
Also removed timeout since it was previously only used for _search queries which often took much longer. And made the pragma body a one-liner for easier reading.

### DIFF
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -1111,55 +1111,6 @@
           "clients": 1,
           "warmup-iterations": 10,
           "iterations": 50
-        },
-        {
-          "operation": "esql_stats_enrich_1",
-          "tags": ["enrich","stats"],
-          "clients": 1,
-          "warmup-iterations": 2,
-          "iterations": 5
-        },
-        {
-          "operation": "esql_stats_enrich_2",
-          "tags": ["enrich","stats"],
-          "clients": 1,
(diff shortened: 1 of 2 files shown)


gbanasiak (commented):
The rewrite LGTM. Do we have a rough idea what is the duration of each request?

craigtaverner:
In my local testing of these queries on a 5% ingest using `--track-params="ingest_percentage:5"` I saw query times up to 5-6 seconds. If the performance is linear, which I would expect, then I presume we'll see 100-120s, right at the timeout threshold we're removing.

craigtaverner:
Additional local testing on 5% ingest shows that all existing `enrich` queries take something close to 30ms on my laptop, and the new `esql_stats_enrich_control` also takes that time. But the last 7 queries, the remaining `esql_stats_enrich_*` queries, all take between 3000 and 9000ms, 100x to 300x longer. The key difference between the control (fast) and the rest (slow) is that the control groups by an existing indexed field, while all the rest group by an enriched field. It would seem that this must prevent an optimization from kicking it.

So I've now removed all 7 slow queries from the schedule. I'll discuss with the ESQL team how best to handle this. One option is to bring back just one slow query, give it a very low iteration rate, and very long timeout. If this performance problem can be fixed, then we can bring back the remaining 6.

gbanasiak (approved):
Thank you! LGTM.

PR #487, opened 2023-10-19
Unblock k8s_metrics for public serverless

gbanasiak:

Remarks:
* not introducing post-ingest pause as this track does not include post-ingest search,
* fixed `append-no-conflicts-metrics-index-with-intermittent-refresh` challenge (`template` property in `create-component-template` instead of `crate-composable-template`),
* fast refresh challenges are not fixable for public use as they refer a system index.


inqueue (approved):
LGTM

PR #491, opened 2023-10-24
Added LIMIT 0 versions of queries to ESQL tasks in elastic/logs and NYC Taxis

craigtaverner:
These should run extremely fast and should measure only query parsing/planning, not runtime.


ebadyano (approved):
lgtm

PR #492, opened 2023-10-24
Unblock percolator, so, and sql tracks for public serverless

gbanasiak:

Remarks:
* I've added specific index selection in operations used by `so` frequent items challenge. That was required to avoid errors in serverless project which never comes empty, i.e. includes indices other than the indices added by the challenge.
* The `so` frequent items requires `post_ingest_sleep_duration` high enough for refresh to occur. The default refresh interval is 15s in serverless, the default post ingest sleep duration is 30s, so defaults work.
* The `sql` track does not support test mode, so I've tested with the following track params (based on https://github.com/elastic/rally-tracks/tree/master/sql#testing):
```
--track-params="ingest_percentage:1,query_percentage:2,post_ingest_sleep:true"
```


ebadyano (approved):
LGTM, thank you

PR #493, opened 2023-10-24
Unblock tsdb tracks for public serverless

gbanasiak:

Remarks:
* `downsample` challenge in `tsdb` track doesn't work because downsample operations get skipped by Rally (downsampling not available for public use yet, see [here](https://github.com/elastic/elasticsearch/blob/ecd13e3f118d34e151081cd1e3d432a609295dad/x-pack/plugin/downsample/src/main/java/org/elasticsearch/xpack/downsample/RestDownsampleAction.java#L25)),
* in `tsdb_k8s_queries` track index settings fall into 2 categories: the ones not available in serverless at all (e.g. `lifecycle.name`), and the ones that are available through operator only (e.g. `number_of_shards`).

### DIFF
--- a/tsdb/README.md
+++ b/tsdb/README.md
@@ -187,6 +187,8 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `codec` (default: default): The codec to use compressing the index. `default` uses more space and less cpu. `best_compression` uses less space and more cpu.
 * `ingest_mode` (default: index) Should be `data_stream` to benchmark ingesting into a tsdb data stream.
 * `corpus` (default: full) Should be `split16` to use a corpus split in 16 to be used with 16 indexing clients and index mostly in @timestamp order.
+* `post_ingest_sleep` (default: false): Whether to pause after ingest and prior to subsequent operations.
+* `post_ingest_sleep_duration` (default: 30): Sleep duration in seconds.
 
 ### License
 
--- a/tsdb/challenges/default.json
+++ b/tsdb/challenges/default.json
@@ -3,6 +3,7 @@
       "description": "Indexes the whole document corpus.",
       "default": true,
       "schedule": [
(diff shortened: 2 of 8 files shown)


tsdb/README.md
```diff
@@ -187,6 +187,8 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `codec` (default: default): The codec to use compressing the index. `default` uses more space and less cpu. `best_compression` uses less space and more cpu.
 * `ingest_mode` (default: index) Should be `data_stream` to benchmark ingesting into a tsdb data stream.
 * `corpus` (default: full) Should be `split16` to use a corpus split in 16 to be used with 16 indexing clients and index mostly in @timestamp order.
+* `post_ingest_sleep` (default: false): Whether to pause after ingest and prior to subsequent operations.
```
martijnvg, line 190: What is the reason behind this post ingest sleep?

martijnvg (approved):
One question. LGTM.

tsdb/README.md
gbanasiak: Pre-serverless benchmarks often force merge after write (sometimes down to a single segment) and then wait until the merge operation finishes by verifying indexing stats periodically. Example [here](https://github.com/elastic/rally-tracks/blob/9293eea47bc850c4041424d59c83352c03692b2e/tsdb/challenges/default.json#L75-L95). The intention of this sequence was increased repeatability of tests.

Neither the force merge, nor index stats are available in serverless (Rally auto-skips those when it detects serverless endpoint) so we're providing a naive but optional alternative in the form of a configurable sleep. We don't have empirical evidence it's helping (sleep long enough should) but since we're going through majority of tracks it looked reasonable to add something. As we are also adding unique markers, this can be later mass-edited, if need be.

martijnvg: Thanks for the explanation. Makes sense.

PR #496, opened 2023-10-26
Added _search versions of the ESQL count(*) queries

craigtaverner:
So we can compare performance between _search and ESQL for these cases.

### DIFF
--- a/elastic/logs/operations/esql.json
+++ b/elastic/logs/operations/esql.json
@@ -1,4 +1,24 @@
     {
+      "name": "enable_query_cache",
+      "include-in-reporting": false,
+      "operation-type": "raw-request",
+      "method": "PUT",
+      "path": "/logs-*/_settings",
+      "body": {
+        "index.requests.cache.enable": true
+      }
+    },
+    {
+      "name": "disable_query_cache",
+      "include-in-reporting": false,
+      "operation-type": "raw-request",
+      "method": "PUT",
+      "path": "/logs-*/_settings",
+      "body": {
+        "index.requests.cache.enable": false
+      }
+    },
+    {
       "name": "esql_basic_count_group_0",
       "operation-type": "esql",
       "query": "FROM logs-* | STATS count=count(*) | SORT count DESC | LIMIT 20"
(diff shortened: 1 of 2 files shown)


b-deam (approved):

craigtaverner:
These new _search queries cannot be benchmarked accurately without turning off the query cache. So this PR cannot be merged until we decide how to deal with that. Is it possible to dynamically turn off the cache, and then we can put that operation in the schedule after the normal query-logging workload, and before starting the ESQL benchmarks?

craigtaverner:
I added operations to disable the query cache, and re-enable it again afterwards. Local testing indicates this has the desired effect.

elastic/logs/operations/esql.json
```diff
@@ -1,3 +1,23 @@
+    {
+      "name": "enable_query_cache",
+      "include-in-reporting": false,
+      "operation-type": "raw-request",
```
b-deam, line 4: We have a dedicated `put-settings` runner for this purpose: https://esrally.readthedocs.io/en/latest/track.html?highlight=cluster%20settings#put-settings 

To complicate things further, the tracks need to be runnable against both flavors of ES (serverless + stateful), and our current conditional approach can be found here:
https://github.com/elastic/rally-tracks/blob/a1cf09881d0a97c19c9aa284f86ab2453d49fd15/geonames/challenges/default.json#L62-L70

craigtaverner: I looked at put-settings and it was only for cluster settings, not index settings, so it seemed like it could not handle this case.

craigtaverner: OK, I've pushed a commit that should do this. Could you verify?

craigtaverner: Local testing shows that this turns off _search queries for my local elasticsearch. I need a condition that only turns it off for serverless.

craigtaverner: The condition related to serverless would seem to be:

    if build_flavor != "serverless" or serverless_operator == true

However, the first predicate sounds like 'not serverless', while the second sounds like 'is serverless'. I'll push a commit with this condition, but would like confirmation on it's validity.

craigtaverner: This latest predicate passed local stateful tests. And the CI passed it and you gave it a positive review, so I'll merge it now. If we want to refine the predicate, we can make another PR.

gbanasiak: > However, the first predicate sounds like 'not serverless', while the second sounds like 'is serverless'. I'll push a commit with this condition, but would like confirmation on it's validity.

It should be read as "execute only if running against non-serverless or with serverless operator privileges".

PR #499, opened 2023-10-27
Remove two troublesome queries

craigtaverner:
There is a bug in ESQL only affecting these queries, so we need to remove them until the bug is fixed.


gbanasiak (approved):

gbanasiak:
Related ES issue: https://github.com/elastic/elasticsearch/issues/101455

PR #501, opened 2023-10-30
Increase timeout from 60m to 90m so that transforms can finish in time

przemekwitek:
After recent changes in search (most probably https://github.com/elastic/elasticsearch/pull/101230), the transform benchmark in rally slowed down, to the point of exceeding the timeout.
This PR increases the timeout from 60m to 90m in order to let the transforms finish and allow further performance investigations.


gbanasiak (approved):

przemekwitek:
run CI / rally-tracks-compat 3.11 (push)

PR #506, opened 2023-11-01
ESQL: Bring back two queries after ESQL bugfix

craigtaverner:
Revert "Remove two troublesome queries (#499)"
This reverts commit a1cf09881d0a97c19c9aa284f86ab2453d49fd15.

After Elasticsearch merged https://github.com/elastic/elasticsearch/pull/101496, these two queries should work again. I tested locally using an elasticsearch installed using `esrally` and it worked, so current builds have this fixed.

PR #507, opened 2023-11-01
Removed duplicate tasks after two PRs merged

craigtaverner:
Merging two PR's that included related content, resulted in duplicate tasks in the schedule. This removes the duplication.

The two PR's that should have conflicted, but did not are:

* https://github.com/elastic/rally-tracks/pull/506
* https://github.com/elastic/rally-tracks/pull/496


inqueue (approved):
Good catch, LGTM

PR #511, opened 2023-11-07
Remove ESQL shard partitioning, and re-introduce segment partitioning

craigtaverner:
Shard partitioning was considered worst case, therefor interesting to compare to best case (doc) and _search. However, we now are more interesting in the default behaviour (segment) since it is more apples-for-apples comparison to _search.


gareth-ellis (approved):
LGTM

PR #513, opened 2023-11-14
Updating k8s track with 8.12 templates and new UIds

gizas:
This Pr includes:
- The the [rally5](https://console.cloud.google.com/storage/browser/cloudnative-rally/rally5;tab=objects?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=elastic-observability&prefix=&forceOnObjectsSortingFiltering=false) folder of GKE bucket
- It adds the new 8.12 Templates for pod and container
- It includes the UID pod, container and node changes of [PR](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/111). In more details commits ([1](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/111/commits/fe73f697695aa0afaddb6e4dea69d0ca6f1db326), [2](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/111/commits/9ef0f5c6367522178b7081b8fa90fe84bcc023a8) and [3](https://github.com/elastic/elastic-integration-corpus-generator-tool/pull/111/commits/46f451f96be3e1c7c166a12bf099b04fa11dfd9d))


tsdb_k8s_queries/container-template.json
```diff
@@ -107,6 +115,9 @@
             "name": "kubernetes"
           }
         },
+        "_source": {
```
martijnvg, line 118: Not critical, but when is index.mode=time_series, then source mode is always synthetic. Do you know where this comes from? Ideally this shouldn't be defined.

I think the solutions should rely the default that is derived from `index.mode` index setting, which determined in Elasticsearch.

martijnvg (commented):
Thanks, this looks good. We need to test drive this with rally or esbench before we merge this. 


It looks like the `index.codec` index setting hasn't been removed from the templates, is this intended?

gizas:
|>It looks like the index.codec index setting hasn't been removed from the templates, is this intended?

No this should be deleted. I had it removed and then when I merged with master I reintroduced it. I removed it


For the _source filed now:
This is part of the mappings that I see in Kubernetes UI under each index. I removed it for now. 

> index.mode index setting, which determined in Elasticsearch.
Do we have such settings in rally side when we define Elasticsearch default values?

gizas:
@martijnvg FYI the  coponent template metrics-kubernetes.pod@package has the "_source" enabled

gareth-ellis:
It seems that the old (and new) location - rally4 / rally5 isn't publicly accessible - so those without access would not be able to run this track. Could we perhaps add them to the rally-tracks bucket or make them accessible in their current location?

gizas:
@gareth-ellis see below eg /rally5 files. Those are publicly accessible

Screenshot 2023-11-20 at 11 11 08 AM

gareth-ellis:
Interesting! I got a permission denied when I checked last week - but yes it seems to work now

martijnvg (approved):
I ran the benchmark and the results looked good.
LGTM - thanks for doing this!

PR #514, opened 2023-11-14
Add back ids for queries and documents in msmarco-passage-ranking

jimczi:
This commit adds the identifiers for documents and queries in the msmarco-passage-ranking track. We need the ids to compute MRR and other evaluation metrics from the judgment lists.


msmarco-passage-ranking/files.txt
```diff
@@ -1,2 +1,2 @@
-documents-1k.json.bz2
-documents.json.bz2
\ No newline at end of file
+documents2-1k.json.bz2
```
kderusso, line 1: Nitpick: `documents2` -> `documents_with_ids`? No strong feelings about it though.

kderusso (approved):
Thank you!

msmarco-passage-ranking/files.txt
jimczi: It's the version of the file  ;) I'll remove the previous one after the merge.

PR #515, opened 2023-11-15
Add new `search_clients` track parameter to TSDB track

b-deam:
Useful if users would like to apply extra load.


gbanasiak (approved):
LGTM

PR #516, opened 2023-11-15
Drop post-ingest-sleep from rampup

inqueue:
This commit removes the `post-ingest-sleep` operation from the `append-no-conflicts-rampup` challenge.


gbanasiak (approved):
LGTM

PR #519, opened 2023-11-21
Reduce Wikipedia default bulk size

b-deam:
As we've discussed, the Wikipedia track has been running into stability issues depending on the configuration, and one of the findings has highlighted that the default bulk size is a little too large and [can exceed Elasticsearch's default `http.max_content_length` value of `100mb`](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#http-settings).

This commit reduces the default bulk size by half.

Relates https://github.com/elastic/rally/pull/1805

cc @elastic/es-perf for awareness

### DIFF
--- a/wikipedia/operations/default.json
+++ b/wikipedia/operations/default.json
@@ -20,13 +20,13 @@
 {
   "name": "initial-documents-indexing",
   "operation-type": "bulk",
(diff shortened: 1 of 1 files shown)


afoucret (approved):

demjened (approved):
LGTM, thank you!

PR #520, opened 2023-11-21
ESQL: Load 500 rows

nik9000:
This adds a test that *just* loads 500 rows with ESQL. Against an index with *thousands* of fields. Those thousands of fields are pretty expensive for ESQL to load, much more expensive than _search. Because ESQL doesn't have any "just load the _source" style optimizations.

### DIFF
--- a/tsdb/challenges/default.json
+++ b/tsdb/challenges/default.json
@@ -164,6 +164,12 @@
           "iterations": 100
         }
         {% endif %}
+        ,{
+          "operation": "esql-fetch-500",
+          "warmup-iterations": 50,
+          "clients": {{ search_clients | default(1) | int }},
+          "iterations": 100
+        }
       ]
     },
     {
--- a/tsdb/operations/default.json
+++ b/tsdb/operations/default.json
@@ -322,4 +322,14 @@
       }
     }
   }
+},
+{
+  "name": "esql-fetch-500",
+  "description": "Grab 500 documents",
+  "operation-type": "esql",
+  {%- if ingest_mode is defined and ingest_mode == "data_stream" %}
+  "query": "FROM k8s | LIMIT 500"
+  {%- else %}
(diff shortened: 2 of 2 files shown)


nik9000:
How do I fix the ` CI / rally-tracks-compat 3.8 (pull_request)` build?

craigtaverner (approved):
Nice to have some ESQL in TSDB track. Do we need a review by anyone in TSDB?

craigtaverner:
> How do I fix the ` CI / rally-tracks-compat 3.8 (pull_request)` build?

The two `rally-tracks-compat` builds are the only ones that run the full track, and both fail with the same error:

```
Cannot run task [esql-fetch-500]: Request returned an error:

"fetching values from a text field [as.organization.name.text] is not supported because synthetic _source is enabled and we don't have a way to load the fields"
```

Is this a conflict between TSDB and ESQL? One needs synthetic source and the other can't handle it?

nik9000:
https://github.com/elastic/elasticsearch/pull/102490 should fix these.

nik9000:
elasticmachine, test this please

nik9000:
> [elastic/elasticsearch#102490](https://github.com/elastic/elasticsearch/pull/102490) should fix these.

I've merged that so the tests should pass.

nik9000:
@elasticmachine, update branch

nik9000:
Looks like this is still failing, but the bug in ES is fixed. I'll bet it needs a snapshot or something.

nik9000:
Neat! It closes as soon as I open it. Weird! I'll remake.

pmpailis:
The issue with automatically closing should be resolved now - reopened :)

nik9000:
> The issue with automatically closing should be resolved now - reopened :)

Thanks! I think I want to merge this now!

PR #532, opened 2023-12-05
Derive project name from Buildkite metadata

gbanasiak:
Derives serverless project name from Buildkite metadata.


ebadyano (approved):
LGTM

PR #534, opened 2023-12-06
Add another model for semantic search in msmarco-passage-ranking

jimczi:
This change adds the text expansion for all docs and queries using the [SPLADE CoCondenser EnsembleDistil](https://huggingface.co/naver/splade-cocondenser-ensembledistil) model. This data was computed for the [big-ann-benchmark](https://github.com/harsha-simhadri/big-ann-benchmarks) so we're using it here to get the performance of the SPLADE model in this track. Two new operations have been added to test this model independently.


kderusso (approved):
LGTM with one typo fix

PR #535, opened 2023-12-06
Adjust index append warmup time period in Geonames

gbanasiak:
The `index-append` warmup time period of 120s is too big.
* In case of 3-node benchmark it might lead to throughput not being reported at all as only normal samples contribute to throughput summary (result).
* In other benchmarks throughput is reported, but it's calculated based on a small amount of samples (%-wise).

As shown in the following chart, the warmup time period of 60s should work better.

Screenshot 2023-12-06 at 19 38 34


gbanasiak:
This requires more investigation, so I'm converting to draft.
* Perhaps there's a better way, e.g. reducing the number of clients to increase the overall ingest time.
* We should also understand what is happening at the end of the run when throughput drops. Is it when some of the workers are already done? Do we need cool-down period?
* The dip around 50 s into the test is also interesting, although it should be filtered through warmup interval regardless.

(last 90 `geonames` runs, 3-node throughput samples)

Screenshot 2023-12-07 at 11 20 32

ebadyano (approved):
lgtm, thank you

ebadyano:
I am good with the suggestion to lower the warm up period. What if we change it to something like 70s or 80s to make sure we definitely passed the weird dip in the throughput?

PR #539, opened 2023-12-11
Update hatch configuration

gbanasiak:
Updates hatch version and limits hatchling version to 1.18.0 as a workaround for https://github.com/pypa/hatch/issues/1113.


gareth-ellis (approved):
LGTM

PR #545, opened 2024-01-04
Added es|ql queries to tsdb k8s query track.

martijnvg:
This PR adds es|ql queries to the tsdb k8s track based on the following already defined searches: `esql_cpu_usage_per_pod_*`, `memory_usage_per_pod_*`, `average_container_cpu_*`, `average_container_memory_usage_*`, `cpu_usage_per_container_*`, `memory_usage_per_container_*`, `unique_deployment_count_*` and `percentile_cpu_usage_per_container_*`.

There is one caveat with the newly defined es|ql queries and that is that es|ql queries group by pod/container and then time interval. Whereas the searches also group by just pod/container. For example `cpu_usage_per_pod_15_minutes` computes average cpu usage by pod and 30 second interval and by pod, whereas the esql variant only computes cpu usage by pod and 30 second interval. In es|ql grouping currently can only be done by one set of aggregation keys. Maybe supporting multiple sets of aggregation kets will be supported in the future by the inline stats command.

Additionally the following searches couldn't be ported to esql queries:

- The `status_per_pod_*` searches, because there is no equivalent yet for the `top_metrics` aggregation in es|ql.
- The `tx_network_usage_per_pod_*` searches, because this search uses an average aggregation on a counter field and es|ql doesn't allow stats operations to be performed yet on counter fields.

### DIFF
--- a/tsdb_k8s_queries/challenges/default.json
+++ b/tsdb_k8s_queries/challenges/default.json
@@ -130,6 +130,28 @@
         {% for name, info in time_intervals.items() %}
         {
           "parallel": {
+            "completed-by": "esql_cpu_usage_per_pod_{{name}}",
+            "clients": 2,
+            "tasks": [
+              {
+                "name": "esql_touch-pod-1-{{name}}",
+                "operation": "touch-pod-index",
+                "clients": {{touch_bulk_indexing_clients}}
+              },
+              {
+                "operation": "esql_cpu_usage_per_pod_{{name}}",
+                "warmup-iterations": {{search_warmup_iterations}},
+                "iterations": {{search_iterations}},
+                "target-interval": {{target_interval}},
+                "clients": 1
+              }
+            ]
+          }
+        },
+        {%- endfor %}
+        {% for name, info in time_intervals.items() %}
+        {
+          "parallel": {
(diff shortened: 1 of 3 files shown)


gareth-ellis (commented):
Hello,

I have a couple of comments:
- By adding es|ql after each of the existing queries, the track will no longer be strictly comparable with previous runs - have you considered adding the es|ql portion at the end of the track or even as a new challenge? 
- Have you run this already somewhere, and do you know the runtime? The track currently takes around 3 hours to run, so I would guess it may be close to double that after this change, given most queries are run both as previously and then as es|ql.

martijnvg:
Thanks for taking a look @gareth-ellis. I just checked the rally log when I was running this track with new changes in esbench. And now it takes roughly 5 hours to complete this track. So I think the benchmark time is getting out of hand. I like the idea of bundling the esql queries in a new track. I will update the PR to do this.

martijnvg:
I've updated the PR and moved the esql queries into a new challenge (instead of a new track). This way a number of things get reused, which it think makes sense (esql queries use the same corpora.

gareth-ellis (approved):
LGTM

martijnvg:
thanks @gareth-ellis!

PR #549, opened 2024-01-11
Fix variable overwriting in ELSER ingest speed test

maxhniebergall:
Closes [this issue](https://github.com/elastic/elasticsearch-benchmarks/issues/1905#issuecomment-1887658670), found by @gbanasiak. The elser-ingest-speedtest rally track has a bug which sometimes causes variables to be erroneously overwritten. This occurs because all challenges are assembled by rally into a single document with a single namespace for all variables. Setting variable values thus effects not only one challenge, but all challenges in a track. This change fixes this bug by always never overwriting the variable values which are parameters for the challenge, but rather creating new variables with prefixes of `p_` for setting parameters and `l_` for loops.


esbenchmachine:
@maxhniebergall

Backporting entails: 

Thank you!

PR #550, opened 2024-01-12
feat: add apm rally track

kyungeunni:
- ~blocked by https://github.com/elastic/rally-teams/pull/83~
- ~can be merged once the reference corpora is published~ update: done
---
Add a basic rally-track for APM. It's moved over from [apm-server repo](https://github.com/elastic/apm-server/tree/main/testing/rally).
It works with Elasticsearch version 8.12 or later, with `x-pack/apm-data` enabled which means the `x-pack-apm` mixin should be used to run it successfully.

Sample command:
(**note**: it won't work until the mentioned blockers are addressed, please see _How to test in development mode_ below)
```
esrally race  --track=apm --car="defaults,x-pack-apm"
```
#### 1. configure _team_
To use the `x-pack/apm-data` plugin, we need to use the specific car/mixin. The [PR](https://github.com/elastic/rally-teams/pull/83) is open, but until it's merged, we can use the custom repository, https://github.com/kyungeunni/rally-teams.

you should add it in `~/.rally/rally.ini`.
Open the file in your editor of choice and add the following line in the section `teams`:
```
apm.url = https://github.com/kyungeunni/rally-teams
```
It should look like this:
```
[teams]
default.url = https://github.com/elastic/rally-teams
apm.url = https://github.com/kyungeunni/rally-teams
```

make sure the car is available by running this command:
❯ esrally list cars --team-repository=apm | grep x-pack-apm
x-pack-apm               mixin   X-Pack APM Data plugin

#### 2. download corpora in the local disk
Until the assets are available in the public archive, we can use this temp repo.

```
# create a temp dir, assume we work in `~/apm-rally-track-test`
# mkdir apm-rally-track-test
# cd apm-rally-track-test

git clone https://github.com/kyungeunni/rally-tracks-apm.git
```

#### 3. run rally 
We can now run rally using `x-pack-apm` car with the track locally:
```
esrally race  --track-path=~/rally-tracks-apm --kill-running-processes  --car="defaults,x-pack-apm" --team-repository=apm --track-params="bulk_iterations:1000,expected_cluster_health:yellow"
```

You can also run the rally with test mode by adding --test-mode:
```
esrally race  --track-path=~/rally-tracks-apm --kill-running-processes  --car="defaults,x-pack-apm" --team-repository=apm --track-params="bulk_iterations:1000,expected_cluster_health:yellow" --test-mode
```

cc @elastic/apm-server


elastic/apm/track.json
```diff
@@ -0,0 +1,57 @@
...
+    {
+      "operation": {
+        "operation-type": "delete-data-stream",
+        "data-stream": [
+          "traces-apm*",
+          "metrics-apm*",
+          "logs-apm*"
+        ]
+      }
+    },
```
simitt, line 31: This only deletes existing datastreams but not the templates, right?

kyungeunni: Yes, it deletes only the existing data streams. more about this operation can be found here: https://esrally.readthedocs.io/en/stable/track.html#delete-data-stream

elastic/apm/component-template.json
```diff
@@ -0,0 +1,10 @@
+{
+  "template": {
+    "settings": {
+      "index": {
+        "number_of_shards": {{number_of_shards | default(1)}},
+        "number_of_replicas": {{number_of_replicas | default(0)}}
+      }
+    }
+  }
+}
```
simitt, line 10: The [index templates from the ES apm plugin](https://github.com/elastic/elasticsearch/tree/main/x-pack/plugin/apm-data/src/main/resources/index-templates) already contain a `composed_of` section; a common one is `apm@settings`. 
If this component template was named `apm@settings`, then it should be included in the default index templates.

elastic/apm/index-template.json
```diff
@@ -0,0 +1,9 @@
+{
+    "data_stream": {},
+    "index_patterns": ["traces-apm*", "logs-apm*", "metrics-apm*"],
+    "_meta": {
+        "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
+    },
+    "composed_of": ["common-apm@custom"],
+    "priority": 1
+}
```
simitt, line 9: As mentioned above, I don't think the index-template should be specificed. This is overwriting the default index templates, and we wanted to rely on them for the settings, mappings and processors.

elastic/apm/tasks/setup.json
```diff
@@ -0,0 +1,47 @@
...
+  "operation": {
+    "operation-type": "delete-component-template"
+  }
+},
+{        
+  "name": "create-all-component-templates",
+  "operation": {
+    "operation-type": "create-component-template"
+  }
+},
```
simitt, line 29: Don't these operations delete all the default templates for apm? Or is this called before the ES apm plugin is loaded? I didn't look at any order.

simitt (changes requested):

elastic/apm/index-template.json
kyungeunni: This index template is needed as a placeholder to apply replica settings when there's no index template installed by ES plug-in, for example, in test mode(it does not install the ES Plugin for apm).

Also, this template is named differently to the index templates for APM so it won't be overwritten but co-existing with them. The thing we need to be careful is the index pattern collision. If we run the race against the cluster with ES Plug-in enabled(using `x-pack-apm` car), this placeholder template has a lower [priority](https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/apm-data/src/main/resources/index-templates/traces-apm%40template.yaml#L4) than the one from the plug-in so the this index template will be ignored according to the [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-templates.html) (_Avoid index pattern collisions_ section)

elastic/apm/tasks/setup.json
kyungeunni: `delete-*` operations will delete the templates specified in the `track.json` specification. So yes if we are going to use the same name for a component(e.g. `apm@settings` as suggested in the other comment) we shouldn't delete the component templates but just update them. Thanks for pointing this out, I will address this  🙏

elastic/apm/component-template.json
kyungeunni: Thanks, I was confused with the one we had in the integration package. Will name this as apm@settings.

elastic/apm/index-template.json
simitt: ah, that clarifies things, I wasn't aware that the ES plugin is not installed for tests;
in that case I agree with the solution you created, and take back my comment to rename the component templates; with your solution the `replica:0` would only be applied in tests, but not when actually performing the rally track -> seems to be the right choice.

elastic/apm/component-template.json
simitt: based on your explanations, let's keep things as you have them (see comment below)

simitt (approved):
LGTM as long as CI is green! 🎉

elastic/apm/index-template.json
```diff
@@ -0,0 +1,9 @@
+{
+    "index_patterns": ["traces-apm*", "logs-apm*", "metrics-apm*"],
+    "data_stream": {},
+    "_meta": {
+        "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
+    },
+    "composed_of": ["apm@settings"],
```
axw, line 7: Using `apm@settings` is a bit brittle. Yes the apm-data plugin does use it, but it's not intended as an extension point -- it's a built-in component template that is not intended to be modified by users.

I would strongly suggest using `traces@custom`, `logs@custom`, and `metrics@custom` instead.

PR #552, opened 2024-01-18
Verify API key in serverless IT

gbanasiak:
In some serverless IT runs API key authentication failures are seen on the first use of the key, while all subsequent uses are working fine. To confirm API key is usable this PR introduces an extra API key verification loop in the test setup.


gareth-ellis (approved):
LGTM

PR #553, opened 2024-01-18
Fix tsdb k8s for serverless

gbanasiak:


tsdb_k8s_queries/container-template.json
```diff
@@ -11,9 +11,6 @@
           {# non-serverless-index-settings-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%}
```
gizas, line 11: `!= "serverless"`  should not this be == "serverless"  ?

Also why we need to variables and not only one?

gizas:
Thank you for this fix @gbanasiak !
Can we add the new parameters to readme here:
https://github.com/elastic/rally-tracks/blob/de22ef18470d058453eca4bdf2fb7a5c7380c016/tsdb_k8s_queries/README.md#parameters

Also usually we set default values (like [here](https://github.com/elastic/rally-tracks/blob/de22ef18470d058453eca4bdf2fb7a5c7380c016/tsdb_k8s_queries/track.json#L4)), so how about setting them in your case as well?

To your questions:

1. I agree feel free to comment your sections for if cases. It is really good timing for me to understand what parts of templates we actually need because we are trying to automate the extraction of the template
2. For 2 I am not totally sure, as we actually merged two templates in one:
- From metrics index template we have the message (see [comment](https://github.com/elastic/rally-tracks/pull/553#discussion_r1457417842))
- From metrics-elastic_agent.metricbeat we have the rest see below:

Screenshot 2024-01-18 at 3 04 42 PM

Shall we merge them? Does this affect final rendering times?

tsdb_k8s_queries/container-template.json
```diff
@@ -11,9 +11,6 @@
...
           "number_of_shards": {{number_of_shards | default(1)}},
           "number_of_replicas": {{number_of_replicas | default(0)}},
-          "lifecycle": {
```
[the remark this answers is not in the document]
gbanasiak: The ILM configuration is still there, see [here](https://github.com/gbanasiak/rally-tracks/blob/de22ef18470d058453eca4bdf2fb7a5c7380c016/tsdb_k8s_queries/pod-template.json#L34-L36) and [here](https://github.com/gbanasiak/rally-tracks/blob/de22ef18470d058453eca4bdf2fb7a5c7380c016/tsdb_k8s_queries/container-template.json#L34-L36).

When running against stateful ES the structure I've mentioned in the PR renders as:
```
### serverless settings restricted for operator
### settings unavailable in serverless
### remaining settings
```

gbanasiak:
> Can we add the new parameters to readme here:

These are global variables documented in https://esrally.readthedocs.io/en/stable/serverless.html#tracks. We haven't added them to README track files anywhere, similarly to `now` global variable which existed prior to serverless.

> I agree feel free to comment your sections for if cases.

Ok, I'll add something more descriptive.

> For 2 I am not totally sure, as we actually merged two templates in one [..]

I see. With nominal integration setup ES would apply either `metrics-elastic_agent.metricbeat` or `metrics` template depending on the name of the datastream. If you need to combine them in Rally, `message` field should get added together with other fields. I'm not even sure what is the effect of specifying `query.default_field` twice.

gbanasiak:
I've addressed comments in https://github.com/elastic/rally-tracks/pull/553/commits/583101e1dbf5f2e9aa57398392f3ebef41fea099.

tsdb_k8s_queries/container-template.json
```diff
@@ -38,13 +32,15 @@
...
           {%- endif -%}{# non-serverless-index-settings-marker-end #}
+            {# Note: index settings which can be applied freely in serverless begin here #}
           "mapping": {
             "total_fields": {
               "limit": "10000"
             }
           },
           "query": {
             "default_field": [
+              "message",
```
gizas, line 43: I would say to completely remove it

tsdb_k8s_queries/pod-template.json
```diff
@@ -38,13 +32,15 @@
...
           {%- endif -%}{# non-serverless-index-settings-marker-end #}
+            {# Note: index settings which can be applied freely in serverless begin here #}
           "mapping": {
             "total_fields": {
               "limit": "10000"
             }
           },
           "query": {
             "default_field": [
+              "message",
```
gizas, line 43: Same here delete it

gizas:
@martijnvg this LGTM. As you are the main stakeholder of this track any other objection to merge?
@gbanasiak I only left one comment to remove `message`

gizas (approved):

tsdb_k8s_queries/container-template.json
gbanasiak: That suggestion was confusing at first to me as I thought the intention was to merge templates described in https://github.com/elastic/rally-tracks/pull/553#issuecomment-1898464136, but then I realised corpus does not include `message` field at all it seems. I've addressed that in https://github.com/elastic/rally-tracks/pull/553/commits/b1bd321591cd7c59dd64798e9c06056b1c2caa43.

tsdb_k8s_queries/pod-template.json
gbanasiak: Resolved in https://github.com/elastic/rally-tracks/pull/553/commits/b1bd321591cd7c59dd64798e9c06056b1c2caa43.

martijnvg (commented):
LGTM

PR #555, opened 2024-01-22
Remove unsupported auto date histograms from tsdb rally track

### LINKED ISSUE #554: Unsupported auto date histogram with nested time series aggregations
Auto date histogram aggregations use `BestBucketsDeferringCollector` which causes `TimeSeriesAggregator` not to be used at all if a time series aggregation is nested as a child of an auto date histogram. Writing one such aggregation at the moment produces a result but that result is incorrect and includes always the same time series bucket as a child of each auto data histogram bucket. This happens because there is no `TimeSeriesIndexSearcher` which fetches, sorts and advances a (doc value) iterator for time series documents. As a result, a typical aggregation would include always the same (first or last depending on sorting) time series bucket with potential issues caused by a (very) large number of buckets returned as a result of the auto date histogram selecting a fine-grain bucket size (depending on the dataset).

We need to remove queries from the Rally dataset which use an auto date histogram with a nested time series aggregation.

salvatore-campagna:
Removes auto date histogram aggregation queries with nested time series aggregations.

Resolves #554


martijnvg (approved):
lgtm

PR #556, opened 2024-01-26
Include dimension fields and their cardinality to README

salvatore-campagna:
It is useful to know which fields are dimensions and their cardinality to
be able to evaluate indexing and querying performances.


martijnvg (approved):
lgtm

PR #559, opened 2024-02-02
Replace PROJECT keyword with KEEP in ESQL queries

luigidellaquila:
Since `PROJECT` keyword [will be removed soon](https://github.com/elastic/elasticsearch/issues/105025) from ESQL grammar, this PR replaces it with the appropriate `KEEP` in the few remaining places where it's still used


craigtaverner (approved):
LGTM, although I wonder if this is also an opportunity to change `keep` to `KEEP`?

PR #560, opened 2024-02-05
Use new name for frequent items agg

### LINKED ISSUE #557: Deprecatied frequent-items, change to frequent-items_sets
When running `frequent-items` challenge in https://github.com/elastic/rally-tracks/tree/master/so on ES 8 there is a deprecation warning:
```
  "event.code": "deprecated_field_frequent_items",
  "message": "Deprecated field [frequent_items] used, expected [frequent_item_sets] instead",
```

davidkyle:
The `frequent_items` name was deprecated in 8.7 and replaced with `frequent_item_sets`. This change updates the SO track to use the new name for the aggregation. 

Closes #557

### DIFF
--- a/so/operations/default.json
+++ b/so/operations/default.json
@@ -166,7 +166,7 @@
         "size": 0,
         "aggs": {
           "fi": {
-            "frequent_items": {
+            "frequent_item_sets": {
               "minimum_set_size": 2,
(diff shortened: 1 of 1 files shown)


walterra (approved):
LGTM

PR #561, opened 2024-02-09
Add a new track to benchmark filtered search on dense vectors

jimczi:
This PR adds a new track designed for benchmarking filtered search on random vectors.

By default, the `flat` vector_index_type is utilized to evaluate the performance of brute force search over vectors filtered by a partition ID.

## Indexing

To begin indexing, the track initiates `index_clients` clients, each executing `index_iterations` bulk operations of size `index_bulk_size`. Consequently, the total number of documents indexed by the track is calculated as follows: `index_clients` * `index_iterations` * `index_bulk_size`.

Each document in the bulk is assigned a random vector of dimensions `dims` and a random partition ID.

## Search Operations

Search operations involve filtering by a random partition ID and scoring against a random query vector. These operations are executed against the index using various DSL flavors, including script score and knn section.


benwtrent (approved):
I like it. It seems like we also get random partition sizes as different bulk calls could be for the same partition (making it larger) or we only do a bulk index for a given partition once.

jimczi:
> It seems like we also get random partition sizes as different bulk calls could be for the same partition (making it larger) or we only do a bulk index for a given partition once.

The total number of partitions is fixed (`partitions`) and we generate one partition id (randomised between 0 and the total number of partitions) per document.

PR #562, opened 2024-02-15
Adds field-caps non empty field task in the many-shards-quantitative challenge

piergm:
Recently we added a new optional query parameter to the _field_caps API in order to 
filter out fields with no values. The parameter is called `include_empty_fields`  and 
defaults to true, and if set to false it will filter out from the field_caps response all the 
fields that has no value in the index. With this PR we want to benchmark this change 
by adding a new task to the many-shard-quantitative challenge.

relates to: https://github.com/elastic/elasticsearch/pull/103651

### DIFF
--- a/elastic/logs/challenges/many-shards-quantitative.json
+++ b/elastic/logs/challenges/many-shards-quantitative.json
@@ -83,7 +83,7 @@
         ]
       }
     },
-    {% include "tasks/field-caps.json" %}
+    {% include "tasks/field-caps.json" %},
     {% include "tasks/field-caps-non-empty-fields.json" %}
   ]
 }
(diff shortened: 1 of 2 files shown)


elastic/logs/challenges/many-shards-quantitative.json
```diff
@@ -84,5 +84,6 @@
       }
     },
     {% include "tasks/field-caps.json" %}
```
gareth-ellis, line 86: You'll need a comma at the end of this line to maintain valid json

piergm: Thanks! Fixed now :D

piergm:
@gareth-ellis it is in main now and will be available from 8.13 onwards. How do I avoid for it to be back-ported to tags before < 8.13?

gareth-ellis:
I'd misremembered, we backport changes that will work with older versions, if it won't work, then we don't need to worry. It's useful that the information is in the PR at least for future reference

gareth-ellis (approved):
LGTM

PR #568, opened 2024-02-19
field-caps non-empty-fields CCS task

piergm:
Same request as the field-caps above but with the optional parameter `include_empty_fields` set to `false`


gareth-ellis (approved):
LGTM

PR #569, opened 2024-02-20
Add autoscale-v2 challenge to NYC Taxis track

gbanasiak:
This adds a new version of `autoscale` challenge where each step is defined with an array, instead of a fixed formula. The new version of the challange allows to specify both clients and target throughput, or only clients.

The challenge relies on bulk `looped` mode introduced in https://github.com/elastic/rally/pull/1830.

### DIFF
--- a/nyc_taxis/README.md
+++ b/nyc_taxis/README.md
@@ -84,7 +84,7 @@ This challenge is intended for serverless index tier autoscaler testing. It cons
 Note: `include_target_throughput` parameter is ignored in this challenge.
 
 * `as_clients` (default: []): An array with the number of indexing clients to be used in each step.
-* `as_time_periods` (default: []): An array with time period of every step. Time period is divided in halves. The first half is configured as warm-up.
+* `as_time_periods` (default: []): An array with time period, in seconds, of every step. Time period is divided in halves. The first half is configured as warm-up.
 * `as_target_throughputs` (default: []): An array with target throughput of each step, expressed in requests/s. Please use `bulk_size` parameter to translate this into docs/s. Target throughput is not configured if the values specified in this array are negative.
 
 ### License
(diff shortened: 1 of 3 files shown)


nyc_taxis/README.md
```diff
@@ -78,6 +78,15 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `include_force_merge` (default: true for non-serverless clusters, false for serverless clusters): Whether to include force merge operation.
 * `include_target_throughput` (default: true for non-serverless clusters, false for serverless clusters): Whether to apply target throughput.
 
+### Parameters specific to autoscale-v2 challenge
+
+This challenge is intended for serverless index tier autoscaler testing. It consists of configurable number of steps provided with array parameters.
+Note: `include_target_throughput` parameter is ignored in this challenge.
+
+* `as_clients` (default: []): An array with the number of indexing clients to be used in each step.
+* `as_time_periods` (default: []): An array with time period of every step. Time period is divided in halves. The first half is configured as warm-up.
```
inqueue, line 87:
```suggestion
* `as_time_periods` (default: []): An array with time period, in seconds, of every step. Time period is divided in halves. The first half is configured as warm-up.
```

nyc_taxis/challenges/common/autoscale-v2-schedule.json
```diff
@@ -0,0 +1,63 @@
...
+  "operation": {
+    "operation-type": "cluster-health",
+    "index": "nyc_taxis",
+    "request-params": {
+      "wait_for_status": "{{cluster_health | default('green')}}"
+    },
+    "retry-until-success": true
+  }
+}
+{%- set p_as_clients = (as_clients | default([]))%}
```
inqueue, line 30: What do you think about including some default values for these parameters? Something basic like `[8, 16, 32]` for clients and whatever makes sense for the other lists would be fine.

inqueue (approved):
I left a couple of comments. In all, LGTM.

nyc_taxis/challenges/common/autoscale-v2-schedule.json
gbanasiak: Sure, addressed in https://github.com/elastic/rally-tracks/pull/569/commits/3c3d709f619a1b85556f82634134bcb8f2ca0205.

PR #572, opened 2024-02-23
elastic/apm: do not update/use `apm@settings` template

kyungeunni:
Addresses https://github.com/elastic/rally-tracks/pull/550/commits/5c8d4fe310a050adf4d2acd2fbf7cde21e3cbc0f#r1500196363

### DIFF
--- a/elastic/apm/index-template.json
+++ b/elastic/apm/index-template.json
@@ -4,6 +4,6 @@
     "_meta": {
         "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
     },
-    "composed_of": ["apm@settings"],
+    "composed_of": ["common-apm@custom"],
     "priority": 1
 }
--- a/elastic/apm/index-template.json
+++ b/elastic/apm/index-template.json
@@ -4,6 +4,6 @@
     "_meta": {
         "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
     },
-    "composed_of": ["common-apm@custom"],
+    "composed_of": ["traces@custom", "metrics@custom", "logs@custom"],
     "priority": 1
 }
--- a/elastic/apm/index-template.json
+++ b/elastic/apm/index-template.json
@@ -1,9 +1,28 @@
 {
-    "index_patterns": ["traces-apm*", "logs-apm*", "metrics-apm*"],
-    "data_stream": {},
-    "_meta": {
-        "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
-    },
-    "composed_of": ["traces@custom", "metrics@custom", "logs@custom"],
-    "priority": 1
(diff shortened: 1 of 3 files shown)


elastic/apm/index-template.json
```diff
@@ -4,6 +4,6 @@
     "_meta": {
         "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
     },
-    "composed_of": ["apm@settings"],
+    "composed_of": ["common-apm@custom"],
```
axw, line 7: What's the reason for defining another non-standard `@custom` component template? Can we use `traces@custom`, `metrics@custom`, and `logs@custom` here?

elastic/apm/track.json
```diff
@@ -10,7 +10,19 @@
   ],
   "component-templates": [
     {
-      "name": "apm@settings",
```
axw, line 13: I think you should change these to `traces@custom`, `metrics@custom`, and `logs@custom`. For example, the `metrics-apm.service_destination.1m-*` index template includes `metrics@custom` and `metrics-apm.service_destination.1m@custom`; it does _not_ include `metrics-apm@custom`.

axw (changes requested):

elastic/apm/index-template.json
kyungeunni: oops, yes you are right. I should just use the standard custom component.

elastic/apm/track.json
kyungeunni: Thanks I updated the PR.

axw (approved):

elastic/apm/index-template.json
```diff
@@ -1,9 +1,28 @@
...
-    },
-    "composed_of": ["traces@custom", "metrics@custom", "logs@custom"],
-    "priority": 1
+  "index_patterns": ["traces-apm*", "logs-apm*", "metrics-apm*"],
+  "data_stream": {},
+  "_meta": {
+      "description": "Placeholder template that'd be used when the x-pack-apm car is not used, e.g. in test mode"
+  },
+  "composed_of": ["traces@custom", "metrics@custom", "logs@custom"],
+  "priority": 101,
```
kyungeunni, line 8: setting it between apmplugin template(140) and metrics-*-* template(100)

elastic/apm/index-template.json
```diff
@@ -1,9 +1,28 @@
...
+                "value": {
+                  "type": "long"
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
```
kyungeunni, line 27: @axw I've added this to fix the tests which is there to check the track very quickly for [syntax errors](https://esrally.readthedocs.io/en/stable/adding_tracks.html?highlight=test%20mode#adding-support-for-test-mode). tests were falling due to mapping errors, which shouldn't happen when the proper ingest pipeline/template is installed. In test mode it is not enabled by default so this is a workaround until we enable the apm plugin by default in ES.

axw (approved):

PR #504, opened 2023-10-31
Unblock elastic/logs for public serverless

gbanasiak:
Similar to https://github.com/elastic/rally-tracks/pull/465.

Remarks:
* unblocking limited to `logging-indexing`, `logging-querying` and `logging-indexing-querying` challenges which are applicable to serverless today,
* shared custom index settings moved to a dedicated component template (`track-custom-shared-settings`) to avoid repetition, this makes `@custom` templates empty, but these are kept as placeholders for any future changes,
* there is no explicit force merge step in `elastic/logs` so optional post-ingest sleep was not introduced,
* `compression-statistics` custom runner had to be excluded at the track level as it is based on index stats.

Tests:
```
TRACK=elastic/logs
esrally race --track-path=../rally-tracks/$TRACK --target-hosts=${ES_HOST}:443 --pipeline=benchmark-only --client-options="use_ssl:true,api_key:${ES_API_KEY}" --on-error=abort --challenge="logging-indexing" --test-mode --track-params="start_date:2021-01-01T00-00-00Z,end_date:2021-01-01T00-10-00Z"
esrally race --track-path=../rally-tracks/$TRACK --target-hosts=${ES_HOST}:443 --pipeline=benchmark-only --client-options="use_ssl:true,api_key:${ES_API_KEY}" --on-error=abort --challenge="logging-indexing" --test-mode --track-params="start_date:2021-01-01T00-00-00Z,end_date:2021-01-01T00-10-00Z,refresh_interval:17s"
esrally race --track-path=../rally-tracks/$TRACK --target-hosts=${ES_HOST}:443 --pipeline=benchmark-only --client-options="use_ssl:true,api_key:${ES_API_KEY}" --on-error=abort --challenge="logging-querying" --test-mode --exclude-tasks="tag:setup" --track-params="query_warmup_time_period:1,query_time_period:1,workflow_time_interval:1,think_time_interval:1"
esrally race --track-path=../rally-tracks/$TRACK --target-hosts=${ES_HOST}:443 --pipeline=benchmark-only --client-options="use_ssl:true,api_key:${ES_API_KEY}" --on-error=abort --challenge="logging-indexing-querying" --test-mode --track-params="start_date:2021-01-01T00-00-00Z,end_date:2021-01-01T00-10-00Z,query_warmup_time_period:1,query_time_period:1,workflow_time_interval:1,think_time_interval:1"
```

PR #573, opened 2024-02-23
Adjust elastic/logs update-custom-package-templates for serverless

gbanasiak:
Follow-up to https://github.com/elastic/rally-tracks/pull/563 which missed to update `update-custom-package-templates` task based on the presence and value of `include_non_serverless_index_settings` track parameter.

We have 2 overlapping mechanisms for specifying index settings in `elastic/logs`:
* static assets: specifically `track-custom-shared-settings.json` component template referenced in composable templates, introduced in https://github.com/elastic/rally-tracks/pull/504, it accompanies another `track-custom-mappings.json` component template,
* pre-existing assets: `update-custom-package-templates` task from `index-setup.json`, introduced in https://github.com/elastic/rally-tracks/pull/336, based on a custom runner which updates _all_ component templates ending with `@custom` in the cluster.

The initial idea in https://github.com/elastic/rally-tracks/pull/336 was to only trigger modification of pre-existing assets if specific track parameter is set, but ultimately an overlap scenario was chosen (see [this](https://github.com/elastic/rally-tracks/pull/336#issuecomment-1284522228) comment). The plan that was never implemented was to get rid of static assets in favor of the custom runner (see [this](https://github.com/elastic/rally-tracks/pull/336#issuecomment-1285712498) comment).

What I don't like about the custom runner approach is that it modifies _all_ component templates ending with `@custom` in the cluster ([src](https://github.com/elastic/rally-tracks/blob/2794e4c4c641c3e0096b08f1b084f2142d8672f3/elastic/shared/runners/update_custom_templates.py#L19)). What about benchmarks against live clusters? This should be an opt-in, not a default, right?

What I'm proposing in this PR is a logical extension to https://github.com/elastic/rally-tracks/pull/563 which maintains the status quo, but I'm open to extending the scope a bit, and making `update-custom-package-templates` optional and disabled by default. WDYT?

### DIFF
--- a/elastic/logs/tasks/index-setup.json
+++ b/elastic/logs/tasks/index-setup.json
@@ -77,8 +77,10 @@
         "settings": {
           "index" : {
             {# non-serverless-index-settings-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%}
+              {% if p_include_non_serverless_index_settings %}
             "number_of_replicas" : "{{ number_of_replicas | default('1')}}",
             "number_of_shards" : "{{ number_of_shards | default('1')}}"
+              {% endif %}
             {%- endif -%}{# non-serverless-index-settings-marker-end #}
           }
         },
--- a/elastic/logs/track.json
+++ b/elastic/logs/track.json
@@ -20,6 +20,7 @@
 {% set p_query_warmup_time_period = (query_warmup_time_period | default(120)) %}
 {% set p_query_time_period = (query_time_period | default(900)) %}
 {% set p_query_request_params = (query_request_params | default({}))%}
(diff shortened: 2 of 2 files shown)


gareth-ellis (approved):
The changes LGTM,

With regard to changing the behaviour and whether updating everything should be opt in or opt out, I'm a little unsure. We do say in various documentation that users of rally and rally-tracks have to be aware that some or all tracks may not be suitable for running on a production server, and instead should be benchmarking on a seperate cluster that they don't mind losing all the data. That said, this does behave slightly differently than others, whereby other tracks tend to just touch the indexes associated with the track.

I think I'd be happy either way, along with a comment in the README.md that points out what's happening

PR #574, opened 2024-03-05
Tasks for benchmarking field_caps with index-filter

piergm:
Added two more tasks to benchmark field_caps requests with index-filter for both standard and non-empty field requests.


gareth-ellis (approved):
LGTM, presuming this has been tested ?

PR #575, opened 2024-03-07
Add new benchmark for msmarco-v2-embed-english-v3 dataset

1stvamp:
## https://github.com/elastic/enterprise-search-team/issues/6793

I've added a new track (`cohere_v3_vector`) alongside the existing `cohere_vector` track added in #444. This track will be used by the Search teams to test vector search performance against 139M doc + 1024 dimension vectors (vs. 39M + 768 dimension vectors in the original track).

I had a build to a good-enough queries dataset from the Cohere APIs as they haven't published a queries dataset for us to copy yet.


1stvamp:
Opened as a draft as I'm waiting for CI for the last push to complete.

cohere_v3_vector/index-vectors-with-text-mapping.json
```diff
@@ -0,0 +1,34 @@
...
+      },
+      "text": {
+        "type": "text"
+      },
+      "emb": {
+        "type": "dense_vector",
+        "element_type": "float",
+        "dims": 1024,
+        "index": true,
+        "similarity": "max_inner_product",
```
jimczi, line 27: The embeddings are normalized already so we can use `dot_product` directly

cohere_v3_vector/README.md
```diff
@@ -0,0 +1,73 @@
+## Cohere msmarco-v2/embed-english-v3 vector track
```
jimczi, line 1: I wonder we should call it msmarco-v2? 
If we want to compare another strategy (BM25 or another model for semantic search) it would be preferable to have a single track. For now it's ok to only test the cohere model but it would be nice to link this track with a dataset rather than a search strategy.

cohere_v3_vector/README.md
```diff
@@ -0,0 +1,73 @@
...
+  "text": "0-60 Times - 0-60 | 0 to 60 Times & 1/4 Mile Times | Zero to 60 Car Reviews.",
+  "emb": [0.00507354736328125, 0.01519775390625, -0.002109527587890625, ...]
+}
+```
+
+### Generating the queries
+
+The `queries.json` can be rebuilt using the `_tools/generate_queries.py`, this will stream a partial training dataset from hugging face, call the Cohere generate API for each to get a question based on it, and then call the Cohere embed API for the question, and store the embeddings in `queries.json`.
+This will take a very long time, maybe grab a ☕️ ?
+
```
jimczi, line 47: That's ok if we "just" want to test the performance but ideally we should use the [official queries](https://ir-datasets.com/msmarco-passage-v2.html#msmarco-passage-v2/train). Then in a follow up it will be easy to add similarity measures in the track based on the associated `qrels`.

cohere_v3_vector/index-vectors-only-mapping.json
```diff
@@ -0,0 +1,29 @@
...
+    },
+    "properties": {
+      "emb": {
+        "type": "dense_vector",
+        "element_type": "float",
+        "dims": 1024,
+        "index": true,
+        "similarity": "max_inner_product",
+        "index_options": {
+          "type": {{ vector_index_type | default("int8_hnsw") | tojson }}
```
jimczi, line 24: The embeddings are normalized already so we can use dot_product directly

jimczi (commented):
Thanks @1stvamp! I left some comments.

cohere_v3_vector/index-vectors-with-text-mapping.json
1stvamp: Annoyingly it looks like they aren't all normalized, ES is complaining that a bunch of vectors aren't normalized, and while I can confirm the queries are definitely unit-length vectors as the embeds came from the Cohere API which serves normalized vectors only, it looks like the dataset isn't.

cohere_v3_vector/index-vectors-only-mapping.json
1stvamp: https://github.com/elastic/rally-tracks/pull/575#discussion_r1517068264

1stvamp:
@jimczi reverted to `max_inner_product` for now due to normalization issues in the dataset, I have pushed a separate branch that adds normalization that we can land after (as it'll take hours to regenerate the dataset and then test against the benchmarks thoroughly): https://github.com/elastic/rally-tracks/commit/66e76b593e5b8d41991b3b7426a22a5e018d781c

cohere_v3_vector/index-vectors-only-mapping.json
jimczi: oh that's weird, we can use `cosine` for now then but I wonder now if the published dataset has the right embeddings. Maybe they just disabled it but that's not possible to do using the API and the v3 models.

jimczi (approved):
LGTM, we need to understand why the vectors are not normalised already but this can be done in a follow up.

PR #576, opened 2024-03-09
Add the ability to override track data source url

gareth-ellis:
Both of these tracks (and probably others as well), have very large corpora, which takes a lot of time to download if downloading cross region. To enable to use a local mirror, I have added an option to allow us to easily override the location via a track parameter.


gbanasiak (approved):
LGTM

PR #577, opened 2024-03-13
Update elastic/apm to use time-period for bulk operation

kyungeunni:
run races for a specified time period instead of iterations.

example command:
```
esrally race --track=elastic/apm --on-error=abort  --kill-running-process \
--track-params="bulk_size:5700,bulk_time_period:3600,bulk_warmup_period:120,bulk_clients:8"  \
--car="defaults,x-pack-apm" \
--client-options="http_compress:true"
```


kyungeunni:
The tests should be fixed by https://github.com/elastic/rally-tracks/pull/572

simitt (approved):
LGTM, not sure if a different team's approval is required also.

PR #578, opened 2024-03-13
Normalize msmarco-v2-vector queries and switch to dot-product

1stvamp:
This:

- Normalizes vectors for queries that come back from the Cohere API
- Switches to using `dot-product` for vector similarity

After an investigation I found the vectors from the msmarco-v2 dataset cohere published **are** normalised, but the queries from their API, despite their API docs saying to the contrary, are not, and another ingestion issue made it look like it was the dataset and not the queries in CI.

Dataset file sizes changed in GCP, and so in `track.json`, due to playing around with normalisation, but is the same.

### DIFF
--- a/msmarco-v2-vector/_tools/parse_documents.py
+++ b/msmarco-v2-vector/_tools/parse_documents.py
@@ -1,8 +1,6 @@
 import json
 import sys
 
-import numpy
-import vg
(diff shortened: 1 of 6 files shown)


jimczi (approved):
LGTM

PR #579, opened 2024-03-18
[Wikipedia] Ensure that indexation always runs during the parallel ingest/search operation

jimczi:
Add `looped` option in the parallel-documents-indexing-bulk operation.


gareth-ellis (approved):
LGTN

PR #580, opened 2024-03-18
[ES-8028] Make the [pragma] for esql segment partitioning only happen on stateful

favilo:
Also related: ES-7623

### DIFF
--- a/nyc_taxis/operations/default.json
+++ b/nyc_taxis/operations/default.json
@@ -1091,9 +1091,6 @@
       "name": "avg_amount_group_by_keyword_esql_segment_partitioning",
       "operation-type": "esql",
       "query" : "FROM nyc_taxis | stats avg(total_amount) by rate_code_id | sort rate_code_id",
-  {# non-serverless-segment-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
-      "body": { "pragma": { "data_partitioning": "segment" } }
-  {%- endif -%}{# non-serverless-segment-partitioning-marker-end #}
     },
   {# non-serverless-doc-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
     {
@@ -1123,9 +1120,6 @@
       "name": "count_group_by_keyword_esql_segment_partitioning",
       "operation-type": "esql",
       "query" : "FROM nyc_taxis | stats count=count(*) by rate_code_id | sort count desc",
-  {# non-serverless-segment-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
-      "body": { "pragma": { "data_partitioning": "segment" } }
-  {%- endif -%}{# non-serverless-segment-partitioning-marker-end #}
     },
   {# non-serverless-doc-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
     {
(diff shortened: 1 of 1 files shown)


nyc_taxis/operations/default.json
```diff
@@ -960,14 +960,18 @@
       "name": "avg_passenger_count_esql_segment_partitioning",
       "operation-type": "esql",
       "query" : "FROM nyc_taxis | stats avg(passenger_count)",
+  {# non-serverless-doc-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
```
craigtaverner, line 963: Would it not be easier to simply delete the entire body, since this is the default option?

favilo: Done

nyc_taxis/operations/default.json
```diff
@@ -959,15 +959,16 @@
     {
       "name": "avg_passenger_count_esql_segment_partitioning",
       "operation-type": "esql",
-      "query" : "FROM nyc_taxis | stats avg(passenger_count)",
-      "body": { "pragma": { "data_partitioning": "segment" } }
+      "query" : "FROM nyc_taxis | stats avg(passenger_count)"
```
craigtaverner, line 962: Nice. The removal of the body and the serverless macros makes this much easier to read. We can still see that it is `segment_partitioning` in the `name`, so that knowledge is there for future readers of this code.

nyc_taxis/operations/default.json
```diff
@@ -1002,15 +1003,19 @@
     {
       "name": "avg_tip_percent_esql_segment_partitioning",
       "operation-type": "esql",
-      "query" : "FROM nyc_taxis | where fare_amount > 0 | eval tip_percent = tip_amount / fare_amount | stats avg(tip_percent)",
+      "query" : "FROM nyc_taxis | where fare_amount > 0 | eval tip_percent = tip_amount / fare_amount | stats avg(tip_percent)"
+  {# non-serverless-sement-partitioning-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%},
```
craigtaverner, line 1007: This body can be removed.

craigtaverner (changes requested):
Looks much better, but you left one segment partitioning body in place in the `avg_tip_percent_esql_segment_partitioning` operation.

nyc_taxis/operations/default.json
favilo: That is a great point.

favilo:
I've cleaned up a bunch of commas that ended up getting pasted in on accident. And I've had a successful benchmark run with these changes

favilo:
Adding es-perf here for double checking

gbanasiak (commented):
The `serverless_operator == true` part of the condition will keep queries with `pragma` when operator privileges are used to access serverless clusters. If the intention is to avoid `pragma` for _all_ serverless runs regardless of privilege level, the condition should be simplified to `{%- if build_flavor != "serverless" -%}`.

An alternative option would be to let operators still run `pragma` queries but add inner condition based on a track parameter as in: https://github.com/elastic/rally-tracks/blob/db69e57cdc96dc9a45e8600b79f62b27ab98263b/geonames/index.json#L5-L12

This would allow an easy include/exclude of `pragma` queries later at the benchmark definition level, depending on availability of `pragma` in ES/QL serverless build, or general serverless build (at a later stage).

favilo:
I've removed the `serverless_operator` condition

dpifke-elastic (approved):
LGTM.

favilo:
@craigtaverner Mind giving this another look?

craigtaverner (approved):
LGTM

PR #581, opened 2024-03-19
msmarco-v2-vector: fix bzip2 data corruption and vector normalization

1stvamp:
- Using `pv` with `bzip2` on extremely large dataset files is leading to corruption, so dataset regenerated just with straight `pbzip2` and instructions updated in readme.
- Dataset published by Cohere despite previous PR's confusion _definitely is not normalized_ so I've added normalization to the `parse_documents.py` script that generates the dataset.
- Compressed file sizes have changed to match uncorrupted bzip2 data in GCS, uncompressed has not as previous dataset run was done using nomalization in the script, when investigating previous issue.

### DIFF
--- a/msmarco-v2-vector/README.md
+++ b/msmarco-v2-vector/README.md
@@ -23,7 +23,7 @@ done
 
 # Zip each document file for uploading
 $ for file in cohere-documents-*; do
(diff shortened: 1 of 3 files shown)


jimczi (approved):
LGTM

PR #584, opened 2024-03-28
Add a logging-indexing variation track for data stream autosharding

andreidan:
This takes some inspiration from the autoscale track - i.e. we'll want to run it in several steps, increasing the load for a few steps and decreasing it for the last stage and check that the number of shards increases and decreases accordingly (we'll also want to modify the cooldown periods, especially the one for decreasing the number of shards which is very large by default i.e. `3d`)


inqueue (approved):
LGTM, thank you!

PR #586, opened 2024-04-03
Update default value for auto sharding excludes

andreidan:
This updates the default for the auto sharding excludes
cluster setting to correctly default to emtpy array `[]` when
not specified (as opposed to `""`)


inqueue (approved):
LGTM

PR #588, opened 2024-04-04
[ES-8186] Add ES|QL queries to the `many-shards-quantitative` track

favilo:
This will allow us to benchmark ES|QL against many shards which is a known issue


ebadyano (approved):
LGTM, but maybe lets wait till Monday to merge?

elastic/logs/challenges/many-shards-quantitative.json
```diff
@@ -86,6 +86,170 @@
...
+    },
+    {
+      "operation": "esql_dissect_duration_and_stats",
+      "clients": 1,
+      "warmup-iterations": 5,
+      "iterations": 20,
+      "tags": ["esql", "dissect"]
+    },
+    {
+      "operation": "esql_from_all_limit_50",
```
bpintea, line 203: I'm fine adding all these, but we can also start with just this query only.

favilo: Should I remove the query cache stuff, too?

craigtaverner: The query cache stuff is only relevant to the non-ESQL queries. We added those for comparison to ESQL (ie. picked similar queries to compare how ESQL performs against _search). ESQL does not support a query cache, and the query cache is turned off for all tracks except elastic/logs, so we had to turn it off here too to make the comparison reasonable. If we do not include the _search queries in this benchmark, we do not need the query cache parts either.

craigtaverner: Personally I think comparing ESQL to _search is useful. On the other hand we do get a reasonable amount of that with the nyc_taxis track already.

PR #590, opened 2024-04-22
Add a categorize-text challenge to the elastic/logs track.

edsavage:
Adds a challenge for the `caegorize-text` aggregation, something for which we're keen to keep an eye on performance.


jan-elastic (approved):
I don't know anything about rally configuration, but these configs look reasonable to me.

PR #591, opened 2024-04-23
Add a categorize-text challenge to the elastic/logs track. (#590)

edsavage:
Adds a challenge for the `categorize-text` aggregation, something for which we're keen to keep an eye on performance.

Backports #590


wwang500 (approved):
LGTM 👍

PR #592, opened 2024-04-25
Skip non-operator serverless IT tests for elastic/logs, geonames and http_logs

gbanasiak:
Some of the query parameters like `batched_reduce_size` became unavailable in serverless which has broken `elastic/logs` track for regular user. This PR skips `elastic/logs` from IT tests when run without `--operator` flag.

`elastic/logs` has a dedicated Python module, so I've used pytest marker at the class level as documented [here](https://docs.pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option).

Reference: ES-8366

### DIFF
--- a/it_serverless/conftest.py
+++ b/it_serverless/conftest.py
@@ -204,3 +204,14 @@ def pytest_generate_tests(metafunc):
         operator = metafunc.config.getoption("operator")
         label = "operator" if operator else "user"
         metafunc.parametrize("operator", [operator], ids=[label])
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "operator_only: mark test for operator only")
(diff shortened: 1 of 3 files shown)


gareth-ellis (approved):
LGTM - we seem to have another issue too with python versions for macos - related https://github.com/actions/setup-python/issues/850

gareth-ellis:
@elasticmachine update branch

gbanasiak:
I've added selected challenges from geonames and http_logs in https://github.com/elastic/rally-tracks/pull/592/commits/c2dc4e0b65164f99a0fcc79717025eea757f51f3 due to `pre_filter_shard_size` parameter.

PR #594, opened 2024-04-29
Improvements to the categorize-text challenge

edsavage:
* bulk import index data
* target only those indices that have the required `message` field
* remove code duplication in the challenge definition


maxhniebergall (approved):
LGTM

gareth-ellis (approved):
LGTM - it may be worth parameterising the number of warmup iterations and `iterations`- when I test ran this it took 2days 16 hours.

PR #595, opened 2024-04-30
Add a new `low-latency` challenge to `tsdb` track

salvatore-campagna:
We use the tsdb track with standard index mode because:
* document ids are known upfront and do not change over time
* it does not use synthetic source and keeps the _source field stored

This new challenge is needed to evaluate the impact of ZSTD comrpession on stored fields for low latency queries such as get and mget. It also fills a gap in our benchmarks by also testing the get and mget apis which we do not really test in any other benchmark.

This addresses https://github.com/elastic/elasticsearch/issues/108091


martijnvg (approved):
LGTM

PR #599, opened 2024-05-06
Add backport action

gareth-ellis:
This adds a backport action that should do the following two things:
- Backport any PR to current latest branch (currently 8.15) that we tag with backport. We potentially could remove the requirement to add a label too - so absolutely everything is backported to 8.15
- Backport to other branches (e.g 8.13) by adding a label of v8.13.

The backport will work by creating a new PR(or PRs) against the relevant branch(es).


.backportrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "upstream": "elastic/rally-tracks",
+    "targetBranches": ["8.15"],
```
gbanasiak, line 3: I did some testing, and it looks like PR labels are converted to branch list based on `branchLabelMapping` which is filtered through `targetBranchChoices` which in turn gets overridden by `targetBranches`. So current configuration will ignore the labels and always backport to `8.15` regardless of the labels.

I think we need to choose between the following:
* configure `targetBranches` to the current latest branch which is closest to what we had earlier,
* configure `branchLabelMapping` and label each time (more work but more flexible).

WDYT?

.backportrc.json
```diff
@@ -0,0 +1,9 @@
+{
+    "upstream": "elastic/rally-tracks",
```
gbanasiak, line 2: Let's use non-deprecated option, i.e. `repoOwner` and `repoName`.

.backportrc.json
```diff
@@ -0,0 +1,9 @@
...
+    "targetBranches": ["8.15"],
+    "branchLabelMapping": {
+        "^v(\\d+).(\\d+)": "$1.$2"
+      },
+    "prDescription": "{defaultPrDescription}\n\n<!--BACKPORT {commits} BACKPORT-->"
+
```
gbanasiak, line 8: Unnecessary empty line.

gbanasiak (commented):
I left some comments.

I would add `targetPRLabels` option with a label that will allow us to find or filter out backport PRs quickly. We currently have `backport` label as a trigger. We could add `backported` label to indicate the PR was created by `backport` tool. For instance https://github.com/elastic/kibana is using `auto-backport` / `backport` label pair for this purpose.

gareth-ellis:
@elasticmachine update branch

gbanasiak:
@elasticmachine update branch

gbanasiak:
We had quite a few backports recently so I think it's time to revisit this one. I've simplified setup in https://github.com/elastic/rally-tracks/pull/599/commits/cf7d836abd4432041e0d345604ce0f6900afd6ab based on https://github.com/sorenlouv/backport-github-action documentation. The intended behavior is as follows:
* backport is triggered by `backport-to-...` label, e.g. `backport-to-8.15`,
* there is no automatic backport to the most recent non-`master` branch, see earlier testing in https://github.com/elastic/rally-tracks/pull/599#discussion_r1591005563,
* the backport PR is labeled with `backport` label.

favilo (approved):
This is really neat. Refactoring might be harder to ballet sometimes, but this i feel would make it actually approachable

PR #602, opened 2024-05-08
ESQL Intersects, Contains, Within, Disjoint benchmarks

craigtaverner:
ESQL added a nuch of spatial search features in 8.14, so we should benchmark them.

We add also matching _search API calls for comparison purposes.

Since we explicitly added a new size:10 to the _search API calls, we include a new -origin benchmark to verify that this has no impact. We should remove that later.

Since the benchmarks take a while to run, and each iteration takes a while, I also shortened the iteration could for the long running queries.

### DIFF
--- a/geopoint/challenges/default.json
+++ b/geopoint/challenges/default.json
@@ -141,8 +141,8 @@
         },
         {
           "operation": "polygon-within-esql",
-          "warmup-iterations": 100,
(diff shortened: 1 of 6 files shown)


iverase:
+1 to add this comparison to the benchmarks

iverase (approved):

PR #605, opened 2024-05-13
Attempt to fix osm* issue in ESQL benchmarks

craigtaverner:
We try to delete the conflicting indices first. This does not seem to be working, though.

The error we get is when running `geoshape` track is:

```
{
  "type":"query_shard_exception",
  "reason":"failed to find type for field [shape]",
  "index_uuid":"mdXPYZ0XQNG1DIEOtDBgeg",
  "index":"osmgeopoints"
}
```

The index `osmgeopoints` is not defined for the track `geoshape`, so this appears to be a leftover from a previous track. This issue occurs when running `--test-mode`, as well as local full runs if the previous track was `geopoint`.

This PR tries to delete the `osmgeopoints` index, and the `osmgeoshapes` index (which might also exist, and would cause the same error presumably, since it also has no `shape` field).

### DIFF
--- a/geoshape/challenges/default.json
+++ b/geoshape/challenges/default.json
@@ -8,6 +8,18 @@
           "tags": ["setup"]
         },
         {
+          "name": "delete-index-osmgeopoints",
+          "operation": "delete-index",
+          "index": "osmgeopoints",
+          "tags": ["setup"]
+        },
+        {
+          "name": "delete-index-osmgeoshapes",
+          "operation": "delete-index",
+          "index": "osmgeoshapes",
+          "tags": ["setup"]
+        },
+        {
           "operation": {
             "operation-type": "create-index",
             "settings": {{index_settings | default({}) | tojson}}
--- a/geoshape/challenges/default.json
+++ b/geoshape/challenges/default.json
@@ -8,15 +8,19 @@
           "tags": ["setup"]
         },
         {
-          "name": "delete-index-osmgeopoints",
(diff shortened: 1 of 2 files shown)


gbanasiak (commented):
IMO it would be more logical to delete entire `osm*` wildcard in this case which should be possible, see example in https://esrally.readthedocs.io/en/stable/track.html#delete-index. Otherwise if any other `osm*` index gets added, this would have to be revisited potentially. WDYT?

Deletes spanning more than the track indices is a precedent. I think it deserves a mention in the README. I'm thinking about `benchmark-only` use against an existing cluster with Rally-unrelated data.

craigtaverner:
Deleting `osm*` would work for this case. My reason for keeping it separate is that I've also, on occasion, relied on having the tracks delete only their own indices. For example, what I've done before locally was to run all three tracks, one at a time, against a locally running ES instance, after which I would use Kibana to test queries against each of the indices, or combinations of indices, before adding those queries to the benchmarks. This requires that I can make a track delete only it's own indices. To enable this case, I will add tags to the external index deletion commands.

gbanasiak:
I've missed the fact that wildcards are not supported in serverless, so we went back to individual indices in https://github.com/elastic/rally-tracks/pull/605/commits/f83b97bab149dd19be09fec23bb3af1102ae67fb.

gbanasiak (approved):
Thanks for iterating. LGTM.

PR #606, opened 2024-05-14
Use a pipeline to generate doc id and use correct doc id values

salvatore-campagna:
Indexing documents in standard mode results in documents
whose document id is random. Random document ids prevent
us from testing `get` and `mget` since every indexing operation
would result in different document ids. Here we use a pipeline
taking advantage of the fingerprint processor to generate consistent
document id and use some of those values as parameters for
testing `get` and `mget`.

This PR needs to be backported to Elasticsearch `8.13` and `8.14`
since we need to run benchmarks for those versions too so that we can
evaluate the impact of ZSTD on such APIs.

### DIFF
--- a/tsdb/challenges/default.json
+++ b/tsdb/challenges/default.json
@@ -331,6 +331,14 @@
         },
         {%- endif %}
         {
+          "operation": "delete-doc-id-pipeline",
+          "tags": ["setup"]
+        },
+        {
+          "operation": "create-doc-id-pipeline",
+          "tags": ["setup"]
+        },
+        {
           "operation": "index",
           "warmup-time-period": 240,
           "clients": {{bulk_indexing_clients | default(8)}},
--- a/tsdb/challenges/default.json
+++ b/tsdb/challenges/default.json
@@ -339,10 +339,10 @@
           "tags": ["setup"]
         },
         {
-          "operation": "index",
+          "operation": "index-with-doc-id-gen-pipeline",
           "warmup-time-period": 240,
-          "clients": {{bulk_indexing_clients | default(8)}},
-          "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
+          "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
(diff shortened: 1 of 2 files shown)


martijnvg (approved):
lgtm

gareth-ellis (approved):
LGTM

salvatore-campagna:
@gareth-ellis @martijnvg I gated execution of operations which might fail if executed in `time_series` mode. That challenge is expected to be executed only in `standard` index mode.

tsdb/challenges/default.json
```diff
@@ -364,17 +372,18 @@
         {
           "name": "refresh-after-force-merge",
           "operation": "refresh"
-        },
+        }
         {# serverless-post-ingest-sleep-marker-start #}{%- if post_ingest_sleep|default(false) -%}
-        {
+        ,{
```
gareth-ellis, line 377: NIT, but can we have the commas trailing rather than leading?

tsdb/challenges/default.json
```diff
@@ -364,17 +372,18 @@
...
           "operation": {
             "operation-type": "sleep",
             "duration": {{ post_ingest_sleep_duration|default(30) }}
           }
-        },
+        }
         {%- endif -%}{# serverless-post-ingest-sleep-marker-end #}
-        {
+        {% if index_mode | default('time_series') is equalto 'standard' %}
+        ,{
```
gareth-ellis, line 386: And here

salvatore-campagna: If I do that running the challenge fails with parsing error due to additional commas depending on the if outcome.

gareth-ellis: Surely if both are changed, it should work fine. The ifs don't overlap, and the final item in the list is not dependent on an if statement. I can see that if one of those was changed it would cause issues, but changing both should work

gareth-ellis:
@elasticmachine  update branch

gareth-ellis:
@elasticmachine update branch

PR #607, opened 2024-05-14
Add GitHub Archive track redo

inqueue:
Add a new track with three challenges using https://www.gharchive.org/, formerly https://github.com/elastic/rally-tracks/pull/340.

#### Challenges

1. `index-and-search` (default)
2. `index-only`
3. `parallel-indexing-and-search`

This PR is meant to be an entry point for the track, after which it will be enhanced for heavy use cases.


gbanasiak (commented):
This looks good. I've left just one comment to clarify serverless status of this track.

PR #603, opened 2024-05-08
Disable lossy _source parameters on serverless

dpifke-elastic:
No longer allowed as of elastic/elasticsearch-serverless#1679.

PR #608, opened 2024-05-14
Improvements to the categorize-text challenge (#594)

edsavage:
* bulk import index data
* target only those indices that have the required message field
* remove code duplication in the challenge definition
* Reduce default number of iterations
* parameterize number of warmup iterations and measurable iterations
* Set default number of warmup iterations to 3, actual iterations 10.

Backport #594

### DIFF
--- a/cohere_vector/index-vectors-only-mapping.json
+++ b/cohere_vector/index-vectors-only-mapping.json
@@ -10,9 +10,11 @@
   },
   "mappings": {
     "dynamic": false,
+    {%- if build_flavor != "serverless" -%}
     "_source": {
       "enabled": false
     },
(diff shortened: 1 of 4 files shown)


edsavage:
The failing Serverless integration test looks to be due to https://github.com/elastic/elasticsearch-serverless/pull/1679. This has been addressed in `master` by https://github.com/elastic/rally-tracks/pull/603. Should #603 be backported to `8.13`?

gareth-ellis:
Yes, bring in #603 too, then it should be good. It needs backporting to 8.13, though we are in the process of getting the backport utility to do that automatically

gareth-ellis (approved):

PR #610, opened 2024-05-16
Add recall and NDCG operations in msmarco-v2-vector

jimczi:
This change adds an operation called knn-recall that computes the following metrics:
  * Recall
  * NDCG
  * Avg number of nodes visited during search

The new `queries-recall.json` file contains all the queries (76 in total) from the testing set along with their embeddings and the top 1000 ids computed with brute force over the entire corpus.
For the relevance metrics, the `qrels.tsv` file contains annotations for all the queries listed in `queries-recall.json`. This file is generated from the original training data available at [ir_datasets/msmarco_passage_v2](https://ir-datasets.com/msmarco-passage-v2.html#msmarco-passage-v2/trec-dl-2022/judged).


msmarco-v2-vector/README.md
```diff
@@ -42,14 +41,24 @@ This will build 47 `cohere-documents-XX.json` file for the entire dataset of 138
...
+```
+
+### Generating the queries for the recall operation
+
+The `queries-recall.json` can be rebuilt using `_tools/parse_queries.py -r`, this will load the msmarco v2 passages test queries, and then call the Cohere embed API for each query, store the embeddings in `queries-recall.json` as well as the true top 1000 for each query computed with brute force.
+For the relevance metrics, the `qrels.tsv` file contains annotations for all the queries listed in `queries.json`. This file is copied from the original training data available at [msmarco-passage-v2/trec-dl-2022/judged](https://ir-datasets.com/msmarco-passage-v2.html#msmarco-passage-v2).
+
+```console
+$ export COHERE_API_KEY='abcdefghijklmnopqrstuvwxyz'
+$ python _tools/parse_queries.py -t
```
1stvamp, line 61:
```suggestion
$ python _tools/parse_queries.py -r
```

msmarco-v2-vector/_tools/parse_queries.py
```diff
@@ -42,10 +58,41 @@ async def output_queries(queries_file):
...
+async def create_recall_queries():
+    with open(OUTPUT_RECALL_FILENAME, "w") as queries_file:
+        await output_recall_queries(queries_file)
+
+
 if __name__ == "__main__":
-    asyncio.run(main())
+    parser = argparse.ArgumentParser(description="Create queries for throughput or recall operations")
+    parser.add_argument("-t", "--throughput", help="Create queries for throughput operations", action="store_true")
+    parser.add_argument("-r", "--recall", help="Create queries for recall operations", action="store_true")
```
1stvamp, line 92:
```suggestion
    parser.add_argument("-r", "--recall", help="Create queries for recall operations", action="store_true")
    
       if len(sys.argv) == 1:
           # Neither -t or -r was called, show the options
           parser.print_help(sys.stderr)
```

msmarco-v2-vector/_tools/parse_queries.py
```diff
@@ -1,3 +1,4 @@
+import argparse
 import asyncio
 import json
```
1stvamp, line 3:
```suggestion
import json
import sys
```

cohere_vector/index-vectors-only-mapping.json
```diff
@@ -11,7 +11,7 @@
   "mappings": {
     "dynamic": false,
     "_source": {
-      "enabled": false
+      "mode": "synthetic"
```
afoucret, line 14: Should this be a param?

msmarco-v2-vector/challenges/default.json
```diff
@@ -55,26 +55,166 @@
     {
       "name": "standalone-search-knn-10-100-single-client",
       "operation": "knn-search-10-100",
-      "warmup-iterations": 100,
+      "warmup-iterations": 1000,
```
afoucret, line 58: I am curious for the reason of changing the warmup config?

jimczi: Variations in the result so mainly to stabilise the report.

msmarco-v2-vector/_tools/parse_queries.py
```diff
@@ -42,10 +59,45 @@ async def output_queries(queries_file):
...
+    async with AsyncElasticsearch(
+        "https://localhost:19200/", basic_auth=("esbench", "super-secret-password"), verify_certs=False, request_timeout=REQUEST_TIMEOUT
+    ) as es:
+        dataset = ir_datasets.load("msmarco-passage-v2/trec-dl-2022/judged")
+        async with AsyncClient(environ["COHERE_API_KEY"]) as co:
+            count = 0
+            for query in dataset.queries_iter():
+                emb = await retrieve_embed_for_query(co, query[1])
+                resp = await es.search(
+                    index="msmarco-v2", query=get_brute_force_query(emb), size=1000, _source=["_none_"], fields=["docid"]
```
afoucret, line 72: Maybe a param?

msmarco-v2-vector/operations/default.json
```diff
@@ -22,20 +22,34 @@
   "name": "parallel-documents-indexing",
   "operation-type": "bulk",
   "corpora": "msmarco-v2-parallel-indexing",
-  "bulk-size": {{parallel_indexing_bulk_size | default(500)}},
+  "bulk-size": {{parallel_indexing_bulk_size | default(50)}},
   "ingest-percentage": {{parallel_indexing_ingest_percentage | default(100)}}
-},
+}
+{%- set p_search_ops = (search_ops | default([(10, 20, 0), (10, 20, 20), (10, 50, 0), (10, 50, 20), (10, 100, 0), (10, 100, 20), (10, 200, 0), (10, 200, 20), (10, 500, 0), (10, 500, 20), (10, 1000, 0), (10, 1000, 20), (100, 120, 0), (100, 120, 120), (100, 200, 0), (100, 200, 120), (100, 500, 0), (100, 500, 120), (100, 1000, 0), (100, 1000, 120)]))%}
+{%- for i in range(p_search_ops|length) %},
```
afoucret, line 29: Just a nit but I would probably tried to keep the loop logic into the challenge schedule to generate tasks instead of generating operations.

It would mean loosing the naming of the operations but I do not think it is really important since we can refer on task name instead.

msmarco-v2-vector/operations/default.json
```diff
@@ -22,20 +22,34 @@
...
+    "name": "knn-recall-{{p_search_ops[i][0]}}-{{p_search_ops[i][1]}}"
+  {%- endif -%},
+  "operation-type": "knn-recall",
+  "param-source": "knn-recall-param-source",
+  "k": {{p_search_ops[i][0]}},
+  "num-candidates": {{p_search_ops[i][1]}},
+  "num-rescore": {{p_search_ops[i][2]}},
+  "include-in-reporting": false
 }
+{%- endfor %}
```
afoucret, line 55: Nit: newline

msmarco-v2-vector/track.py
```diff
@@ -51,5 +136,103 @@ def params(self):
...
+        return {
+            "index": self._index_name,
+            "cache": self._params.get("cache", False),
+            "size": self._params.get("k", 10),
+            "num_candidates": self._params.get("num-candidates", 100),
+            "num_rescore": self._params.get("num-rescore", 0),
+        }
+
+
+class KnnRecallRunner:
```
afoucret, line 164: This runner compute the relevance metrics on two different "judgment lists":
- ndcg is computed from qrels
- recall is computed from the exact knn as a source of truth

It can be misleading and it took me a while to understand why you can not rely on the qrel to compute the recall. 
My understanding is that you are interested in the differences between exact and approx knn search for the recall.

I would personally split it using two runners at least in a follow up. I would also compute a recall from the judgement list.

afoucret (approved):
Few comment but nothing that would prevent you to merge the PR

msmarco-v2-vector/track.py
jimczi: Discussed offline, we'll address the cleanup in a follow up.

gareth-ellis:
@jimczi  should this be backported to 8.15? I tried to backport #708 due to it adding about 50 mins to IT tests, but it seems that this PR was never backported to 8.15, so the changes are only in master (By default rally will choose the 8.15 branch when benchmarking against 8.X, where the version being tested is 8.15 or later - serverless always runs from master)

jimczi:
Sorry for the delay here @gareth-ellis .
Are you trying to add this challenge somewhere? I can do the backport but I am not sure I understand whether that's what you imply here?

PR #612, opened 2024-05-22
Prepare elastic/logs track for the new logs index mode.

martijnvg:
By replacing index_sorting and synthetic_source_mode parameters (added via #522) by the index_mode parameter.

Closes elastic/elasticsearch#107222

PR #616, opened 2024-06-05
[8.13] Prepare elastic/logs track for the new logs index mode. (#612)

gareth-ellis:
# Backport

This will backport the following commits from `master` to `8.13`:
 - [Prepare elastic/logs track for the new logs index mode. (#612)](https://github.com/elastic/rally-tracks/pull/612)

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)


martijnvg (approved):
👍

martijnvg:
I may have  approved this change to soon. This backport the change to 8.13, which means this then runs against Elasticsearch 8.13? The new logs index mode has only been added to main and will be available with Elasticsearch 8.15.0 as tech preview. So I think setting index_mode to logs would fail here?

gareth-ellis:
Ah, ok!. In which case perhaps we need to cut a new 8.15 branch and use that.

martijnvg:
> Ah, ok!. In which case perhaps we need to cut a new 8.15 branch and use that.

Yes, I think that would be the right fix.

PR #617, opened 2024-06-05
Use documented track param parallel_indexing_target_throughput

wangch079:
The [Wikipedia](https://github.com/elastic/rally-tracks/tree/master/wikipedia) track documented that it supported track param `parallel_indexing_target_throughput` but it doesn't. 

This PR updates the track to use the param `parallel_indexing_target_throughput`

### DIFF
--- a/wikipedia/challenges/default.json
+++ b/wikipedia/challenges/default.json
@@ -71,7 +71,8 @@
             "operation": "query-string-search",
             "clients": {{ parallel_indexing_search_clients | default(20) | int }},
             "time-period": {{ parallel_indexing_search_time_period | default(300) | int }},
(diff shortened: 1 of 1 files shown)


gareth-ellis (commented):
I see that this is following the default value in the README, however you should be aware that this will possibly change the results of the benchmark. The number for target throughput is a total requests per second, across all clients, not a percentage. You could add some logic that checks if the parameter has been set, and if so include target-throughput, otherwise exclude it

wangch079:
>The number for target throughput is a total requests per second, across all clients, not a percentage.

This is my understanding as well. If we use the default value for `parallel_indexing_search_clients` (20), and `parallel_indexing_target_throughput` (100), each client will try to send 5 requests per second.

PR #618, opened 2024-06-13
[ES-8187] ES|QL elastic/logs scale testing challenge

favilo:
Adding special queries specifically for scale testing ESQL queries


bpintea (commented):
I've left some comments mostly concerning queries that might no longer be needed.
I think it'd be great if we only have those queries present that we want timing details about.

gbanasiak:
@favilo I've modified queries in https://github.com/elastic/rally-tracks/pull/618/commits/3747188c588884c41c998f8de705c747e8f5f958 to reduce response time but hard-coded timestamp is problematic. I don't think this should land in `master` branch.

craigtaverner (approved):
Approved, but I wonder were all four of these really long running? If not, we should not push the iteration count down for fast queries.

gbanasiak:
> I wonder were all four of these really long running?

Yes.

gbanasiak:
https://github.com/elastic/rally-tracks/pull/705 includes the queries from this PR, so I'll rework it.

PR #620, opened 2024-06-18
Make exclude conditional on not serverless

gareth-ellis:
excludes is not available in serverless, thus index creation currently fails. This excludes the excludes on serverless


maxhniebergall (approved):
LGTM, thanks Gareth!

PR #622, opened 2024-06-19
Add three searches that fetch _source as separate challenges to elastic/logs track

martijnvg:
These search are based from searches that are executed by search/discovery search challenge that fetch top documents. The queries are without query and fetch 100, 500 and 1000 documents. The source is fetches using the field fetch feature, like in the searches that execute as part of search/discovery workflow.

The problem is that we see combined latency and service time for search/discovery challenge and not for each search that is executed as part of this challenge (it is composite operation). 

By adding these searches we can get latency and service time of searches that specifically fetch _source. This information is useful for the logsdb effort. Synthetic source makes fetching _source more expensive, but currently we can't introspect at a closer level what the impact is (since search/discovery search challenge report latency / service time for multiple operations).

### DIFF
--- a/elastic/logs/challenges/logging-querying.json
+++ b/elastic/logs/challenges/logging-querying.json
@@ -82,20 +82,20 @@
     {
       "operation": "discovery-search-request-size-100",
       "clients": 1,
-      "warmup-iterations": 10,
+      "warmup-iterations": 0,
       "iterations": 50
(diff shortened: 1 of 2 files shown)


martijnvg:
buildkite test this

martijnvg:
buildkite test this

PR #626, opened 2024-07-03
Added many more distance benchmarks for ESQL

craigtaverner:
Most of these are optimized in main, except two, the ESQL versions of sorting by distance, which we've set to one iteration for now to save on benchmarking time before we do performance optimizations on those.


iverase (approved):
LGTM

PR #627, opened 2024-07-03
Add query rule tests to the wikipedia track

kderusso:
Adds some tests for query rules of various sizes to the Wikipedia track.


wikipedia/track.py
```diff
@@ -113,7 +121,60 @@ def params(self):
...
+    def params(self):
+        try:
+            query = next(self._queries_iterator)
+            # TODO Update this to use current syntax with 8.15.0+
+            return {
+                "method": "POST",
+                "path": "/_search",
+                "body": {
+                    "query": {
+                        "rule": {
```
kderusso, line 161: Note: This syntax will only work for versions 8.15.0+ - is there anything else we should do for these challenges?

afoucret (approved):

PR #628, opened 2024-07-15
Update elastic/logs integrations to 8.13.3 (#613)

ebadyano:
backport for #613 to 8.15


gbanasiak (approved):
LGTM.

I've noticed we still define `number_of_shards` in 2 places although I don't think we should for serverless:
```
component/track-custom-shared-settings.json <-- OK
8:        "number_of_shards": "{{ number_of_shards | default(1) }}",

component/auditbeat-mappings.json <-- NOT OK ?
13:        "number_of_shards": 1,

composable/logs-k8-application.log.json <-- NOT OK ?
16:        "number_of_shards": "{{ number_of_shards | default(1) }}",
```

PS. The above is just a BTW comment - we're not touching number of shards here.

ebadyano:
@gbanasiak I am blanking out a bit, I thought ` {# non-serverless-index-settings-marker-start #}{` was supposed to do something about the settings we dont want in serverless..but dont recall the details

gbanasiak:
The thing is we allow the restricted settings in `track-custom-shared-settings.json` when executed as operator user but only if explicitly asked for (`include_non_serverless_index_settings`):

https://github.com/elastic/rally-tracks/blob/25ca4ff1c4596bf3be83cfd923b6914eb46c3015/elastic/logs/templates/component/track-custom-shared-settings.json#L6-L11

But auditbeat and logs-k8-application have slipped through it seems, and will have `number_of_shards` configured when run as operator, regardless of the defaults applied by the platform. We should address that in a separate PR I think.

PR #632, opened 2024-07-25
New `large-logs-dataset` challenge in `elastic/logs`

### LINKED ISSUE #631: New `large-logs-dataset` challenge in `elastic/logs` track
We would like to run an experiment in Rally which uses considerable amount of data. The idea is to be able to fill the disk of an AWS instance with 7.5 TB of storage. Indexing such large amount of data poses at least two challenges, anyway, which are a result of the way the `elastic\logs` Rally track is designed:
1. Data generation needs to be done before running each experiment (note that we could do data generation once and the use it multiple times, but that would be possible only if multiple experiments are ok with using the same dataset).
2. Raw data to Json expansion means we need to generate far more Json data to index in to Elasticsearch. If we imagine a raw-to-json expansion factor of 10, it means that filling storage with 7.5 TB of data needs 75 TB of Json data on the Rally load driver.

For our experiment described in Jira at https://elasticco.atlassian.net/browse/ES-9013?focusedCommentId=152542 we:
1. Can't reuse the same dataset (we can't do data generation just once), because at least the document `@timestamp` needs to change depending on how much data we need to index per each day (`raw_data_volume_per_day`).
2. Finding an AWS instance with such large storage is challenging and expensive, one of them is `is4gen.8xlarge` that has 4 x 7.5 TB = 30 TB of storage available. Note that if we assume x10 raw-to-json expansion we would need 75 TB of Json data to have 7.5 TB of raw data. This means that even the instance with the largest storage can't handle the amount of data we need.

As a result, benchmarking this scenario is practically impossible because of resource constraints but also because of the time data generation and indexing would require (we estimated something like 1 week).

So the idea is to adopt the following strategy which we would like to implement in a new challenge part of the `elastic/logs` track:
1. Index 100 GB of raw data, which means generating about 1 TB of Json data on the load driver
2. Create a snapshot out of the indexed data
3. Restore the snapshot multiple times (ideally using a challenge parameter)
4. Execute queries part of the `logging-querying` existing challenge to collect query latencies

For the use case above where we need to fill the instance with 7.5 TB of raw data it means restoring the snapshot 75 times. We expect:
1. time required to have the full dataset indexed to be far less than generating the full dataset and indexing it as it would normally happen with the `elastic/logs` track and the `logging-querying` existing challenge.
2. queries to crunch more documents because of data duplication: that is ok as long as we just compare query latencies with other setups using the same track and challenge (i.e. "standard" index mode versus LogsDB)
3. lower storage footprint due to data duplication and better data compression

salvatore-campagna:
Introduce a new `large-logs-dataset` challenge to `elastic/logs` track which duplicates data indexed by restoring
a snapshot multiple times. The number of snapshot restore operations is controlled by the variable `snapshot_restore_counts` which by default has a value of `100`.

This would result in indexing `raw_data_volume_per_day` bytes multiplied by `snapshot_restore_counts`.
As an example if `raw_data_volume_per_day` is 50 GB then the index will have about 5 TB of raw data.
Note that the index, anyway, will include duplicated data.

This is meant to be used just as a fast way to increase the amount of data in an index skipping the expensive data
generation and indexing process.

Resolves #631

### DIFF
--- a/elastic/logs/track.json
+++ b/elastic/logs/track.json
@@ -42,9 +42,10 @@
 {% set p_restore_data_streams = (restore_data_streams | default("logs-*")) %}
 {% set p_snapshot_metadata = (snapshot_metadata | default({}))%}
 {% set p_snapshot_rename_suffix = (snapshot_rename_suffix | default("") ) %}
+{% set p_snapshot_restore_counts = (snapshot_restore_counts | default(100) ) %}
 {% set p_dsl_poll_interval = (dsl_poll_interval | default(false) ) %}
 {% set p_dsl_default_rollover = (dsl_default_rollover | default(false) ) %}
-
+{% set p_skip_fleet_globals = (skip_fleet_globals | default(false) ) %}
 {% set p_worker_threads_enabled = (worker_threads_enabled | default(true))  %}
 
 {% set es_version = "7.13.2" %}
@@ -200,11 +201,13 @@
         "template-path": "component_template"
 
     },
+    {% if p_skip_fleet_globals == false %}
     {
         "name": ".fleet_globals-1",
         "template": "./templates/component/.fleet_globals-1.json",
         "template-path": "component_template"
     },
+    {% endif %}
     {
         "name": "logs-apache.access@package",
         "template": "./templates/component/logs-apache.access@package.json",
--- a/elastic/logs/challenges/large-logs-dataset.json
+++ b/elastic/logs/challenges/large-logs-dataset.json
@@ -122,9 +122,8 @@
             "rename_replacement": "$1{{ p_snapshot_rename_suffix }}-{{ snapshot_restore_iter }}"
           }
         }
-      } {{ ", " if not loop.last else "" }}
+      },
       {% endfor %}
-      ,
       {
         "name": "reset-recovery-settings-to-defaults",
         "operation": {
--- a/elastic/logs/challenges/large-logs-dataset.json
+++ b/elastic/logs/challenges/large-logs-dataset.json
@@ -113,7 +113,7 @@
           "operation-type": "restore-snapshot",
           "repository": {{ p_snapshot_repo_name | tojson }},
           "snapshot": {{ p_snapshot_name | tojson }},
-          "wait-for-completion": false,
+          "wait-for-completion": true,
           "body": {
             "indices": {{ p_restore_data_streams | tojson }},
             "ignore_unavailable": true,
--- a/elastic/logs/challenges/large-logs-dataset.json
+++ b/elastic/logs/challenges/large-logs-dataset.json
@@ -104,7 +104,7 @@
         "name": "refresh-after-force-merge",
         "index": "logs-*",
         "operation": "refresh"
-      }
+      },
       {%- endif %}
       {% for snapshot_restore_iter in range(p_snapshot_restore_counts) %}
       {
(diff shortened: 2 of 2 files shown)


elastic/logs/track.json
```diff
@@ -42,9 +42,10 @@
 {% set p_restore_data_streams = (restore_data_streams | default("logs-*")) %}
 {% set p_snapshot_metadata = (snapshot_metadata | default({}))%}
 {% set p_snapshot_rename_suffix = (snapshot_rename_suffix | default("") ) %}
+{% set p_snapshot_restore_counts = (snapshot_restore_counts | default(100) ) %}
 {% set p_dsl_poll_interval = (dsl_poll_interval | default(false) ) %}
 {% set p_dsl_default_rollover = (dsl_default_rollover | default(false) ) %}
-
+{% set p_skip_fleet_globals = (skip_fleet_globals | default(false) ) %}
```
salvatore-campagna, line 48: I just copied this from another PR. Will eventually remove after the other PR is merged.
This is necessary to avoid an error which happens when deleting component templates.

salvatore-campagna:
@elastic/es-perf at the moment this challenge is failing with the following error:

```
esrally.exceptions.RallyAssertionError: Request returned an error. Error type: api, Description: repository_verification_exception ({'error': {'root_cause': [{'type': 'repository_verification_exception', 'reason': '[logging] path [observability/logging] is not accessible on master node'}], 'type': 'repository_verification_exception', 'reason': '[logging] path [observability/logging] is not accessible on master node', 'caused_by': {'type': 'i_o_exception', 'reason': 'Unable to upload object [observability/logging/tests-CcJdFLNpQ0SwoHLxwOkv7w/master.dat] using a single upload', 'caused_by': {'type': 'amazon_s3_exception', 'reason': 'Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: SZ54XD85DV32K3RX; S3 Extended Request ID: WdDrng9Ah0EmndL4q6y58b+vTMxCCmKgdffmUIALUXcSqpMrjAJS7enlwhkzkqWOQf3oYCW7ENg=; Proxy: null)'}}}, 'status': 500}), HTTP Status: 500
```

any idea how to fix it? It looks like some S3 configuration might be missing.

salvatore-campagna:
I tried running another challenge in the `elastic/logs` track which uses snapshots and s3 and I see the same issue. I guess this means my user is lacking some permission.

```
esrally.exceptions.RallyError: Cannot run task [register-snapshot-repository]: Request returned an error. Error type: api, Description: repository_verification_exception ({'error': {'root_cause': [{'type': 'repository_verification_exception', 'reason': '[logging] path [observability/logging] is not accessible on master node'}], 'type': 'repository_verification_exception', 'reason': '[logging] path [observability/logging] is not accessible on master node', 'caused_by': {'type': 'i_o_exception', 'reason': 'Unable to upload object [observability/logging/tests-R6KKPdvrTTK3FnuBeibL1w/master.dat] using a single upload', 'caused_by': {'type': 'amazon_s3_exception', 'reason': 'Access Denied (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: SWVHM6P139B4H9J7; S3 Extended Request ID: wsLQpvhYYCqMvFppInjYTLzXO83vxMvpZy3kIqlmKgRyZ88cvUCLOex1tFGOuHcro8zcjFAFzA1/WqO2RGCFwHvHS7s4ODV1; Proxy: null)'}}}, 'status': 500}), HTTP Status: 500
```

inqueue:
Closing for now.

PR #633, opened 2024-07-30
Update msmarco-passage-ranking track to use sparse_vector instead of weighted_tokens

kderusso:
Example command to test: 
```
esrally race --kill-running-processes --preserve-install --track-path=msmarco-passage-ranking --on-error=abort  --test-mode
```
Requires 8.15.0 or higher


markjhoy (approved):
👍

PR #635, opened 2024-08-02
Add exclude to query ruleset benchmarks

kderusso:
Adds `exclude` query rules to our query ruleset benchmarking, and loads additional random IDs to include in rulesets for query building. 

Note: Requires `8.16.0-SNAPSHOT` or later. 

Sample command to run this: 
```
esrally race --kill-running-processes --preserve-install --track-path=wikipedia --on-error=abort --car="4gheap,trial-license"  --test-mode
```

### DIFF
--- a/wikipedia/track.py
+++ b/wikipedia/track.py
@@ -197,7 +197,7 @@ class PinnedSearchParamSource(QueryIteratorParamSource):
                     "query": {
                         "pinned": {
                             "organic": {"query_string": {"query": query, "default_field": self._params["search-fields"]}},
-                            "ids": [self.ids],
+                            "ids": [random.choice(self.ids)],
                         }
                     },
                     "size": self._params["size"],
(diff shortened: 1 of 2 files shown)


pquentin:
buildkite test this please

wikipedia/track.py
```diff
@@ -130,14 +140,14 @@ def partition(self, partition_index, total_partitions):
...
 
     def params(self):
+        ids = ids_samples()
         rules = []
         for i in range(self.query_ruleset_params.ruleset_size):
-            # Note: `AccessibleComputing` is one of the _ids indexed in the dataset
             rule = {
                 "rule_id": "rule_{{i}}",
-                "type": "pinned",
+                "type": random.choice(["pinned", "exclude"]),
```
Samiul-TheSoccerFan, line 148: Learning purposes: We are choosing either one in this scenario. Do we need rule type where both of these presents similar to your API tests (If I remember correctly)? Or it is avoided because having both does not make any impact on the performance?

Samiul-TheSoccerFan (approved):
Not a python expert but LGTM.

wikipedia/track.py
kderusso: The idea is that there will likely be multiple matching rules, so this is just trying to get a read on performance.

Mikep86 (approved):
LGTM!

kderusso:
buildkite test this please

PR #642, opened 2024-08-28
Allow skipping delete of fleet component template

salvatore-campagna:
The fleet component template is used when we try to delete it. Here we introduce a parameter that allows us to skip deletion of the component template. The default value is false, which means normally we attempt to delete it. Setting it explicitly to true we avoid deleting it. This prevents errors happening if we try to delete it and it is in use.

This needs to be backported to Elasticsearch 8.15.


gareth-ellis (approved):
LGTM

PR #654, opened 2024-09-05
Remove `index.mode` gate on `copy_to`

salvatore-campagna:
This PR will be used to test an implementation of Elasrticsearch that supports copy_to for text fields. We make the field stored to mimic the actual copy_to implementation which will later use a stored field.

### DIFF
--- a/elastic/security/README.md
+++ b/elastic/security/README.md
@@ -99,6 +99,7 @@ The following parameters are available:
 * `number_of_replicas` (default: 1) - The number of replicas to set per Data Stream. The same value is used for all Data Streams.
 * `bulk_indexing_clients` (default: 8) - The number of clients issuing indexing requests.
 * `bulk_size` (default: 50) - The number of documents to send per indexing request.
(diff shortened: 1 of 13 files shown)


salvatore-campagna:
NOTE: the CI is faling with an error that is expected with Elasticsearch 8.15 (missing `copy_to` support for `text` fields in Synthetic Source. This PR, anyway, is tested against an Elasticsearch PR which works-around that issue.

gareth-ellis:
OK, will this change then be in 8.16? If so we should wait for the feature to merged into elasticsearch main, then backport this to a new 8.16 branch

salvatore-campagna:
> OK, will this change then be in 8.16? If so we should wait for the feature to merged into elasticsearch main, then backport this to a new 8.16 branch

This is using a workaround that I just need for testing purposes. The actual fix will not require setting `store: true`. So once `copy_to` is supported it will end-up in Elasticsearch main and I will update this PR so to remove `store:true`. In the meanwhile I will use it like it is now.

salvatore-campagna:
In order to get the original behaviour including copy_to I will just revert the original PR, merge and backport it.

PR #655, opened 2024-09-06
Keep array source in logsdb mode

kkrik-es:
Test this change in nightlies, to decide whether to enable it by default.

Related to https://github.com/elastic/elasticsearch/issues/112354


kkrik-es:
run buildkite/rally-tracks-it-serverless

PR #657, opened 2024-09-09
Include ability to force merge segments in `elastic/security`

salvatore-campagna:
Force merging allows us to compare segment size and overall storage size reasoning about
storage saving. The security track is missing a way to force merge.

We are not going to set this parameter in nightlies but we still would like to have it for ad-hoc runs, since it makes reasoning about storage footprint easier.

This needs back-porting to Rally `8.15` branch.


salvatore-campagna:
I tested this running it with a stateful configuration and it works.

gareth-ellis (approved):
LGTM, Since the default action is to not run, I don't think we need the check for whether we are running serverless or not

martijnvg (approved):
lgtm

PR #658, opened 2024-09-10
Modify the default vector index type to int8_hnsw

mayya-sharipova:
From ES v8.14 the default index type for dense_vectors is int8_hnsw. 
This modifies our rally tracks to reflect it.


gareth-ellis (approved):
LGTM

PR #664, opened 2024-09-16
Modify the default vector index type to int8_hnsw (#658)

mayya-sharipova:
From ES v8.14 the default index type for dense_vectors is int8_hnsw. 
This modifies our rally tracks to refect it.

Backport for #658


AI-IshanBhatt (approved):
LGTM

PR #668, opened 2024-09-18
Add standard component templates to security track composable templates

achuguy:
Adding logs@settings, logs@mappings, ecs@mappings to the composable templates of the endpoint indices in the security track.

### DIFF
--- a/elastic/security/templates/composable/.logs-endpoint.action.responses.json
+++ b/elastic/security/templates/composable/.logs-endpoint.action.responses.json
@@ -21,6 +21,9 @@
       }
     },
     "composed_of": [
+      "logs@mappings",
+      "logs@settings",
+      "ecs@mappings",
       ".logs-endpoint.action.responses@package",
       ".logs-endpoint.action.responses@custom",
       ".fleet_globals-1",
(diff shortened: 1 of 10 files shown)


elastic/security/templates/composable/.logs-endpoint.action.responses.json
```diff
@@ -21,6 +21,9 @@
       }
     },
     "composed_of": [
+      "logs@mappings",
```
salvatore-campagna, line 24: I see standard component templates are added here and used by all existing templates...I don't see them as part of the track. This means that we rely on KIbana to install them?

gbanasiak: AFAIU the component templates added are installed by ES.

gbanasiak (approved):

elastic/security/templates/composable/.logs-endpoint.action.responses.json
achuguy: Yes, install these particular templates and with Kibana included in the security track we do not have a mechanism to  distinguish if these templates are not installed if Kibana is not used.

PR #672, opened 2024-09-19
Add logsdb support to http_logs track

martijnvg:
This PR adds the following track params:
* `enable_logsdb` (default: false) Determines whether the logsdb index mode gets used. If set then index sorting is configured to only use `@timestamp` field and the `source_enabled` parameter will have no effect.
* `force_merge_max_num_segments` (default: unset): An integer specifying the max amount of segments the force-merge operation should use.

### DIFF
--- a/http_logs/index.json
+++ b/http_logs/index.json
@@ -18,9 +18,12 @@
   },
   "mappings": {
     "dynamic": "strict",
+    {% if enable_logsdb %}
+    {%- else %}
     "_source": {
       "enabled": {{ source_enabled | default(true) | tojson }}
     },
+    {%- endif %}
     "properties": {
       "@timestamp": {
         {%- if ingest_pipeline is defined and ingest_pipeline == "grok" %}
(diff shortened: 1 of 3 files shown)


http_logs/index.json
```diff
@@ -3,14 +3,23 @@
     {# non-serverless-index-settings-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%}
     "index.number_of_shards": {{ number_of_shards | default(5) }},
     "index.number_of_replicas": {{ number_of_replicas | default(0) }},
-    "index.requests.cache.enable": false
+    "index.requests.cache.enable": false{%- if enable_logsdb %},{%- endif %}
```
inqueue, line 6: This should not be needed if the `,` is placed at the end of line 8.

```
{%- if enable_logsdb %},
```

```suggestion
    "index.requests.cache.enable": false{%- if enable_logsdb %},{%- endif %}
```
```suggestion
    "index.requests.cache.enable": false
```

inqueue (approved):
Made a comment, otherwise LGTM

inqueue (approved):
:shipit:

PR #677, opened 2024-09-23
Configure strict dynamic mapping

salvatore-campagna:
This PR is expected to be used just for a test in order to validate
the new `elastic/security` mappings.

### DIFF
--- a/elastic/security/templates/composable-logsdb/logs-endpoint.alerts.json
+++ b/elastic/security/templates/composable-logsdb/logs-endpoint.alerts.json
@@ -7,6 +7,7 @@
     "template": {
       "settings": {},
       "mappings": {
+        "dynamic": "strict",
         "_meta": {
           "package": {
             "name": "endpoint"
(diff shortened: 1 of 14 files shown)


salvatore-campagna:
I tested the `elastic/security` benchmark applying `dynamic: strict` as a way for us to validate that we can index a quite old dataset using up-to-date mappings coming from templates installed by Kibana. It is possible, indeed that changes into templates make indexing of the dataset not correct.

Using `dynamic: strict` I see messages like the following one in `.rally/logs/rally.log`

```
2024-09-24 09:18:49,585 ActorAddr-(T|:45557)/PID:36085 esrally.driver.runner WARNING Bulk request failed: [HTTP status: 400, message: [1:22] mapping set to strict, dynamic introduction of [hostname] within [agent] is not allowed]
```

I wrote a script (attached) to extract all lines like that and make sure to capture all fields which are dynamically mapped as a result of not being mapped using templates. As of today I only see field `agent.hostname` being dynamically mapped. Note that this field is not the same as `host.hostname` used for sorting purposes when `index.mode` is set to `logsdb`.
See PR #678 where we apply index sorting when LogsDB is used.

PR #678, opened 2024-09-23
`host.name` is empty we need to use `host.hostname`

salvatore-campagna:
If the `host.name` field does not exists, indices created as backing indices of a data stream
are injected with empty values of `host.name`. Sorting on `host.name` and `@timestamp`
results in sorting just on `@timestamp`. Looking at some mappings I see a `host.hostname`
exists. Also a cardinality aggregation results in hundreds of distinct values which suggests
the filed is not empty.

We would like to test using a meaningful combination of fields to sort on. Ideally we expect
better benchmark results despite being possible that other, more effective, combinations of
fields might exist. We are interested, anyway, in changes over time **given a valid set of fields
to sort on**.

This needs backporting to `8.15`

### DIFF
--- a/elastic/security/templates/component/track-shared-logsdb-mode.json
+++ b/elastic/security/templates/component/track-shared-logsdb-mode.json
@@ -1,17 +1,14 @@
 {
-  "template": {
-    "settings": {
-      {% if index_mode %}
-      "index": {
-          "mode": {{ index_mode | tojson }},
-          "sort": {
-          "order": [
-            { "host.hostname": "desc" },
-            { "@timestamp": "desc" }
-          ]
+    "template": {
+      "settings": {
+        {% if index_mode %}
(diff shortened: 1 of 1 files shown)


gbanasiak (approved):

gbanasiak:
Review notes:
* `host.hostname` field seems to be present in all `endpoint` documents (checked `file` and `process`),
* the cardinality of this field is 100,
* value format: `<english-first-name>.[linux|windows]`.

salvatore-campagna:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|8.15|[](https://github.com/elastic/rally-tracks/pull/679)|


### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sqren/backport)

PR #681, opened 2024-09-25
Continue backport on error

gbanasiak:
When backport does not find a branch to backport to it returns with an error failing GH action step which in turn fails entire job (see example in https://github.com/elastic/rally-tracks/actions/runs/11019352166). This will happen frequently as not every PR will be labeled with `backport-to-*`.

The intention here is to prevent such error from failing entire job without repeating ourselves, i.e. including all possible `backport-to-*` labels in the `if` statement of the job. The list of backport-allowed branches is already defined in `targetBranchChoices` in `.backportrc.json`.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -15,6 +15,7 @@ jobs:
     steps:
       - name: Backport Action
         uses: sorenlouv/backport-github-action@v9.5.1
+        continue-on-error: true
         with:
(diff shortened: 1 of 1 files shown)


favilo (approved):

gbanasiak:
With this change `failure()` condition in the `Debug log` step will never be true, but I'm leaving this as is for now.

PR #682, opened 2024-09-30
Include a variable to control `synthetic_source_keep` parameter

salvatore-campagna:
This PR introduces a new track parameter, `synthetic_source_keep` which is used to control the
behaviour of synthetic source for all field types. It can have values `none`, `arrays` or `all` (`all`
not usable when set at index level).
See https://github.com/elastic/elasticsearch/pull/112706 to understand the effect of each value.

Later on we will use this to change the behaviour in our nightlies and run benchmarks on both `elastic/logs`
and `elastic/security` using value `arrays`.

This needs backporting to `8.15`

### DIFF
--- a/elastic/logs/README.md
+++ b/elastic/logs/README.md
@@ -220,7 +220,7 @@ The following parameters are available:
 * `corpora_uri_base` (default: `https://rally-tracks.elastic.co`) - Specify the base location of the datasets used by this track.
 * `lifecycle` (default: unset to fall back on Serverless detection) - Specifies the lifecycle management feature to use for data streams. Use `ilm` for index lifecycle management or `dlm` for data lifecycle management. By default, `dlm` will be used for benchmarking Serverless Elasticsearch.
 * `workflow-request-cache` (default: `true`) - Explicit control of request cache query parameter in searches executed in a workflow. This can be further overriden at an operation level with `request-cache` parameter.
-* `synthetic_source_keep` (default: unset) - Allows overriding the default synthetic source behaviour for all field types with the following values: `none` (equivalent to unset) - no source is stored, `arrays` - source stored as is only for multi-value (array) fields, `all` - source is stored as is for single-value and multi-value fields.
+* `synthetic_source_keep` (default: unset) - Allows overriding the default synthetic source behaviour for all field types with the following values: `none` (equivalent to unset) - no source is stored, `arrays` - source stored as is only for multi-value (array) fields.
 
 ### Data Download Parameters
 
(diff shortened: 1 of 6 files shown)


elastic/logs/README.md
```diff
@@ -220,6 +220,7 @@ The following parameters are available:
 * `corpora_uri_base` (default: `https://rally-tracks.elastic.co`) - Specify the base location of the datasets used by this track.
 * `lifecycle` (default: unset to fall back on Serverless detection) - Specifies the lifecycle management feature to use for data streams. Use `ilm` for index lifecycle management or `dlm` for data lifecycle management. By default, `dlm` will be used for benchmarking Serverless Elasticsearch.
 * `workflow-request-cache` (default: `true`) - Explicit control of request cache query parameter in searches executed in a workflow. This can be further overriden at an operation level with `request-cache` parameter.
+* `synthetic_source_keep` (default: unset) - Allows overriding the default synthetic source behaviour for all field types with the following values: `none` (equivalent to unset) - no source is stored, `arrays` - source stored as is only for multi-value (array) fields, `all` - source is stored as is for single-value and multi-value fields.
```
kkrik-es, line 223: `all` is not allowed at the index level, as it effectively disables synthetic source. The only values are `none`, same as default, and `arrays.

salvatore-campagna: I see

```
public static final Setting<SourceKeepMode> SYNTHETIC_SOURCE_KEEP_INDEX_SETTING = Setting.enumSetting(
        SourceKeepMode.class,
        "index.mapping.synthetic_source_keep",
        SourceKeepMode.NONE,
        value -> {
            if (value == SourceKeepMode.ALL) {
                throw new IllegalArgumentException("index.mapping.synthetic_source_keep can't be set to [" + value.toString() + "]");
            }
        },
        Setting.Property.IndexScope,
        Setting.Property.ServerlessPublic
    );
```

elastic/logs/track.json
```diff
@@ -21,6 +21,7 @@
 {% set p_query_time_period = (query_time_period | default(900)) %}
 {% set p_query_request_params = (query_request_params | default({}))%}
 {% set p_include_esql_queries = (include_esql_queries | default(build_flavor != "serverless")) %}
+{% set p_synthetic_source_keep = (synthetic_source_keep | default) %}
```
gbanasiak, line 24: That's unusual use of `default` function. Can you define the default value explicitly with `default(<SOME-VALUE>)`? It will be easier to read later on. I think this default resolves to an empty string which is not obvious.

elastic/security/track.json
```diff
@@ -9,6 +9,7 @@
 {% set p_number_of_shards = (number_of_shards | default(1)) %}
 {% set p_number_of_replicas = (number_of_replicas | default(1)) %}
 {% set p_skip_delete_component_template = (skip_delete_component_template | default(false) ) %}
+{% set p_synthetic_source_keep = (synthetic_source_keep | default) %}
```
gbanasiak, line 12: See above.

kkrik-es (approved):

elastic/logs/track.json
salvatore-campagna: I wanted to make explicit the fact that the default value is not set...maybe I am too explicit here?

gbanasiak: You can use `default("")` or resign from `p_synthetic_source_keep` similarly to `index_mode` which does not have a corresponding `p_index_mode`. Let's make those 2 behave similar.

gbanasiak (approved):

PR #691, opened 2024-10-11
Add search/mteb/dbpedia relevance evaluation track

kderusso:
Adds a `search/mteb/dbpedia` relevance evaluation track to Rally.


dbpedia_ranking/README.md
```diff
@@ -0,0 +1,48 @@
...
+  "_id": "<dbpedia:Animalia_(book)>",
+  "title": "Animalia (book)",
+  "text": "Animalia is an illustrated children's book by Graeme Base. It was originally published in 1986, followed by a tenth anniversary edition in 1996, and a 25th anniversary edition in 2012. Over three million copies have been sold.   A special numbered and signed anniversary edition was also published in 1996, with an embossed gold jacket."
+}
+```
+</details>
+
+### Example Query
+
+Queries are structured within a JSON array, where each individual object signifies a unique 'query' and its corresponding expansion achieved through ELSER v2, which is stored pre-computed in the 'text_expansion_elser' field.:
```
jimczi, line 28: ELSER v2 doesn't seem to be used in this track?

dbpedia_ranking/challenges/default.json
```diff
@@ -0,0 +1,66 @@
...
+      }
+    },
+    {
+      "name": "default-text-search-relevance",
+      "operation": "default-text-search-relevance"
+    },
+    {
+      "name": "english-analyzed-text-search-relevance",
+      "operation": "english-analyzed-text-search-relevance"
+    }
```
jimczi, line 64: Let's also report the search throughput to compare the cost of the strategies?

jimczi (commented):
The track LGTM.
Since it's a new dataset I wonder if that makes sense to create it under `search/mteb` though. This way we can share code between the mteb datasets. Aurelien has a full design to handle all datasets in the same track but we can start simple and just introduce the hierarchy? 
@afoucret WDYT?

dbpedia_ranking/README.md
kderusso: Good catch - copy paste error :)

kderusso:
> Since it's a new dataset I wonder if that makes sense to create it under search/mteb though

That makes sense, I will refactor the directories for now.

dbpedia_ranking/challenges/default.json
kderusso: Any idea why I'm only seeing 100 pctile latency when running this, instead of a breakdown?

search/mteb/dbpedia/operations/default.json
```diff
@@ -0,0 +1,52 @@
...
+   "text_boost": 1,
+   "track_total_hits": false,
+   "include-in-reporting": true
+},
+{
+   "name": "default-text-search-relevance",
+   "operation-type": "text_search_relevance",
+   "param-source": "relevance_param_source",
+   "query_source": "queries.json",
+   "qrels_source": "test.tsv",
```
jimczi, line 32: Do we need the `dev.tsv` file or should it be removed?

jimczi (approved):
LGTM

search/mteb/dbpedia/operations/default.json
kderusso: I will remove it.

PR #693, opened 2024-10-17
Added two more distancefilter benchmarks and updated iterations after new optimizations

craigtaverner:
The recent optimization of ESQL distance-sort allows us to benchmark that properly. That PR also did a few bug-fixes to distance filtering, so we added an alternative version of the query that is now also optimized.

### DIFF
--- a/geopoint/operations/default.json
+++ b/geopoint/operations/default.json
@@ -454,9 +454,55 @@
       }
     },
     {
+      "name": "distanceFilterCount200x300-alt",
+      "operation-type": "search",
+      "body": {
+        "query": {
+          "bool": {
+            "must": [
+              {
+              "geo_shape": {
+                "location": {
+                    "shape": { "type": "circle", "radius": "300km", "coordinates": [7, 55] },
+                    "relation": "intersects"
+                  }
+                }
+              },
+              {
+                "bool": {
+                  "must_not": [
+                    {
+                      "geo_shape": {
+                        "location": {
+                          "shape": { "type": "circle", "radius": "200km", "coordinates": [7, 55] },
+                          "relation": "intersects"
+                        }
+                      }
+                    }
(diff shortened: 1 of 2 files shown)


geopoint/operations/default.json
```diff
@@ -453,11 +453,57 @@
         }
       }
     },
+    {
+      "name": "distanceFilterCount200x300-alt",
+      "operation-type": "search",
+      "body": {
+        "query": {
+          "bool": {
+            "must": [
```
iverase, line 462: Use a filter instead of must?

geopoint/operations/default.json
```diff
@@ -453,11 +453,57 @@
...
+              "geo_shape": {
+                "location": {
+                    "shape": { "type": "circle", "radius": "300km", "coordinates": [7, 55] },
+                    "relation": "intersects"
+                  }
+                }
+              },
+              {
+                "bool": {
+                  "must_not": [
```
iverase, line 473: Instead of a must not, could we use a disjoint spatial relationship?

iverase (commented):
I think we can improve the _search, wdyt?

geopoint/operations/default.json
craigtaverner: The disjoint version was the original version. This is an alternative query that deliberately does things differently, to see if that makes a difference to the performance. It does not seem to make any difference.

craigtaverner:
> I think we can improve the _search, wdyt?

The original `distanceFilterCount200x300` has been there for a few months. This new `distanceFilterCount200x300-alt` is deliberately an alternative way of doing the same thing, in particular the `must_not` approach instead of a disjoint. The benchmark is simply testing whether there is any performance difference between the two approaches.

iverase (approved):

PR #701, opened 2024-10-29
Add new ESQL CCS challenge in Logs

gbanasiak:
The PR adds new `esql-ccs-snapshot` challenge in `elastic/logs` which reuses _existing_ ESQL queries but executes them in CCS context.

The challenge has 2 modes of operation:
* `skip_esql_ccs_remote_clusters: true` - queries are sent only to the local cluster
* `skip_esql_ccs_remote_clusters: false` - queries are sent only to remote clusters


elastic/logs/challenges/esql-ccs-snapshot.json
```diff
@@ -0,0 +1,167 @@
...
+          "operation-type": "cluster-health",
+          "index": "_all",
+          "request-params": {
+            "wait_for_status": "{{cluster_health | default('green')}}",
+            "wait_for_no_relocating_shards": "true"
+          },
+          "retry-until-success": true
+        }
+      },
+    {% else %}
```
favilo, line 48: Want to add a comment here, pointing to the corresponding `if`?

elastic/logs/challenges/esql-ccs-snapshot.json
```diff
@@ -0,0 +1,167 @@
...
+          "body": {
+            "persistent": {
+              "indices.recovery.max_bytes_per_sec": null,
+              "cluster.routing.allocation.node_concurrent_recoveries": null,
+              "cluster.routing.allocation.node_initial_primaries_recoveries": null
+            }
+          }
+        }
+      },
+    {% endif %}
```
favilo, line 164: Ditto here?

favilo (approved):
Left some style comments, but it looks good overall

gbanasiak:
@elasticmachine update branch

elastic/logs/challenges/esql-ccs-snapshot.json
gbanasiak: Good idea, https://github.com/elastic/rally-tracks/pull/701/commits/13f190c2caa59c6c6c6c08868c40bfb790f7f69e.

luigidellaquila (approved):
LGTM, thanks @gbanasiak!

I like the idea of making the tasks conditional and using them for both single cluster and CCS (rather than creating a separate set of tasks). This will ensure that we keep them aligned, so that we can compare them and evolve them together.

PR #702, opened 2024-10-29
Adding recall testing to openAI track

benwtrent:
This adds recall and greatly adjusts the OpenAI track. 

I thinking we should do something similar with ALL our vector tracks, even for the dense vector track as recalculating the true knn, even for dense vector is wasted compute time.

This is a draft as its a hack for some local testing, if we really want to move forward with this, I can clean this up and we can commit it.

### DIFF
--- a/openai_vector/challenges/default.json
+++ b/openai_vector/challenges/default.json
@@ -42,7 +42,7 @@
         "retry-until-success": true,
         "include-in-reporting": false
       }
-    },
+    }
     {# serverless-post-ingest-sleep-marker-start #}{%- if post_ingest_sleep|default(false) -%}
     {
       "name": "post-ingest-sleep",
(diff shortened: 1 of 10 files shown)


jimczi (approved):
LGTM

openai_vector/challenges/default.json
```diff
@@ -42,61 +42,56 @@
         "retry-until-success": true,
         "include-in-reporting": false
       }
-    },
+    }
     {# serverless-post-ingest-sleep-marker-start #}{%- if post_ingest_sleep|default(false) -%}
     {
```
gareth-ellis, line 47:
```suggestion
    ,{
```

We need a leading comma here, otherwise the json will be invalid if post_ingest_sleep is true

gareth-ellis (changes requested):
LGTM, with the exception that it adds 11 minutes to the system tests - this is because of the way the runner initialises, even in test mode. 

I don't have a good solution right now - we have another track that does the same, we got around it by skipping in the system tests - like here https://github.com/elastic/rally-tracks/pull/708

I dont think this will run for serverless, so just needs adding to the skip list for standard it.

One possibility, that I need to try, is to create a second file that contains many less queries, and use that based on a parameter being set. Then we could add a variant where we run e.g msmarco and openai setting this special parameter. I will add an issue so we don't forget about this tech debt...

benwtrent:
@gareth-ellis ah, I could add two "test mode" files? I have seen that done in other rally tracks.

benwtrent:
Let me know if I need to make any changes @gareth-ellis.

But, I do think once we get this finished, we can then make progress of moving our nightlies to this benchmark with different parameters.

gareth-ellis:
When you wrote earlier, I started looking at how we could do this - i've got it semi working in msmarco track, if i get it working 100% by later today,  i'll try and apply the same to yours, otherwise will temporarily add your new track to the skip list and then we should be good

gareth-ellis (approved):

PR #703, opened 2024-11-04
Two more benchmarks for partial sorting with ESQL

craigtaverner:
These cannot be replicated in _search since that only supports what can be pushed down to lucene, and this feature explicitly only pushes down part of the sort, and then does the other part in the compute engine.

### DIFF
--- a/geopoint/challenges/default.json
+++ b/geopoint/challenges/default.json
@@ -279,20 +279,14 @@
         },
         {
           "operation": "distanceFilterCount200x300",
-          "warmup-iterations": 200,
-          "iterations": 200,
+          "warmup-iterations": 50,
+          "iterations": 50,
           "tags": ["distance", "esql"]
         },
         {
-          "operation": "distanceFilterCount200x300-must_not",
-          "warmup-iterations": 200,
-          "iterations": 200,
-          "tags": ["distance", "esql"]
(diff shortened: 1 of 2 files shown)


geopoint/challenges/default.json
```diff
@@ -337,6 +337,12 @@
           "iterations": 50,
           "tags": ["distance", "esql", "sort"]
         },
+        {
+          "operation": "distanceSort-esql-partial",
```
iverase, line 341: For my curiosity, why do we have different warmups and iterations in these new benchmarks and the rationally behind this numbers?

iverase (approved):

geopoint/challenges/default.json
craigtaverner: I chose numbers that are related to the inverse of the latency. So a very long running benchmark (like the 10-20 seconds of the sort before the performance) should have very few iterations (we really don't care about precision on such long queries), while fast benchmarks (10-50ms) usually have high iteration counts, so we can get more stable, and predictable results. But in general I've also started moving away from really high numbers like 200/100 because they do not seem more stable than numbers like 50/50. I think variability is coming from the cloud infrastructure, so high iteration counts are costing time and money without bringing value.

PR #706, opened 2024-11-18
Add ESQL AVG query for non-runtime field

gbanasiak:


dnhatn (approved):
LGTM. Thanks @gbanasiak.

PR #719, opened 2024-12-31
ES|QL: add track for LOOKUP JOIN scale tests

luigidellaquila:
Adding a track to measure the performance of ES|QL LOOKUP JOIN at scale.

The PR contains
- a set of scripts to generate the corpus
- queries with the corresponding challenges
- track configuration to load the corpus and run the queries


joins/README.md
```diff
@@ -0,0 +1,56 @@
...
+  "field_0": "text with value 0_56",
+  "field_1": "text with value 1_56",
+  "field_2": "text with value 2_56",
+  ...
+  "field_99": "text with value 99_56"
+}
+```
+
+The cardinality of the keys is the same as the key name, eg. `key_1000` will have 1000 different values in the dataset,
+from `0` to `999`.
```
idegtiarenko, line 31: This assumes the file was generated with at least 1k documents.
This is likely true for small cardinality fields, but might not be the case for 100m.
I wonder if it is worth mentioning explicitly?

luigidellaquila: 👍 I added a small note

joins/operations/default.json
```diff
@@ -0,0 +1,243 @@
...
+    },
+    {
+      "name": "esql_lookup_join_1k_keys_where_limit1000",
+      "operation-type": "esql",
+      "query": "FROM join_base_idx | lookup join lookup_idx_1000_f10 on key_1000 | where lookup_keyword_0 like \"val 1*\" | limit 1000"
+    },
+    {
+      "name": "esql_lookup_join_1k_keys_where_no_match",
+      "operation-type": "esql",
+      "query": "FROM join_base_idx | lookup join lookup_idx_1000_f10 on key_1000 | where concat(lookup_keyword_0, \"foo\") == \"bar\""
```
luigidellaquila, line 67: For now these queries work the same as the `*keys_where_limit*` (see above), but in the future we could be able to push down to lucene one of the two, but most likely not the other one.

idegtiarenko (approved):

joins/files.txt
```diff
@@ -0,0 +1,2 @@
+documents.json.bz2
+documents-1k.json.bz2
```
gbanasiak, line 2: What is the purpose of this file?

joins/challenges/default.json
```diff
@@ -0,0 +1,336 @@
...
+            "operation-type": "cluster-health",
+            "index": "join_base_idx",
+            "request-params": {
+              "wait_for_status": "{{cluster_health | default('green')}}",
+              "wait_for_no_relocating_shards": "true"
+            },
+            "retry-until-success": true
+          },
+          "tags": ["setup"]
+        #}
```
gbanasiak, line 38: Let's uncomment this, it's a good practice to not start a benchmark in an unhealthy cluster.

joins/track.json
```diff
@@ -0,0 +1,194 @@
...
+        {
+          "target-index": "join_base_idx",
+          "source-file": "join_base_idx.json.bz2",
+          "#COMMENT": "1k docs version of the main index",
+          "document-count": 1000,
+          "compressed-bytes": 37788,
+          "uncompressed-bytes": 9346120
+        }
+      ]
+    #}
```
gbanasiak, line 186: Nit: cleanup unused copies?

joins/index-join_base_idx.json
```diff
@@ -0,0 +1,43 @@
+{% set p_include_non_serverless_index_settings = (include_non_serverless_index_settings | default(build_flavor != "serverless")) %}
+
+{
+  "settings": {
+    {# non-serverless-index-settings-marker-start #}{%- if build_flavor != "serverless" or serverless_operator == true -%}
+    "index.requests.cache.enable": false
+    {%- endif -%}{# non-serverless-index-settings-marker-end #}
+  },
```
gbanasiak, line 8: Let's default to 0 replicas with:
```
"index.number_of_replicas": {{number_of_replicas | default(0)}},
```

Like in: https://github.com/elastic/rally-tracks/blob/d5697bd8d44c4dd099851993d59505c342e847be/geopoint/index.json#L5-L11

Unless the track is _not_ supposed to be used in serverless in case of which this is unnecessary boilerplate.

Please update remaining index files accordingly.

joins/track.json
```diff
@@ -0,0 +1,194 @@
+{% import "rally.helpers" as rally with context %}
+
+{% set p_include_non_serverless_index_settings = (include_non_serverless_index_settings | default(build_flavor != "serverless")) %}
+{% set p_include_force_merge = (include_force_merge | default(build_flavor != "serverless")) %}
+{% set p_include_target_throughput = (include_target_throughput | default(build_flavor != "serverless")) %}
```
gbanasiak, line 5: Not needed.

joins/README.md
```diff
@@ -0,0 +1,56 @@
...
+* `index_settings`: A list of index settings. Index settings defined elsewhere need to be overridden explicitly.
+* `cluster_health` (default: "green"): The minimum required cluster health.
+* `error_level` (default: "non-fatal"): Available for bulk operations only to specify ignore-response-error-level.
+* `detailed_results` (default: `false`): Adds additional [metadata](https://esrally.readthedocs.io/en/latest/track.html?highlight=detailed-results#meta-data) to challenges using the track `update` operation. Be aware using this option can add client side overhead due to the deserialization of API responses.
+* `include_non_serverless_index_settings` (default: true for non-serverless clusters, false for serverless clusters): Whether to include non-serverless index settings.
+* `include_force_merge` (default: true for non-serverless clusters, false for serverless clusters): Whether to include force merge operation.
+* `include_target_throughput` (default: true for non-serverless clusters, false for serverless clusters): Whether to apply target throughput.
```
gbanasiak, line 51: Can we update the list of parameters to reflect what is actually in the track?

* `bulk_indexing_clients` - not supported now, perhaps worth adding
* `number_of_shards` - same
* `force_merge_max_num_segments` - unused
* `error_level` - unused
* `detailed_results` - unused
* `include_force_merge` and `include_target_throughput` - unused

joins/files.txt
luigidellaquila: It's a leftover from a copy-paste. I'm not sure if Rally uses it, I'll try to just delete it.

joins/index-join_base_idx.json
luigidellaquila: Doing it for the base index, but I cannot do it for the "lookup" indexes. They will have N replicas as the number of nodes, that's specific to the type of index. I hope it's not a problem for the CI

gbanasiak (approved):
Many thanks for iterating. Forgot to mention earlier: if the track is serverless-ready please add `joins` in [here](https://github.com/elastic/rally-tracks/blob/2ff3235b4c5b447cf2409865230a2ebbeec71309/it_serverless/test_selected_tracks_and_challenges.py#L26). Can be done in a separate PR.

luigidellaquila:
Thanks @gbanasiak!
We are still fine-tuning the Serverless tests, I'll add it with a follow-up PR

joins/operations/default.json
```diff
@@ -0,0 +1,243 @@
...
+      "operation-type": "esql",
+      "query": "FROM join_base_idx | lookup join lookup_idx_100000000_f10 on key_100000000 | where concat(lookup_keyword_0, \"foo\") == \"bar\""
+    },
+
+
+
+    {
+    "name": "esql_lookup_join_1k_100k_200k_500k",
+    "operation-type": "esql",
+    "query": "FROM join_base_idx | lookup join lookup_idx_1000_f10 on key_1000 | rename lookup_keyword_0 as lk_1k | lookup join lookup_idx_100000_f10 on key_100000 | rename lookup_keyword_0 as lk_100k | lookup join lookup_idx_200000_f10 on key_200000 | rename lookup_keyword_0 as lk_200k | lookup join lookup_idx_500000_f10 on key_500000 | rename lookup_keyword_0 as lk_500k | keep id, key_1000, key_100000, key_200000, key_500000, lk_1k, lk_100k, lk_200k, lk_500k | limit 1000"
```
alex-spies, line 237: The limit 1000 at the end is pushed down, and therefore all lookups will happen on the coordinator node, I think. This means we also perform the lookups against max 1000 rows.

The same is true essentially for all other queries in this file because we add an implicit `LIMIT 1000` per default.

I think it'd be good to have versions of these queries where we add a `SORT ...` at the end - this should force the lookups onto the data nodes (best to confirm, though, by looking at the plans).

joins/_tools/README.md
```diff
@@ -0,0 +1,89 @@
...
+The dataset for this benchmark was generated with the following:
+
+```shell
+./lookup_idx.sh 1000 10 1 | shuf | bzip2 -c > lookup_idx_1000_f10.json.bz2
+./lookup_idx.sh 100000 10 1 | shuf | bzip2 -c > lookup_idx_100000_f10.json.bz2
+./lookup_idx.sh 200000 10 1 | shuf | bzip2 -c > lookup_idx_200000_f10.json.bz2
+./lookup_idx.sh 500000 10 1 | shuf | bzip2 -c > lookup_idx_500000_f10.json.bz2
+./lookup_idx.sh 1000000 10 1 | shuf | bzip2 -c > lookup_idx_1000000_f10.json.bz2
+./lookup_idx.sh 5000000 10 1 | shuf | bzip2 -c > lookup_idx_5000000_f10.json.bz2
+./joins_main_idx.sh 10000000 | bzip2 -c > join_base_idx-10M.json.bz2
```
alex-spies, line 88: nit: maybe a lower number of lookup indices would be sufficient? Like, 1k, 50k, 1M, 10M?

I think it's more interesting to have different numbers of repetitions IMHO.

joins/challenges/default.json
```diff
@@ -0,0 +1,341 @@
...
+          "warmup-iterations": 10,
+          "iterations": 50
+        },
+        {#
+          "operation": "esql_lookup_join_100k_keys_where_no_match",
+          "tags": ["lookup", "join"],
+          "clients": 1,
+          "warmup-iterations": 10,
+          "iterations": 50
+        #}
```
alex-spies, line 143: Are the `..._where_no_match` challenges commented out, resp. is this on purpose?

alex-spies (commented):
Thanks a lot Luigi! Looks good, although I do have suggestions for further iterations on the track.

I think in a follow up, we should deliberately force the lookups onto the coordinator or the data nodes, resp. See my comment below.

Additionally, I'm curious about how the following queries would perform:
- lookup join against a lookup index where the non-lookup fields contain large values (long strings or many multi-values, for instance) - but maybe that's less relevant as the default dataset has 10 additional text fields already.
- What happens when the lookup indices have many rows matching the same keys? Maybe the default dataset should have some repetitions?

Especially the last one will be important IMHO as with the planned SQL-like semantics, the cardinality of the whole result set will be changing with every LOOKUP JOIN - but it's already interesting in the current state, as the current implementation may pool a lot of multivalues.

joins/operations/default.json
luigidellaquila: Good point, I think it makes sense to have it as next iteration. 
I'll add some queries right away

joins/challenges/default.json
luigidellaquila: Yes, it's on purpose. These are super expensive and just time out.
My plan is to run some iterations with a higher timeout (tens of minutes)  and test the limits here

joins/operations/default.json
alex-spies: Actually, I realized I was being imprecise. The queries with a `WHERE` clause will not have the limit pushed down past the LOOKUP JOINs, because the WHERE clause should prevent the pushdown. So there's already queries that force the execution of LOOKUP JOINs onto the data nodes.

PR #722, opened 2025-01-10
Introduce a new operation to benchmark the shard changes recovery api

salvatore-campagna:
With [PR elastic/elasticsearch#118608](https://github.com/elastic/elasticsearch/pull/118608), we introduced a new API that we would like to benchmark in order to evaluate the latency of synthetic source recovery. This benchmarking will help us assess how the API latency evolves over time and understand the potential impact of new features on recovery latency. The targets for this benchmark will include indices or data streams in the `elastic/logs`, `elastic/security`, and `tsdb` tracks.

- Retention Lease Requirement:  
  It is important to add a retention lease before starting document indexing. This is necessary for ensuring that the subsequent shard recovery API call executes successfully without missing `seq_no` values. A missing retention lease may result in missing `seq_no` values during shard recovery.

- Support for Data Streams and Index Aliases:  
  We extended the original API in [PR elastic/elasticsearch#118937](https://github.com/elastic/elasticsearch/pull/118937) to support data streams and index aliases. This extension is necessary for data stream use cases, as the backing index name may change depending on when the index is created. Supporting data streams and aliases simplifies benchmarking and provides greater flexibility in targeting indices.


tsdb/challenges/synthetic-source-recovery.json
```diff
@@ -0,0 +1,128 @@
...
+    {# serverless-post-ingest-sleep-marker-start #}{%- if post_ingest_sleep|default(false) -%}
+    {
+      "name": "post-ingest-sleep",
+      "operation": {
+        "operation-type": "sleep",
+        "duration": {{ post_ingest_sleep_duration|default(30) }}
+      }
+    },
+    {%- endif -%}{# serverless-post-ingest-sleep-marker-end #}
+    {
```
martijnvg, line 111: With this approach, I think we would need a new config in nighly? Maybe we can move `shard-recovery` to `logging-querying.json` so that the operation can run with already configured jobs in nighly? The task would then be include if `recovery_target` is set.

martijnvg (commented):
Thanks Salvatore! I left a comment about inlining shard-recovery in existing challenge.

gareth-ellis:
@elasticmachine update branch

elastic/logs/track.json
```diff
@@ -60,6 +60,12 @@
 {% set p_dsl_default_rollover = (dsl_default_rollover | default(false) ) %}
 {% set p_skip_fleet_globals = (skip_fleet_globals | default(false) ) %}
 
+{% set p_recovery_target = (recovery_target | default('logs-k8-application.log-default') ) %}
```
martijnvg, line 63: maybe not define default here? Then in `logging-query.json` we can only include `shard-recovery` operation if `recovery_target` track param is defined.

martijnvg (approved):

elastic/logs/track.json
martijnvg: I don't think we need to run this on all variants of the track we run in nighly. For example the routing variant doesn't need this.

salvatore-campagna: If I don't have a default value the CI would run with an empty value and fail because the index does not exist...this is the reason why I define a default.

salvatore-campagna: I am adding a `include_shard_recovery_op` which is `false` by default and that we can eanble only for selected configurations.

elastic/logs/challenges/synthetic-source-recovery.json
```diff
@@ -0,0 +1,99 @@
...
+      },
+      "clients": {{ p_bulk_indexing_clients }},
+      "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
+    },
+    {
+      "name": "refresh-after-index",
+      "index": "logs-*",
+      "operation": "refresh"
+    },
+    {
```
martijnvg, line 40: I don't think we need to these operations related to merging? It just makes this challenge run longer?

tsdb/challenges/synthetic-source-recovery.json
```diff
@@ -0,0 +1,95 @@
...
+      "operation": "index",
+      "warmup-time-period": 240,
+      "clients": {{bulk_indexing_clients | default(8)}},
+      "ignore-response-error-level": "{{error_level | default('non-fatal')}}"
+    },
+    {
+      "name": "refresh-after-index",
+      "operation": "refresh"
+    },
+    {
```
martijnvg, line 50: I think we can remove force-merge and wait-until-merges-finish here?

tsdb/challenges/synthetic-source-recovery.json
salvatore-campagna: As we discussed adding a new challenge is necessary. What we do here would affect other metrics like  storage. As a result we can't really introduce a step in an existing challenge.

martijnvg (approved):
LGTM2

PR #726, opened 2025-01-15
Temporarily disable joins in IT

gareth-ellis:
https://github.com/elastic/elasticsearch/pull/120073 has meant that lookup indices in the joins track are left in yellow. Discussion is ongoing, but to unblock CI i suggest temporarily disabling joins for IT tests

### DIFF
--- a/it/test_all_tracks_and_challenges.py
+++ b/it/test_all_tracks_and_challenges.py
@@ -21,7 +21,7 @@ pytest_rally = pytest.importorskip("pytest_rally")
 
 
 class TestTrackRepository:
-    skip_tracks = ["elastic/logs", "elastic/security", "k8s_metrics", "sql", "elser-ingest-speedtest", "msmarco-v2-vector"]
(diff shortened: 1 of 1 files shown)


kkrik-es (approved):
@dnhatn fyi

martijnvg (approved):
LGTM

gareth-ellis:
No, they started after the Elasticsearch PR stated in the description, that changes the default for lookup indices to have 1 replica, whilst before the default was 0

gbanasiak:
Can we add `auto_expand_replicas` to all join indices instead? It's easy to skip, way harder to remember to reintroduce later.

gareth-ellis:
You mean set it to false ? We can do, if the ESQL team are happy with that? @martijnvg is going to discuss later if there are any other changes required 

I'll test first it works and then update the PR

gbanasiak:
> You mean set it to false ?

I would try with:
```
"auto_expand_replicas": "0-all"
```

gareth-ellis:
created https://github.com/elastic/rally-tracks/pull/727

gareth-ellis:
Will close this for now, #727 seems to have fixed it

PR #727, opened 2025-01-15
Add auto_expand_replicas to joins track

gareth-ellis:
The behaviour of auto_expand_replicas has changed which means that even in IT tests, we default to having a single replica. This unblocks the IT tests and also adds a new option to control the auto_expand_replicas behaviour for lookup indices


gbanasiak (approved):

fressi-elastic (approved):
LGTM

luigidellaquila (approved):

PR #728, opened 2025-01-17
ES|QL JOIN: query templates to reduce redundancy

luigidellaquila:
Refactoring ES|QL JOIN track to make it less verbose and repetitive.

### DIFF
--- a/joins/challenges/default.json
+++ b/joins/challenges/default.json
@@ -112,15 +112,15 @@
           "warmup-iterations": 10,
           "iterations": 50
         },
+{% endfor %}
+
         {
-          "operation": "esql_lookup_join_{{idx_suffix[i]}}_keys_where_no_match",
+          "operation": "esql_lookup_join_1k_keys_where_no_match",
           "tags": ["lookup", "join"],
           "clients": 1,
           "warmup-iterations": 5,
           "iterations": 20
         },
-{% endfor %}
-
 
         {
           "operation": "esql_lookup_join_1k_100k_200k_500k",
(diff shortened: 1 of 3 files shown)


joins/challenges/default.json
```diff
@@ -69,361 +69,57 @@
...
-          "operation": "esql_lookup_join_1k_keys_where_limit1000",
+          "operation": "esql_lookup_join_{{idx_suffix[i]}}_keys_where_limit1000",
           "tags": ["lookup", "join", "limit1000"],
           "clients": 1,
           "warmup-iterations": 10,
           "iterations": 50
         },
         {
-          "operation": "esql_lookup_join_1k_keys_where_no_match",
+          "operation": "esql_lookup_join_{{idx_suffix[i]}}_keys_where_no_match",
```
alex-spies, line 116: This enables the no match queries in this challenge to run against all indices, while currently this is commented out for all indices except the 1k one. I think this is not intended, as we know this task is particularly long running, right?

alex-spies (commented):
I think we change the default benchmark w.r.t. the no match tasks - let's double check. Otherwise this LGTM, this is lovely!

joins/challenges/default.json
luigidellaquila: Right. I'd like to be able to run them, but probably it's better to do it as a separate PR, and in a brand new "large" challenge.

luigidellaquila: Done

alex-spies (approved):
Thanks Luigi!

PR #731, opened 2025-01-24
Two shard recovery requests with different batch size

salvatore-campagna:
Here we update both the `elastic/logs` and `tsdb` tracks to include two shard recovery
requests, one with batch size of `4MB` and the other with batch size of `32MB`.
The value for the translog operations count is kept high to make sure we always hit
the limit on the batch size first.


martijnvg (approved):
LGTM, looks like default chunk size for shard recovery is `512kb`.

PR #732, opened 2025-01-28
Set source mode through an index setting

kkrik-es:
`mappings._source.mode` is getting deprecated and replaced by setting `index.mapping.source.mode`.


martijnvg (approved):
LGTM

PR #734, opened 2025-01-29
Add auto follow pattern challenge

martijnvg:
Add a new challenge that uses auto follow pattern to auto follow data streams with `logs-*` pattern into the local cluster. This allows to benchmark CCR in a semi realistic way while the data set gets indexed, without running multiple clusters.


elastic/logs/challenges/synthetic-source-recovery.json
```diff
@@ -46,8 +46,56 @@
           "max_operations_count": {{ p_recovery_max_operations_count | tojson }}
         }
       },
-      "warmup-iterations": 10,
-      "iterations": 20
+      "warmup-iterations": 0,
+      "iterations": 1
+    },
+    {
+      "name": "shard-recovery-small_2",
```
salvatore-campagna, line 53: nit: here you use "_2" (underscore) while below you use "-2" (dash)

elastic/logs/track.json
```diff
@@ -62,10 +62,13 @@
 
 {% set p_recovery_target = (recovery_target | default('logs-k8-application.log-default') ) %}
 {% set p_recovery_from_seq_no = (recovery_seq_no | default(0) ) %}
+{% set p_recovery_from_seq_no_2 = (recovery_seq_no_2 | default(1000000) ) %}
+{% set p_recovery_from_seq_no_3 = (recovery_seq_no_3 | default(2000000) ) %}
+{% set p_recovery_from_seq_no_4 = (recovery_seq_no_4 | default(3000000) ) %}
```
salvatore-campagna, line 67: Probably this is way to high. AFAIK there are only 1_230_000 docs in k8s logs.

salvatore-campagna:
LGTM apart a nit.

salvatore-campagna (approved):

elastic/logs/challenges/synthetic-source-recovery.json
```diff
@@ -46,9 +46,299 @@
...
+          "max_batch_size": {{ p_recovery_small_max_batch_size | tojson }},
+          "poll_timeout": {{ p_recovery_poll_timeout | tojson }},
+          "max_operations_count": {{ p_recovery_max_operations_count | tojson }}
+        }
+      },
+      "warmup-iterations": 0,
+      "iterations": 1
+    },
+    {
+      "name": "shard-recovery-large_3",
```
salvatore-campagna, line 69: I think this should be `small_3`

elastic/logs/track.json
```diff
@@ -61,11 +61,51 @@
...
+{% set p_recovery_small_from_seq_no_12 = (recovery_seq_no_12 | default(11000) ) %}
+{% set p_recovery_small_from_seq_no_13 = (recovery_seq_no_13 | default(12000) ) %}
+{% set p_recovery_small_from_seq_no_14 = (recovery_seq_no_14 | default(13000) ) %}
+{% set p_recovery_small_from_seq_no_15 = (recovery_seq_no_15 | default(14000) ) %}
+{% set p_recovery_small_from_seq_no_16 = (recovery_seq_no_16 | default(15000) ) %}
+{% set p_recovery_small_from_seq_no_17 = (recovery_seq_no_17 | default(16000) ) %}
+{% set p_recovery_small_from_seq_no_18 = (recovery_seq_no_18 | default(17000) ) %}
+{% set p_recovery_small_from_seq_no_19 = (recovery_seq_no_19 | default(18000) ) %}
+
+{% set p_recovery_large_from_seq_no = (recovery_seq_no | default(0) ) %}
```
salvatore-campagna, line 85: I think the issue is here...you use "-" instead of "_".

salvatore-campagna:
LGTM 2.

elastic/logs/challenges/auto-follow-pattern.json
```diff
@@ -0,0 +1,47 @@
+{% import "rally.helpers" as rally %}
+{% set p_follow_index_pattern = "copy-{{leader_index}}" %}
+{
+  "name": "auto-follow-pattern",
```
salvatore-campagna, line 4: Maybe we can use a better name for the challenge...something that reminds what this does? Something like "ccr-shard-recovery". I guess this track is not only useful for synthetic source... but probably it would provide useful info in Serverless too where we probably retrieve data from S3?

salvatore-campagna:
LGTM

elastic/logs/challenges/auto-follow-pattern.json
martijnvg: > Maybe we can use a better name for the challenge...something that reminds what this does?

+1 - I will add ccr replication to the name and add a description.

> . I guess this track is not only useful for synthetic source... but probably it would provide useful info in Serverless too where we probably retrieve data from S3?

IIRC CCR currently only works in stateful. To keep the follower indices up to date it relies on operation based replication, which doesn't exist in serverless.

PR #737, opened 2025-01-30
Support retries in MultiClusterWrapper

gbanasiak:
The `multi-cluster-wrapper` custom runner unwraps the runner before calling it for every remote cluster. This strips the wrapper responsible for retries ([src](https://github.com/elastic/rally/blob/c5b469d2729dd9065c5d86122399dd85dfcd9bf8/esrally/driver/runner.py#L2953)) from the runner making operation parameters such as `retry-until-success` ignored.

Another wrapper removed during the unwrap is `MultiClientRunner` ([src](https://github.com/elastic/rally/blob/c5b469d2729dd9065c5d86122399dd85dfcd9bf8/esrally/driver/runner.py#L338), [src](https://github.com/elastic/rally/blob/c5b469d2729dd9065c5d86122399dd85dfcd9bf8/esrally/driver/runner.py#L263-L270)). This wrapper assumes that ES client is specified through a `{"<cluster-name>": <client>}` dictionary, and for single-cluster runners (all built-in Rally runners) this wrapper extracts the client from `default` cluster, while for multi-cluster runners (custom ones), provides entire dictionary.

This PR tries to address it by resigning from the unwrap. For this to work ES client is passed in a dictionary with `default` key which assumes the underlying runner will be a single-cluster one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ PYENV_REGEX = .pyenv/shims
 PY_BIN = python3
 # https://github.com/pypa/pip/issues/5599
 PIP_WRAPPER = $(PY_BIN) -m pip
-export PY38 = "3.8.13"
+export PY312 = "3.12.2"
 VIRTUAL_ENV ?= .venv
 VENV_ACTIVATE_FILE = $(VIRTUAL_ENV)/bin/activate
 VENV_ACTIVATE = . $(VENV_ACTIVATE_FILE)
@@ -31,8 +31,8 @@ PYENV_PREREQ_HELP = "\033[0;31mIMPORTANT\033[0m: please type \033[0;31mpyenv ini
 VE_MISSING_HELP = "\033[0;31mIMPORTANT\033[0m: Couldn't find $(PWD)/$(VIRTUAL_ENV); have you executed make venv-create?\033[0m\n"
 
 prereq:
-	pyenv install --skip-existing $(PY38)
-	pyenv local $(PY38)
+	pyenv install --skip-existing $(PY312)
+	pyenv local $(PY312)
 
 venv-create:
 	@if [[ ! -x $$(command -v pyenv) ]]; then \
--- a/elastic/shared/runners/remote_cluster.py
+++ b/elastic/shared/runners/remote_cluster.py
@@ -215,10 +215,13 @@ class MultiClusterWrapper(Runner):
             if cluster_name in params.get("ignore-clusters", []):
                 self.logger.info(f"Multi cluster wrapped runner [{base_runner}] ignoring cluster [{cluster_name}].")
                 continue
-            runner_for_op = unwrap(runner_for(base_runner))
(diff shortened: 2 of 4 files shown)


gbanasiak:
I couldn't make unit tests work locally for me without Python and package upgrades.

Makefile
```diff
@@ -31,8 +31,8 @@ PYENV_PREREQ_HELP = "\033[0;31mIMPORTANT\033[0m: please type \033[0;31mpyenv ini
 VE_MISSING_HELP = "\033[0;31mIMPORTANT\033[0m: Couldn't find $(PWD)/$(VIRTUAL_ENV); have you executed make venv-create?\033[0m\n"
 
 prereq:
-	pyenv install --skip-existing $(PY38)
-	pyenv local $(PY38)
+	pyenv install --skip-existing $(PY312)
```
gareth-ellis, line 34: How come we jump to 3.12 instead of 3.9 ?

elastic/shared/runners/remote_cluster.py
```diff
@@ -215,10 +215,13 @@ async def __call__(self, multi_es, params):
             if cluster_name in params.get("ignore-clusters", []):
                 self.logger.info(f"Multi cluster wrapped runner [{base_runner}] ignoring cluster [{cluster_name}].")
                 continue
-            runner_for_op = unwrap(runner_for(base_runner))
+            runner_for_op = runner_for(base_runner)
             self.logger.info(f"Multi cluster wrapped runner [{base_runner}] executing on cluster [{cluster_name}].")
-            # just call base runner op, don't mess with 'return' values
-            coroutines.append(runner_for_op(cluster_client, params))
+            # call original runner assuming a single-cluster one, see https://github.com/elastic/rally/pull/488
+            # and https://github.com/elastic/rally/pull/1563 for additional context
```
gbanasiak, line 221: Actually we cannot assume that, this requires a modification.

Makefile
gbanasiak: This follows dev environment preparation from Rally's Makefile, where 3.12 is used as well. It aligns both.

gareth-ellis: Ah, I hadnt spotted we had switched there (We used to use 3.8, then when we upgraded to 3.12 we jumped straight to 3.12). I suppose from a testing perspective we still test on 3.9, so thats ok I guess

elastic/shared/runners/remote_cluster.py
gbanasiak: I've generalized in https://github.com/elastic/rally-tracks/pull/737/commits/e3f42eab452cb9338e87d204b619433e1c7fa6d8. Today `multi-cluster-wrapper` is only used with built-in Rally runners which are exclusively single-cluster. But I can imagine a situation where someone wants to iterate through all the clusters providing a dictionary of clusters for the runner to choose from. Perhaps that's too much.

PR #741, opened 2025-02-10
Restore the source mode setting in the remaining tests

kkrik-es:
These were not updated in #732 as the setting was not available in serverless.


martijnvg (approved):
Good catch! LGTM

PR #743, opened 2025-02-13
Make the `tsdb` track compatible with Elasticsearch 7.17

salvatore-campagna:
Here we make sure we exclude things from the configuration which are incompatible with
Elasticsearch v7. This is required to be able to use the challenge and the dataset in an
upgrade test which indexes metrics data into Elasticsearch 7.17.17 in standard mode. In that version
some things like `source.mode`, synthetic source, time series aggregations and so on were not available.
Also we need to disable the downsampling and low latency challenges because:
* downsampling requires a time series index as a source index, which is not available in Elasticsearch v7
* the low latency challenge uses document IDs to fetch documents which are likely different in Elasticsearch v7

### DIFF
--- a/tsdb/README.md
+++ b/tsdb/README.md
@@ -195,6 +195,7 @@ This track allows to overwrite the following parameters using `--track-params`:
 * `number_of_shards` (default: 1)
 * `refresh_interval` (default not defined)
 * `force_merge_max_num_segments` (default: unset): An integer specifying the max amount of segments the force-merge operation should use.
+* `include_source_mode` (default: true): used to explicitly exclude `index.mapping.source.mode` setting when running with Elasticsearch v7.
 * `source_mode` (default: synthetic): Should the `_source` be `stored` to disk exactly as sent (the Elasticsearch default outside of TSDB mode), thrown away (`disabled`), or reconstructed on the fly (`synthetic`)
 * `index_mode` (default: time_series): Whether to make a standard index (`standard`) or time series index (`time_series`)
 * `codec` (default: default): The codec to use compressing the index. `default` uses more space and less cpu. `best_compression` uses less space and more cpu.
--- a/tsdb/challenges/default.json
+++ b/tsdb/challenges/default.json
@@ -126,7 +126,7 @@
           "clients": {{ search_clients | default(1) | int }},
           "iterations": 100
         }
-        {% if index_mode | default('time_series') is equalto 'time_series' and skip_running_tsdb_aggs is not defined %}
+        {% if p_index_mode == "time_series" and skip_running_tsdb_aggs is not defined %}
         ,{
           "operation": "date-histo-memory-usage-hour",
           "warmup-iterations": 50,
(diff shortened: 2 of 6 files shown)


dnhatn:
@gareth-ellis I was able to rally this track with ES 7.17 with these params:

```
{
  "force_merge_max_num_segments": 1,
  "bulk_indexing_clients": 16,
  "corpus": "split16",
  "include_esql_queries": false,
  "include_source_mode": false,
  "skip_running_tsdb_aggs": true,
  "enable_downsample_challenge": false,
  "enable_low_latency_challenge": false
}
```

Can you take another look? Thank you!

gareth-ellis:
This is being merged into main, and seems that tsdb downsample challenge  now is incompatible with main - 
```	
Cannot run task [downsample-1m]: Request returned an error. Error type: api, Description: exception ({'error': {'root_cause': [{'type': 'exception', 'reason': 'Rollup requires setting [index.mode=time_series] for index [tsdb]'}], 'type': 'exception', 'reason': 'Rollup requires setting [index.mode=time_series] for index [tsdb]'}, 'status': 500}), HTTP Status: 500
```

salvatore-campagna:
> This is being merged into main, and seems that tsdb downsample challenge now is incompatible with main -
> 
> ```
> Cannot run task [downsample-1m]: Request returned an error. Error type: api, Description: exception ({'error': {'root_cause': [{'type': 'exception', 'reason': 'Rollup requires setting [index.mode=time_series] for index [tsdb]'}], 'type': 'exception', 'reason': 'Rollup requires setting [index.mode=time_series] for index [tsdb]'}, 'status': 500}), HTTP Status: 500
> ```

Yes:
1. downsampling was introduced in v8 and never backported
2. downsampling requires the source index to be a TSDB index. TSDB is an index mode introduced in v8 and never backported (as it requires TSDB).

So that challenge is incompatible with Elasticsearch v7 and the CI tries to run all of them, that is why I added `enable_downsample_challenge`.

The other challenge `low_latency` requires using specific document IDs...which are likely different in v7...that is the reason why I introduced `enable_low_latency_challenge`.

Anyway in the CI both challenges are checked because the default value for both `enable_downsample_challenge` and `enable_downsample_challenge` is true.

martijnvg (approved):
LGTM 👍 

> should we merge this?

Yes.

gareth-ellis (approved):
LGTM

salvatore-campagna:
I will wait on Monday to merge this, just to avoid nightly benchmark issues affecting the TSDB track during the weekend.

salvatore-campagna:
This is a sample `esbench` configuration which runs the TSDB challenge with Elasticsearch `7.17.27`.

```
{
    "elasticsearch.version": "7.17.27",
    "track.name": "tsdb",
    "track.challenge": "append-no-conflicts",
    "on_error": "abort",
    "track.params": {
      "force_merge_max_num_segments": 1,
      "bulk_indexing_clients": 16,
      "corpus": "split16",
      "include_source_mode": false,
      "skip_running_tsdb_aggs": true,
      "enable_downsample_challenge": false,
      "enable_low_latency_challenge": false
    },
    "client.options": {
        "default": {
            "timeout": 3600
        }
    },
    "elasticsearch.preserve_install": true
  }
  ```

PR #744, opened 2025-02-13
Import and register all required components for elastic/security

gareth-ellis:
Whilst running some testing on 7.17, i discovered elastic/security wasnt properly backported, so was missing the registration of the components for data generation.

This imports them so the track works under 7.17. The changes seem to be in master, so just backporting to 7.17

### DIFF
--- a/elastic/security/track.json
+++ b/elastic/security/track.json
@@ -13,7 +13,34 @@
   "packetbeat",
   "winlogbeat"
 ]))%}
-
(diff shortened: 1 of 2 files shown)


gareth-ellis:
Not really sure what to do here.....adding serverless to the 7.17 branch seems a little unecessary, since none of the tracks have the the changes to make parts of the track conditional on version

PR #745, opened 2025-02-17
Remove synthetic source recovery challenge

salvatore-campagna:
This challenge is not required anymore after moving to benchmarking synthetic source recovery
using (local) CCR and the auto-follow pattern.


gareth-ellis (approved):

martijnvg (approved):
LGTM

PR #748, opened 2025-02-21
Work around ESQL query parsing bug

gbanasiak:
After https://github.com/elastic/elasticsearch/pull/121948 we're seeing the following error:

```
{
  "error" : {
    "root_cause" : [
      {
        "type" : "parsing_exception",
        "reason" : "line 1:327: no viable alternative at input 'avg(kubernetes.pod.cpu.usage.limit'"
      }
    ],
    "type" : "parsing_exception",
    "reason" : "line 1:327: no viable alternative at input 'avg(kubernetes.pod.cpu.usage.limit'",
    "caused_by" : {
      "type" : "no_viable_alt_exception",
      "reason" : null
    }
  },
  "status" : 400
}
```

This PR is a workaround until https://github.com/elastic/elasticsearch/pull/123085 is in.


gbanasiak:
Empirically `STATS percentile(kubernetes.container.cpu.usage.nanocores, 95)` modification was not required but I've changed it for completeness. I don't know what are the exact conditions that trigger, is just the `limit` keyword?

ChrisHegarty:
> Empirically `STATS percentile(kubernetes.container.cpu.usage.nanocores, 95)` modification was not required but I've changed it for completeness. I don't know what are the exact conditions that trigger, is just the `limit` keyword?

The issue is specifically with `limit` and `sort` (only).

ChrisHegarty (approved):
LGTM. And of course this can be reverted once the issue is fixed in Elasticsearch.

gareth-ellis (approved):

gbanasiak:
> The issue is specifically with `limit` and `sort` (only).

Thanks for clarifying. I went too far initially, will go for the minimum change required.

PR #750, opened 2025-03-05
Add ES|QL specific tracks

luigidellaquila:
Add ES|QL specific track.
For now it includes tests for basic use cases with very broad index patterns with many indices.
It also includes options to test parameterize number of indices and shards, number of field per index, number of query clients (it's specifically designed for scale tests)


bpintea (approved):
Nice

PR #753, opened 2025-03-11
elastic/logs: use the common custom settings component template for K8s application logs

inqueue:
Allow common index settings configurable in the `track-custom-shard-settings` component template to be utilized by the K8s application logs composable template. Also includes some minor jinja clean up.


martijnvg (approved):
LGTM

PR #755, opened 2025-03-12
Add challenge to the elastic/log track which upgrades a data stream

parkertimmins:
Add a new challenge which test the performance of upgrading a data stream using the reindex data stream endpoints.

This challenge does the following:
* restores a data stream from a snapshot
* starts a data stream upgrade with a call to `/_migration/reindex`
* waits for the upgrade to complete, polling `/_migration/reindex/<data-stream>/_status` once a second

The following upgrade specific parameters can be set. These describe the settings described [here](https://www.elastic.co/guide/en/elasticsearch/reference/8.18/data-stream-reindex-api.html#reindex-data-stream-api-settings).
* reindex_max_concurrent_indices
* reindex_max_requests_per_second

 The only data stream indices which (currently) require upgrades are 7.x indices in version 8 ES, in preparation for an upgrade to 9.x.  To obtain a data stream with 7.x indices we load the data stream from a snapshot. This challenge reuses the existing snapshot parameters used by the `logging-snapshot-restore` challenge. 
The following snapshot parameters should be set:
```   
  "restore_data_streams"
  "snapshot_repo_name"
  "snapshot_repo_type"
  "snapshot_repo_settings"
  "snapshot_name"
  "snapshot_rename_suffix"
```


elastic/logs/README.md
```diff
@@ -405,6 +405,14 @@ Note: `include_target_throughput` parameter is ignored in this challenge.
 Runs the categorize-text aggregation with varying values of shard-size and with/without the use of a sampler
 aggregation. The challenge targets a specific set of indices by way of an index alias.
 
+### Reindex Data Stream (logging-reindex-data-stream)
+
+Restores a data stream from a 7.x snapshots and reindexes all indices into version 8.x.
```
inqueue, line 410: nit minor
```suggestion
Restores a data stream from 7.x snapshots and reindexes all indices into version 8.x. 
```

elastic/logs/challenges/logging-reindex-data-stream.json
```diff
@@ -0,0 +1,93 @@
...
+        "repository": {{ p_snapshot_repo_name | tojson }},
+        "body": {
+          "type": {{ p_snapshot_repo_type | tojson }},
+          "settings": {{ p_snapshot_repo_settings | tojson(indent=2)}}
+        },
+        "include-in-reporting": false
+      }
+    },
+    {
+      "name": "recover-snapshot-{{ p_snapshot_name }}",
```
inqueue, line 35: I am not catching where `p_snapshot_name`, `p_restore_data_streams`, and `p_snapshot_rename_suffix` are set. Should they be in `track.json`?

inqueue (changes requested):

elastic/logs/challenges/logging-reindex-data-stream.json
parkertimmins: Yep, I'm using the values defined here: https://github.com/elastic/rally-tracks/blob/c9c5da5e1670856b8e9fdb9018aeb985923459af/elastic/logs/track.json#L56

They seem to be used by several challenges that use snapshots, so I just reused the same parameters.

inqueue: Right, I know those parameters. Sorry for the noise.

inqueue (approved):
LGTM, nice work

PR #757, opened 2025-03-14
Add ES|QL JOINS track designed to run nightly

luigidellaquila:
New nightly track for ES|QL joins.
The intention is to run it with `ingest_percentage:10`, so that it takes less than one hour to finish.
It runs only with key cardinality = 100k, but it also executes expensive queries, like join on the full dataset.


bpintea (approved):
AFAI can tell, it LGTM.

PR #760, opened 2025-03-20
Measure limit 10 performance

idegtiarenko:
This adds a track with `limit 10` for a sample like queries.

### DIFF
--- a/esql/challenges/default.json
+++ b/esql/challenges/default.json
@@ -43,6 +43,20 @@
           "tags": ["setup"]
         },
         {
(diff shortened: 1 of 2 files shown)


luigidellaquila (approved):
Thanks @idegtiarenko, this looks good, but to make these queries actually run you'll also have to add the operations to the default challenge, see https://github.com/elastic/rally-tracks/blob/master/esql/challenges/default.json#L59-L65

luigidellaquila (approved):
LGTM, thanks!

PR #761, opened 2025-03-20
Fix ignore parameters in dbpedia and msmarco-passage-ranking

jimczi:
The `size` and `track_total_hits` params are ignored in the dbpedia and msmarco-passage-ranking tracks. This change fixes the issue by providing the params as part of the search request body.

### DIFF
--- a/search/mteb/dbpedia/track.py
+++ b/search/mteb/dbpedia/track.py
@@ -103,13 +103,13 @@ class QueryParamsSource:
     def params(self):
         query_obj = self._queries[self._iters]
         query = generate_query(query_obj["text"], self._title_field, self._title_boost, self._text_field, self._text_boost)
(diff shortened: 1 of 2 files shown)


search/mteb/dbpedia/track.py
```diff
@@ -103,13 +103,13 @@ def partition(self, partition_index, total_partitions):
     def params(self):
         query_obj = self._queries[self._iters]
         query = generate_query(query_obj["text"], self._title_field, self._title_boost, self._text_field, self._text_boost)
+        query['track_total_hits'] = self._track_total_hits
```
kderusso, line 106: Nitpick: We use double quotes everywhere else

kderusso (approved):
Thanks for fixing!

PR #765, opened 2025-03-27
Add ingest and search scaling challenges for Wikipedia

jimczi:
This change adds threw new challenges for the Wikipedia track:

- `ingest-autoscale`
- `search-autoscale`
- `ingest-search-autoscale`


kostasb (approved):
Thank you

PR #767, opened 2025-03-31
Wikipedia track - fix StopIteration bug on queries

carlosdelest:
Wikipedia track incorrectly places the try ... except clause, so this causes some benchmarks to hang.

See [this comment](https://github.com/elastic/rally-tracks/commit/ef11d3e981727321185d9d20eead14aab8aae74c#r154636338)

### DIFF
--- a/wikipedia/track.py
+++ b/wikipedia/track.py
@@ -224,19 +224,19 @@ class EsqlSearchParamSource(QueryIteratorParamSource):
         self._query_type = self._params["query-type"]
 
     def params(self):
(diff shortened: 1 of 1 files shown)


gareth-ellis:
The change looks fine - though note we aren't using the `EsqlSearchParamSource` in the nightly that is hanging - we are instead using `QueryParamSource`.

The return is not included in the try ....except, however that should not be an issue, since it is the `next` that would likely be throwing StopIteration

PR #768, opened 2025-03-31
[WIP] Use `itertools.cycle` to iterate over random queries from the query file

fressi-elastic:
This is a follows up over PR #767

### DIFF
--- a/wikipedia/track.py
+++ b/wikipedia/track.py
@@ -28,7 +28,7 @@ def query_samples(k: int, seed: int | None = None) -> list[str]:
         queries_reader = csv.reader(queries_file)
         # It skips the file header
         next(queries_reader)
-        # It reads the file road by road
+        # It reads the file row by row
         for query, probability in queries_reader:
             queries.append(QUERY_CLEAN_REXEXP.sub(" ", query).lower())
             probabilities.append(float(probability))


gareth-ellis:
I need to have a go at running this, but my initial reaction is that I think this changes the behaviour of the track. In rally we have two modes of running a task - either time based, or "unit" based - where we would in the track say run this for 10 minutes, or run this 200 times - by not allowing the iterator to renew, it means that if a user says run this 12000 times, then actually it will only run 10000 times, since the query file has 10000 rows. 

As i say, I have only very quickly looked at this - i need to run myself to validate.

AI-IshanBhatt (commented):
Parentheses in the dataclass decoratore

wikipedia/tests/test_track.py
```diff
@@ -0,0 +1,148 @@
...
+    got = [source.params() for _ in want]
+    assert got == want
+
+
+@dataclasses.dataclass
+class DummyIndex:
+    name: str = "dummy"
+
+
+@dataclasses.dataclass()
```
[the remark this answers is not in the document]
fressi-elastic: When the `cls` parameter is `None` it will [return the `wrap`](https://github.com/python/cpython/blob/24bee4e5cf3b13410e069ac05f542d6170304aa4/Lib/dataclasses.py#L1302) function. There is no actual difference between [using](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass) with or without parenthesis Anyway for coherence I would always use it.

wikipedia/track.py
```diff
@@ -16,23 +17,33 @@
 QUERY_RULES_ENDPOINT: str = "/_query_rules"
 
 QUERY_CLEAN_REXEXP = regexp = re.compile("[^0-9a-zA-Z]+")
+DEFAULT_SEED = hash(__name__)
```
gbanasiak, line 20: I like the idea of pinning the seed. We're doing something similar in `elastic/logs` through default track parameter value: https://github.com/elastic/rally-tracks/blob/f3e4ad5d1e6a0ece30dd1735c8c77e79ec4b5f1a/elastic/logs/README.md?plain=1#L242 https://github.com/elastic/rally-tracks/blob/f3e4ad5d1e6a0ece30dd1735c8c77e79ec4b5f1a/elastic/logs/track.json#L155

Can we document the default in track's README?

wikipedia/track.py
```diff
@@ -69,43 +80,44 @@ def __init__(self, track, params):
...
 
-class QueryIteratorParamSource(ParamSource):
+class RandomQueriesParamSource(ParamSource, metaclass=abc.ABCMeta):
     def __init__(self, track, params, **kwargs):
         super().__init__(track, params, **kwargs)
-        self._batch_size = self._params.get("batch_size", 100000)
-        self._random_seed = self._params.get("seed", None)
-        self._sample_queries = query_samples(self._batch_size, self._random_seed)
-        self._queries_iterator = None
+        self._queries_it = itertools.cycle(query_samples(k=self._params.get("batch_size", 100000), seed=self._params.get("seed")))
```
gbanasiak, line 86: Back to @gareth-ellis point I think this is correct, i.e. `itertools.cycle` will cycle back to the beginning of an array returned by `query_samples()` which is what current code is doing. It's convenient at the cost of doubling memory consumption because `itertools.cycle` stores a local copy of an array I think (see https://docs.python.org/3/library/itertools.html#itertools.cycle).

The `queries.csv` has 10k lines, and we're asking for x10 more as the default `batch_size/k` is 100k. Assuming pure ASCII (1 byte / character), and string length of 32 characters on average, that's an increase in the region of 128k * 32 ~ 4MB. I think that's OK.

gbanasiak (commented):
This looks good overall. I've asked to document the default random seed.

I'd like to hear @elastic/search-relevance opinion. Would you like to pursue with this change? The benefits I see are:
* random seed pinning for _potentially_ better reproducibility of results
* the use of `itertools.cycle` makes code simpler as there's no need to handle iterator exhaustion
* unit tests might help us avoid unintended changes in the future but you might see it as unnecessary boilerplate.

wikipedia/track.py
```diff
@@ -16,23 +17,33 @@
...
 
-def query_samples(k: int, random_seed: int = None) -> List[str]:
     with open(QUERIES_FILENAME) as queries_file:
-        csv_reader = csv.reader(queries_file)
-        next(csv_reader)
-        queries_with_probabilities = list(tuple(line) for line in csv_reader)
+        queries_reader = csv.reader(queries_file)
+        # It skips the file header
+        next(queries_reader)
+        # It reads the file road by road
```
gbanasiak, line 31:
```suggestion
        # It reads the file row by row
```

gbanasiak:
I'm adding `WIP` label in the title to mute alerts. @fressi-elastic Please remove it once you have some free cycles to go back to this.

PR #769, opened 2025-03-31
Enable force_merge for ESQL

nik9000:


fang-xing-esql:
@elasticmachine update branch

PR #778, opened 2025-05-23
http_logs add search with int sort

mayya-sharipova:
This adds sorting on integer fields of "size" and "status".
We are optimizing integer sort in elastic/elasticsearch/pull/127968,
and it would be nice to have dedicated operations for integer sort.

Notice, no target-throughput for these operations, as when optimization is merged
we expect massive speedups.


jimczi (approved):
LGTM

gareth-ellis (approved):
LGTM

PR #781, opened 2025-06-04
Remove fixed refresh interval from k8s-logging template

martijnvg:
This is hiding a performance issue, that we're going to investigate (again).
@inqueue would this be ok for nightly env? iirc the issue that benchmark didn't complete in time was fixed, by introducing `elastic-logs-logsdb-just-basic` config.

Revert "Partial revert of 9fae039: use a 5s refresh for k8s application logs unless the refresh_interval is specified (#766)"

This reverts commit 2b89a0162c62ffc4b70a3587feb562b576e581f2.

Relates to #2463

### DIFF
--- a/elastic/logs/templates/composable/logs-k8-application.log.json
+++ b/elastic/logs/templates/composable/logs-k8-application.log.json
@@ -17,8 +17,7 @@
         }
         {#- non-serverless-index-settings-marker-start #}
         {%- if build_flavor != "serverless" or serverless_operator == true %},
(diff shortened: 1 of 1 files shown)


inqueue (approved):
> would this be ok for nightly env? iirc the issue that benchmark didn't complete in time was fixed, by introducing elastic-logs-logsdb-just-basic config.

Sounds good. LGTM

martijnvg:
Thanks Jason!

PR #782, opened 2025-06-04
Add match phrase tests to Wikipedia track

kderusso:
Adds tests for `match_phrase` in query DSL and ES|QL to the Wikipedia Rally track

### DIFF
--- a/wikipedia/track.py
+++ b/wikipedia/track.py
@@ -263,13 +263,13 @@ class QueryParamSource(QueryIteratorParamSource):
             elif self._query_type == "kql":
                 query_body = {"kql": {"query": f'{ self._params["search-fields"] }:"{ query }"'}}
             elif self._query_type == "match":
-                query_body = {"match": {"content": query}}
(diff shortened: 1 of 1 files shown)


wikipedia/track.py
```diff
@@ -266,6 +268,8 @@ def params(self):
                 query_body = {"bool": {"should": [{"match": {"title": query}}, {"match": {"content": query}}]}}
             elif self._query_type == "term":
                 query_body = {"bool": {"should": [{"term": {"title": query}}, {"term": {"content": query}}]}}
+            elif self._query_type == "match_phrase":
```
carlosdelest, line 271: Now that I see it - I wonder why we're doing match over content and not on content and title as the other queries 🤔 .

We're comparing apples to oranges a bit, as ESQL queries both. I think we should correct that - can be in a separate PR.

WDYT?

carlosdelest (approved):
LGTM! 👍

wikipedia/track.py
kderusso: Good point. I'll update for this PR.

PR #783, opened 2025-06-05
Remove fixed refresh interval from auditbeat template

martijnvg:
The refresh interval is a setting that optionally can be set, and is controlled in `track-custom-shared-settings.json` template. So the fixed refresh interval in audit beat template should be removed.


inqueue (approved):
LGTM

PR #789, opened 2025-06-10
Adding more filter tasks and filter recall for so_vector

benwtrent:
Here I am adding some more filter searches (knn only) and recall tasks for each of them.

I think 10% and 20% hit difficult spots for HNSW and are important to calculate. making them random ensures we don't hit "nice" pockets of the graph where all the nearest neighbors actually match the filter criteria already or a particular part of the graph actually matches filter criteria.

### DIFF
--- a/so_vector/operations/default.json
+++ b/so_vector/operations/default.json
@@ -16,6 +16,7 @@
   "param-source": "knn-recall-param-source",
   "k": 10,
   "num_candidates": 50,
+  "include-in-reporting": false,
   "filter": {
     "exists": {
       "field": "acceptedAnswerId"
@@ -30,7 +31,9 @@
   "num_candidates": 50,
   "filter": {
     "terms": {
-       "tags": ["java"]
+      "tags": [
+        "java"
+      ]
     }
   }
 },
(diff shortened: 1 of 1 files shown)


so_vector/operations/default.json
```diff
@@ -46,6 +83,74 @@
...
+},
+{
+  "name": "knn-search-10-50-random-10-percent",
+  "operation-type": "search",
+  "param-source": "knn-param-source",
+  "k": 10,
+  "num_candidates": 50,
+  "filter": {
+    "range": {
+			"questionId": {
```
john-wagster, line 106: nit: formatting on some of these looks wonky

so_vector/operations/default.json
```diff
@@ -10,6 +10,18 @@
     }
   }
 },
+{
```
mayya-sharipova, line 13: Should we add  "include-in-reporting": false to all recall operations>

mayya-sharipova (approved):
@benwtrent Thanks, nice enhancement of the track!

Indeed, would be nice to cache exact neighbours for filtered queries, as they need to be executed twice.

tteofili (approved):

PR #790, opened 2025-06-11
Make number of shards configurable in dense_vector

thecoop:
Make the shards configurable in `dense_vector`, same as the other vector tracks


gareth-ellis (approved):
LGTM

PR #792, opened 2025-06-12
Adding k100 tasks to so_vector

benwtrent:
The final update to so_vector (at least for my current work concerns).

We should also have a k-100 task to measure the higher k recall and latency values for matchall, filter 10%/20%. 

I didn't do it for ALL the tasks as that would get pretty danged expensive, so the two random filter & matchall seemed good.

I also adjusted the recall tasks to NOT do a warm up. We don't care about performance here (we do that with the knn search tasks), so doing it twice (once for warm up and once for measure) was wasted work.


john-wagster (approved):
lgtm

PR #793, opened 2025-06-13
Switch random_vector track to use data stream

jimczi:
Use a data stream to index random vectors and update the parameters.


kderusso (approved):
LGTM

PR #797, opened 2025-06-19
Remove duplicate task definition in msmarco

kostasb:
Running the track yields error:

> esrally.track.loader.TrackSyntaxError: Track 'msmarco-v2-vector' is invalid. Challenge 'ingest-search-autoscale' contains multiple tasks with the name 'parallel-search-1-c50-s10-t100'. Please use the task's name property to assign a unique name for each task.

The `ingest-search-autoscale` has a duplicate block that creates the same task twice. This removes the extraneous block.


jimczi (approved):
LGTM, thanks for fixing

PR #812, opened 2025-07-08
[8.15] Use distribution version in ITs

gbanasiak:
Non-default ITs do not use _released_ builds, so we are blind to failures caused by using features available only in snapshot builds. This PR changes that for the most recent non-default branch `8.15`. The intention is to configure `--distribution-version` in every new "version" branch created going forward.

### DIFF
--- a/nyc_taxis/operations/default.json
+++ b/nyc_taxis/operations/default.json
@@ -966,7 +966,10 @@
       "name": "avg_passenger_count_esql_doc_partitioning",
       "operation-type": "esql",
       "query" : "FROM nyc_taxis | stats avg(passenger_count)",
-      "body": { "pragma": { "data_partitioning": "doc" } }
+      "body": {
+        "accept_pragma_risks": true,
+        "pragma": { "data_partitioning": "doc" }
+      }
     },
   {%- endif -%}{# non-serverless-doc-partitioning-marker-end #}
     {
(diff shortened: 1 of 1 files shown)


nyc_taxis/challenges/default.json
```diff
@@ -938,14 +938,6 @@
...
           "iterations": 50
         },
-      {# non-serverless-doc-partitioning-marker-start #}{%- if build_flavor != "serverless" -%}
-        {
-          "operation": "avg_passenger_count_esql_doc_partitioning",
-          "clients": 1,
-          "warmup-iterations": 10,
-          "iterations": 50
-        },
-      {%- endif -%}{# non-serverless-doc-partitioning-marker-end #}
```
favilo, line 948: Why did we get rid of these at the same time?

gbanasiak: An alternative to `pragma` removal would be a track param which defaults to `false`, perhaps that's a better idea. @elastic/es-analytical-engine Any preferences?

dnhatn: You can add a URL parameter `accept_pragma_risks=true `to allow pragmas in release builds.

favilo (approved):
Looks good. Thanks for that pragma change. That is certainly an improvement overall, too.

PR #814, opened 2025-07-10
Fix oversample params for knn tracks

benwtrent:
`oversample: 0` is a valid parameter value, it indicates no oversampling and overwrites any default oversampling applied at the index level.

This fixes msmarco-v2-vector and openai_vector to allow skipping oversampling


jimczi (approved):
LGTM

PR #816, opened 2025-07-11
Streams challenge

flash1293:
Related to https://github.com/elastic/streams-program/issues/368

Set up the `logs` data stream like streams does it and throw the `elastic/logs` dataset at it.

This challenge measures indexing performance and storage usage, a comparable challenge to measure query performance is located in https://github.com/elastic/rally-tracks/blob/master/elastic/logs/challenges/logging-chicken.json

The processing pipeline still contains a couple polyfills for features that didn't land yet, but this is expected for now - I will keep it in sync with what happens in practice.


elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,315 @@
...
+        "operation-type": "raw-bulk",
+        "param-source": "processed-source",
+        "time-format": "milliseconds",
+        "profile": "fixed_interval",
+        "bulk-size": {{ p_bulk_size }},
+        "detailed-results": true
+      },
+      "clients": {{ p_bulk_indexing_clients }}{% if p_throttle_indexing %},
+      "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
+      "schedule": "timestamp-throttler",
```
gbanasiak, line 286:
```suggestion
      "clients": {{ p_bulk_indexing_clients }},
      "ignore-response-error-level": "{{error_level | default('non-fatal')}}",
      {% if p_throttle_indexing %}
      "schedule": "timestamp-throttler",
```

I know this was copied over, but let's not repeat the mistake of combining throttling with ignore errors configuration. These are orthogonal things.

gbanasiak (commented):
Integration tests grind to a halt in a test that happens _after_ the newly added streams. I haven't looked why exactly that happens but it might be necessary to add cleanup of items added in `logging-streams` challenge in https://github.com/elastic/rally-tracks/blob/master/elastic/logs/tasks/index-setup.json which is shared between challenges. Note integration tests for `elastic/logs` _reuse_ the same ES cluster. Whatever is left by the previous tests and not cleaned-up by current test can potentially affect the outcome.

Similarly, the new challenge should be [removing](https://github.com/elastic/rally-tracks/blob/24c4b0b052a16758618e07c5159e20fffb69632c/elastic/logs/tasks/index-setup.json#L21-L25) data streams.

flash1293:
Thanks for the comments, @gbanasiak - I addressed your comments:
* Move `ignore-response-error-level` out of throttling config
* Clean up after tests are run (let's see whether that makes it work)

elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,331 @@
...
+    },
+    {
+      "name": "logs-index-template",
+      "operation": {
+        "operation-type": "create-composable-template",
+        "template": "logs@stream",
+        "body": {
+          "index_patterns": [
+            "logs"
+          ],
```
gbanasiak, line 225: This should match data streams created by the track (see [here](https://github.com/elastic/rally-tracks/blob/24c4b0b052a16758618e07c5159e20fffb69632c/elastic/logs/track.json#L181-L221)). Please change from `logs` to `logs-*`.

elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,331 @@
...
+        "request-timeout": 60
+      }
+    },
+    {
+      "name": "compression-stats-after-merge",
+      "operation": {
+        "operation-type": "compression-statistics",
+        "data-stream": "logs"
+      }
+    },
```
gbanasiak, line 314: Do you care about compression stats and force merging? If not, please simplify removing those. The data streams and index pattern are incorrect here, see other comments.

gbanasiak (commented):
The problem is definitely due to state in which ES cluster is left after streams challenge test. The streams challenge test takes about a minute to complete which is fine. The following test was taking around a minute earlier, and now gets stuck for more than 1h.

Please take a look at further comments. If this doesn't help, the test should be reproduced locally, and state of ES cluster verified.

elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,331 @@
...
+    {
+      "name": "delete-data-stream",
+      "operation": "delete-data-stream",
+      "data-stream": "logs"
+    },
```
gbanasiak, line 319: This does not remove data streams created by this challenge. Can you remove `data-stream` or change it to `logs-*` ?

flash1293:
@gbanasiak Thanks for the look.

This is working with `logs` instead of `logs-*` on purpose - see https://github.com/elastic/rally-tracks/pull/816/files#diff-352462dbff160a0065b6f5dadc84cf7005070e2353a0663774cbf3b9d62165dbR267 - no data is actually put into data streams other than `logs`.

I solves this with the reroute because that was the easiest way, maybe there is a better one to represent this.

My understanding was that it's https://github.com/elastic/rally-tracks/pull/816/files#diff-352462dbff160a0065b6f5dadc84cf7005070e2353a0663774cbf3b9d62165dbR267 what is actually creating these data streams, and the new challenge is not calling the setup task on purpose (because it's all inlined for this special case).

I just re-tested this locally and it seems to leave behind a `logs` datastream, will try to find out what's happening there

flash1293:
Should we maybe introduce a param similar to `p_mapping` here? https://github.com/elastic/rally-tracks/blob/24c4b0b052a16758618e07c5159e20fffb69632c/elastic/logs/track.json#L223

and based on that change the data streams mentioned in `track.json`?

I tried to contain the necessary changes to the challenge, but maybe that's the wrong way of doing it.

flash1293:
Still times out, but it gets a little further now

flash1293:
> Note logs@custom pipeline is not removed, which affects remaining test runs even if logs data stream is gone.

*facepalm* - will fix, thanks!

> Is the reroute absolutely necessary? Couldn't you achieve the same through conditional modification of composable templates used in elastic/logs to stay within the overall set of entities created and destroyed by the challenges in this track?

The name of the data stream doesn't matter too much, but we need all of the data from the different data sets in a single data stream, that's the whole idea. Not sure how to achieve that without a reroute.

We could list the `logs` data stream in `track.json`, this way it should be cleaned up automatically (not sure whether it breaks something if it isn't always there, but I would guess no). For the reroute I guess the easiest approach is to have the `logs@custom` reroute and remove it at the end of the challenge.

I guess forking the track completely is still possible, but I would like to avoid if we can.

gbanasiak:
> We could list the `logs` data stream in `track.json`, this way it should be cleaned up automatically (not sure whether it breaks something if it isn't always there, but I would guess no). For the reroute I guess the easiest approach is to have the `logs@custom` reroute and remove it at the end of the challenge.

Ok, let's try that. Can we use something more telling than `logs`, e.g. something with "streams" in the name? To make it easier to correlate with specific setup/challenge. Also wouldn't hurt to update track's README once we're sure this works fine.

elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,102 @@
...
+          "processors": [
+            {
+              "reroute": {
+                "destination": "logs"
+              }
+            }
+          ]
+        }
+      }
+    },
```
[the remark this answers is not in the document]
gbanasiak: @elastic/es-storage-engine Looks like the only place where we refer to `logs-k8-application` explicitly is in https://elasticsearch-benchmarks.elastic.co/#tracks/logsdb-ccr/nightly/default/90d benchmark, see `CCR Details` section. Would you be OK to edit filters there?

elastic/logs/pipelines/logs@stream.processing.json
```diff
@@ -0,0 +1,57 @@
...
+    },
+    {
+      "dot_expander": {
+        "field": "*",
+        "ignore_failure": true
+      }
+    },
+    {
+      "normalize_for_stream": {}
+    },
```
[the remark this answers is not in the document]
flash1293: yeah, it's not available for serverless.

flash1293:
Let's see whether it works with your latest change.

elastic/logs/challenges/logging-streams.json
```diff
@@ -0,0 +1,102 @@
...
+    {
+      "name":"reset-pipelines",
+      "description": "Put pipelines back so k8s-app pipeline is restored",
+      "operation": {
+        "operation-type": "create-pipeline",
+        "param-source": "add-track-path"
+      }
+    }
```
gbanasiak, line 100: Is this required? We are creating pipelines at the beginning of other challenges anyway.

flash1293: It's not, that was me trying to leave the state as clean as possible. Removed.

flash1293:
@gbanasiak It shouldn't be backported, just 9.2 and soon (hopefully) serverless should be supported.

PR #823, opened 2025-07-22
Make streams test compatible with chicken test

flash1293:
Make sure the `logs` stream works with the `mapped = "unmapped"` parameter. This isn't the case right now because all other data streams are captured by the default `logs-*-*` mappings, but `logs` is not.

### DIFF
--- a/it/logs/test_logs.py
+++ b/it/logs/test_logs.py
@@ -21,26 +21,6 @@ from it.logs import BASE_PARAMS, params
 
 pytest_rally = pytest.importorskip("pytest_rally")
 
-BASE_PARAMS = {
-    "start_date": "2021-01-01T00-00-00Z",
-    "end_date": "2021-01-01T00-01-00Z",
-    "max_total_download_gb": "18",
-    "raw_data_volume_per_day": "72GB",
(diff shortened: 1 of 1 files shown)


it/test_logs.py
```diff
@@ -66,6 +66,15 @@ def test_logs_streams(self, es_cluster, rally):
         )
         assert ret == 0
 
+    def test_logs_chicken(self, es_cluster, rally):
+        custom = {"mapping": "unmapped"}
+        ret = rally.race(
+            track="elastic/logs",
+            challenge="logging-chicken",
```
gbanasiak, line 73: This is `logging-insist-chicken`, see https://github.com/elastic/rally-tracks/blob/502a4112b7c0bd5a71c5fea5e52d2117ba027883/elastic/logs/challenges/logging-chicken.json#L3 We could rename `logging-chicken.json` to `logging-insist-chicken.json` to avoid this in the future.

flash1293:
Kind of obvious in hindsight, but this has the same issue in the other direction... Not sure what the least hacky solution for this would be - any thoughts, @gbanasiak ?

Issue #828, opened 2025-07-29
elastic/shared component template update is broken


inqueue:
Updating component templates in elastic tracks is broken after https://github.com/elastic/elasticsearch/pull/131536.

```sh
Traceback (most recent call last):
  File "/home/esbench/rally/esrally/driver/driver.py", line 1937, in __call__
    total_ops, total_ops_unit, request_meta_data = await execute_single(runner, self.es, params, self.on_error)
                                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/rally/esrally/driver/driver.py", line 2031, in execute_single
    return_value = await runner(es, params)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/rally/esrally/driver/runner.py", line 298, in __call__
    return await self.delegate(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/rally/esrally/driver/runner.py", line 413, in __call__
    return_value = await self.delegate(*args)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/rally/esrally/driver/runner.py", line 346, in __call__
    return await self.delegate(self.client_extractor(args[0]), *args[1:])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/.rally/benchmarks/tracks/default/elastic/shared/runners/update_custom_templates.py", line 28, in update_custom_templates
    await es.cluster.put_component_template(name=name, body={**original, **custom})
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esbench/.local/lib/python3.11/site-packages/elasticsearch/_sync/client/utils.py", line 414, in wrapped
    return api(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
TypeError: ClusterClient.put_component_template() got an unexpected keyword argument 'created_date_millis'
```

https://github.com/elastic/rally-tracks/blob/1480089a7a8ad25ef07e639b4248af9ec92980b7/elastic/shared/runners/update_custom_templates.py#L19-L29

[...]

PR #829, opened 2025-07-29
[elastic/shared] Remove system managed template fields

inqueue:
Remove system fields when updating templates.

Closes https://github.com/elastic/rally-tracks/issues/828

### DIFF
--- a/elastic/shared/runners/update_custom_templates.py
+++ b/elastic/shared/runners/update_custom_templates.py
@@ -25,6 +25,9 @@ async def update_custom_templates(es, params):
         name = template["name"]
         if name.endswith("@custom"):
             original = template["component_template"]
+            # Remove fields added by https://github.com/elastic/elasticsearch/pull/131536
+            original.pop("created_date_millis", None)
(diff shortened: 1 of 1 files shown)


elastic/shared/runners/update_custom_templates.py
```diff
@@ -25,6 +25,9 @@ async def update_custom_templates(es, params):
         name = template["name"]
         if name.endswith("@custom"):
             original = template["component_template"]
+            # Remove fields added by https://github.com/elastic/elasticsearch/pull/131536
+            original.pop("created_date_millis", None)
+            original.pop("modified_date_millis", None)
```
szybia, line 30: just for transparency, doubt a problem, but we can also return `{created,modified}_date` which is human-readable, which is only returned if we supply the `?human=true` HTTP param

but doubt that's an issue here...

inqueue: It isn't an issue here, but that would be a nice to have for the API.

PR #833, opened 2025-08-18
Remove source excludes in all vector tracks

jimczi:
Now that is #131907 merged, remove all the source exclusion from our tracks since this is handled by default.


mromaios (approved):
LGTM

PR #837, opened 2025-08-26
Implement KNN comparison with ESQL for so_vector

svilen-mihaylov-elastic:


so_vector/track.py
```diff
@@ -78,12 +79,19 @@ def params(self):
...
+        self._k = self._params.get("k", 10)
+        self._num_candidates = self._params.get("num_candidates", 50)
+
         query_vec = self._queries[self._iters]
         self._iters += 1
         if self._iters >= self._maxIters:
             self._iters = 0
 
         if self._exact_scan:
+            if self._in_esql_mode:
```
svilen-mihaylov-elastic, line 92: Not sure how to handle this (if we need to at this point)

so_vector/track.py
```diff
@@ -98,18 +106,34 @@ def params(self):
...
         else:
-            result["body"] = {
-                "knn": {
-                    "field": "titleVector",
-                    "query_vector": query_vec,
-                    "k": self._params.get("k", 10),
-                    "num_candidates": self._params.get("num_candidates", 50),
+            if self._in_esql_mode:
+                # Construct options JSON.
+                options_param = '{"num_candidates":' + str(self._num_candidates)
```
carlosdelest, line 111: This will change to "min_candidates" in https://github.com/elastic/elasticsearch/pull/132944. `k` will also be removed, and instead we will need to specify a LIMIT to the query.

It's not a blocker for this PR, but just a heads up that we will need to change this when the above gets merged.

so_vector/track.py
carlosdelest: We can use the vector similarity functions to generate an exact query, using something like

`EVAL score = V_DOT_PRODUCT('titleVector', params.query) + 1.0`

and sorting by score and applying a LIMIT

carlosdelest (commented):
Looks good! 

I think we can use `param-source` to have specific param handling for ESQL related operations, and use `operation-type` to help run specific ESQL workloads.

so_vector/operations/default.json
```diff
@@ -10,6 +10,14 @@
     }
   }
 },
+{
+  "name": "esql-knn-search-10-50-acceptedAnswerId",
+  "operation-type": "esql",
```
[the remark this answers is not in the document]
svilen-mihaylov-elastic: Left a comment below, and will resolve here. "esql" is a built-in operation-type (in the rally framework) AFAICT.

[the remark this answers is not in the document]
carlosdelest: good point. This track creates a couple of custom types so I thought it would be good to have the same. Let's keep using the `esql` type to avoid specific runners, and then we can fine tune if necessary to run specific tasks. Thanks!

so_vector/track.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, track, params, **kwargs):
         self._index_name = params.get("index", default_index)
         self._cache = params.get("cache", False)
         self._exact_scan = params.get("exact", False)
+        self._in_esql_mode = params.get("operation-type", "") == "esql"
```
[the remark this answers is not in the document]
svilen-mihaylov-elastic: Update, please take another look

so_vector/challenges/default.json
```diff
@@ -47,6 +49,15 @@
       "iterations": 100,
       "clients": 1
     },
+{% if is_esql_enabled %}
```
svilen-mihaylov-elastic, line 52: @benwtrent Decided to make each esql conditional to avoid complications with the serverless conditional checks towards the end of the file

esbenchmachine:
@svilen-mihaylov-elastic

Backporting entails: 

Thank you!

PR #838, opened 2025-08-27
ES|QL: Add queries for LOOKUP JOIN with multiple join keys

luigidellaquila:
Adding some queries for joins on multiple keys

### DIFF
--- a/joins/operations/default.json
+++ b/joins/operations/default.json
@@ -15,7 +15,8 @@
         "lookup_idx_1000_f10",
         "lookup_idx_100000_f10",
         "lookup_idx_200000_f10",
-        "lookup_idx_500000_f10"
+        "lookup_idx_500000_f10",
+        "lookup_idx_1000_100000_200000_f10"
       ],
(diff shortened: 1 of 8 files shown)


joins/operations/default.json
```diff
@@ -146,6 +146,21 @@
       "query": "FROM join_base_idx | lookup join lookup_idx_{{key_suffix[i]}}_f10 on key_{{key_suffix[i]}} | where lookup_keyword_0 == \"val 10 rep 0\"",
       "body": { "pragma": { "max_concurrent_shards_per_node": {{max_concurrent_shards_per_node | default(10)}} } }
     },
+
+
+    {
```
julian-elastic, line 151: Do we have the same test but with 1 key? Ideally we want to see how the performance is affected for the same query going from 1 key to 2 keys lookup join. I think we might need to actually change one of the indexes and make sure that there are the same 2 columns without eval for that. Otherwise it seems this would detect regressions, but it will not make sure that the performance with 2 keys is at least comparable to 2x the 1key join performance.

PR #842, opened 2025-09-04
Add elastic/logs patterned-text queries challenge

parkertimmins:
Add challenge which queries message field with several term and phrase queries. This is meant to test the patterned_text mapping type, but can also be used to test match_only_text message fields.


elastic/logs/challenges/logging-querying-patterned-text.json
```diff
@@ -0,0 +1,132 @@
+{% import "rally.helpers" as rally %}
+{
+  "name": "logging-querying-patterned-text",
```
martijnvg, line 3: I think we can include these queries into the `logging-querying` challenge (behind the `p_include_esql_queries` flag)? Then we don't need another nightly config and can just update the existing dashboard?

martijnvg (commented):
Looks good! I just think we can include into the existing challenge?

martijnvg (approved):
LGTM

PR #844, opened 2025-09-04
Fix so_vector for serverless operator

ebadyano:
Misplaced comma


ebadyano:
rally-tracks servelss build failed due to time out unrelaed to the change

PR #847, opened 2025-09-09
Remove routing_path from tsdb index template

felixbarny:
Setting this is not necessary as the routing path is set automatically for data streams and this template defines `"data_stream": {}`. It would also prevent an optimization added in https://github.com/elastic/elasticsearch/pull/132566.


gareth-ellis (approved):
LGTM 

Fyi @AI-IshanBhatt

martijnvg (approved):
LGTM

PR #859, opened 2025-09-17
Run serverless CI only on PRs targeting master

NickDris:
Test [triggers](https://github.com/elastic/rally-tracks/blob/master/catalog-info.yaml#L28-L64) for serverless rally-tracks are fired off whenever a PR is created against all branches without exception, which is redundant since serverless rally-tracks should be tested only against main Elasticsearch version. 

This change filters the base branch of a PR to be 'master' so we can test only the PRs that target master branch.


gbanasiak (commented):
This pipeline runs not only for PRs but also in a schedule. I would expect the condition to cover this case too, e.g. `... or build.source == "schedule"`.

gbanasiak (commented):
Provided a comment about the condition earlier.

gbanasiak:
@elasticmachine update branch

esbenchmachine:
@NickDris

Backporting entails: 

Thank you!

PR #861, opened 2025-09-18
Adds benchmarks for INLINESTATS

ncordon:
This PR adds benchmarks to test out several things:

## `INLINESTATS` queries counterparts for existing `STATS` ones

Namely these ones (segment paritioning and docs partitioning)

```sql
FROM nyc_taxis | stats avg(passenger_count)
FROM nyc_taxis | where total_amount > 60 and rate_code_id==\"2\"| stats avg(passenger_count)
```

Hence why I included the tag `inlinestats` in the `STATS` counterparts existing queries.

## Queries to compare `INLINESTATS` and `STATS`

These queries have the following shape:

```sql
FROM nyc_taxis METADATA _id | LIMIT 1000 | stats avg(passenger_count) by _id
FROM nyc_taxis METADATA _id | LIMIT 1000 | inlinestats avg(passenger_count) by _id
```

The reason for grouping by an unique identifier and limiting the number of rows is so that we can compare performance under the same conditions. `INLINESTATS` is going to append an extra result to each existing row, so we want to measure the impact of that extra step.

## Queries chaining `INLINESTATS` after `STATS` and viceversa

For example:

```sql
FROM nyc_taxis | inlinestats a = avg(passenger_count) | stats avg(a)
FROM nyc_taxis | stats a = avg(passenger_count) | inlinestats avg(a)
```

## Queries with several `STATS` / `INLINESTATS`

For example:

```sql
FROM nyc_taxis | LIMIT 1000 | inlinestats sum = sum(passenger_count), count = count(*), avg = avg(passenger_count)
FROM nyc_taxis | LIMIT 1000 | stats sum = sum(passenger_count), count = count(*), avg = avg(passenger_count)
```

## Queries chaining several `INLINESTATS` one after another

```sql
FROM nyc_taxis | LIMIT 1000 | inlinestats s1 = sum(passenger_count)
FROM nyc_taxis | LIMIT 1000 | inlinestats s1 = sum(passenger_count) | inlinestats s2 = sum(s1)
FROM nyc_taxis | LIMIT 1000 | inlinestats s1 = sum(passenger_count) | inlinestats s2 = sum(s1) | inlinestats s3 = sum(s2)
```

### DIFF
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -994,61 +994,61 @@
         {
           "operation": "avg_tip_percent_aggregation",
           "clients": 1,
-          "warmup-iterations": 10,
-          "iterations": 50
+          "warmup-iterations": 5,
+          "iterations": 10
         },
         {
           "operation": "avg_tip_percent_esql_segment_partitioning",
           "clients": 1,
-          "warmup-iterations": 10,
-          "iterations": 50
+          "warmup-iterations": 5,
+          "iterations": 20
         },
       {# non-serverless-doc-partitioning-marker-start #}{%- if build_flavor != "serverless" -%}
         {
           "operation": "avg_tip_percent_esql_doc_partitioning",
           "clients": 1,
-          "warmup-iterations": 10,
-          "iterations": 50
+          "warmup-iterations": 5,
+          "iterations": 20
         },
       {%- endif -%}{# non-serverless-doc-partitioning-marker-end #}
         {
           "operation": "avg_amount_group_by_integer_aggregation",
           "clients": 1,
-          "warmup-iterations": 10,
-          "iterations": 50
+          "warmup-iterations": 5,
+          "iterations": 20
         },
         {
           "operation": "avg_amount_group_by_integer_esql_segment_partitioning",
           "clients": 1,
-          "warmup-iterations": 10,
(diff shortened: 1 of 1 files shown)


nyc_taxis/challenges/default.json
```diff
@@ -936,14 +936,37 @@
           "operation": "avg_passenger_count_esql_segment_partitioning",
           "clients": 1,
           "warmup-iterations": 10,
-          "iterations": 50
+          "iterations": 50,
+          "tags": ["inlinestats"]
```
ncordon, line 940: Does adding a tag on an existing challenge have unintended consequences (like the historical for the data resetting or something)?

nyc_taxis/operations/default.json
```diff
@@ -961,7 +961,246 @@
       "operation-type": "esql",
       "query" : "FROM nyc_taxis | stats avg(passenger_count)"
     },
+    {
+      "name": "inlinestats_avg_esql_segment_partitioning",
+      "operation-type": "esql",
+      "query" : "FROM nyc_taxis | inlinestats avg(passenger_count)"
```
ncordon, line 967: It's giving me OCD to have some keywords uppercase and some lowercase but this is how the existing stats queries were written 😂

PR #864, opened 2025-09-19
Fix index.refresh_interval setting

mayya-sharipova:
The setting should be index.refresh_interval instead of current index.index_refresh_interval


mayya-sharipova:
@elasticmachine test this please

PR #875, opened 2025-10-03
CI builds ES revision from sources based on es-version file

NickDris:
- Github workflows now execute CI with the additional --revision option, effectively building a specific revision of Elasticsearch from sources
- es-version added to root directory which determine what is the value of the --revision argument that will be used in CI.


es-version
```diff
@@ -0,0 +1 @@
+current
```
gbanasiak, line 1: Nit: EOL

gbanasiak (approved):

PR #889, opened 2025-10-15
Add backport reminder and reorganize the backporting actions.

NickDris:
Introducing backport.py script that interacts with GH API to
- Label merged PRs lacking a 'backport pending' or a version vX(.Y) label.
- Post a backporting reminder as a comment in the PR.

Github Workflow changes:
- Backporting action is separated from the rest of backporting, like it was initially.
- Backporting reminder and labeler added in a workflow that can be also run manually.

Backport reminder workflow:
- Mainly uses backport.py to add labels and post reminders in merged PRs.
- Runs either on schedule, manually or on closing PR, discerning the mode of operation 
with the github.event_name.
- Based on the mode of operation, backport.py either processes a single PR, or 
fetches the list of PRs, merged the last X days from GH API 
- Checking PR merged status is made inside backport.py.

Inputs:
- lookback_days: Look for merged PRs within a lookback window.
- pending_label_age_days: Post reminder based on how old is the last reminder.


.github/scripts/backport.py
```diff
@@ -0,0 +1,408 @@
...
+"""Backport CLI (label + remind)
+
+Label merged PRs lacking a version label with 'backport pending' or send reminder
+comments on such PRs until a version label (e.g. v9.2) is added.
+
+Quick usage:
+	backport.py help
+	backport.py --event-name pull_request_target label
+	backport.py --repo owner/repo --token TOKEN --event-name schedule label --lookback-days 7
+	backport.py --repo owner/repo --token TOKEN --event-name workflow_dispatch remind --lookback-days 30 --pending-label-age-days 14
```
gbanasiak: Reading a token should be done from env variable instead of command line. Command line leaves a trace in shell history which can be avoided but requires extra care.

gbanasiak (commented):
Initial high-level feedback/questions:
1) Having a CLI tool which can be used in GH actions as well as independently is a great idea.
2) A dry-run option is highly recommended. It would allow for testing (part of PR review).
3) Token should not be passed through command-line option, see below.
4) The tool uses standard libraries only it seems. Is this intentional? Example: why `argparse` and something newer, like `click`?
5) `backport.action.yml` still references `scripts/backport-pending.py` which gets removed.

gbanasiak:
For the record, we ultimately resigned from `click` for better portability.

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+# specific language governing permissions and limitations
+# under the License.
+
+"""Backport CLI (label + remind)
+
+Label merged PRs lacking a version label with 'backport pending' or send reminder
+comments on such PRs until a version label (e.g. v9.2) is added.
+
+Quick usage:
+	backport.py help
```
gbanasiak: `help` alone is not helpful because `--trigger` is required

```
% python .github/scripts/backport.py help                   
usage: backport.py [-h] [--repo REPO] --trigger {pull_request_target,schedule,workflow_dispatch} [--dry-run] [--debug] {label,remind,help} ...
backport.py: error: the following arguments are required: --trigger
```

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+    CONFIG["log_level"] = "DEBUG" if getattr(args, "debug", False) else "INFO"
+    CONFIG["command"] = getattr(args, "command", None)
+    if getattr(args, "repo", None):
+        CONFIG["repo"] = args.repo
+    setup_logging(CONFIG["log_level"])
+    require_mandatory_vars()
+
+
+# ----------------------------- Logging -----------------------------
+logger = logging.getLogger("backport")
```
gbanasiak: Nit: logger instance is typically declared at the top of the module

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+                logger.debug(f"PR #{info.number}: No label action needed")
+        except Exception as e:
+            logger.error(f"[label] PR #{pr.get('number','unknown')}: {e}")
+            rc = 1
+    return rc
+
+
+# ----------------------------- Reminder Logic -----------------------------
+COMMENT_MARKER_BASE = "<!-- backport-pending-reminder -->"  # static for detection
+REMINDER_BODY = "A backport is pending for this PR.\n\n" "Please add an appropriate version label (e.g. v9.2)\n" "Thank you!"
```
gbanasiak: I would move this to the top together with other constants.

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+            type=int,
+            required=True,
+            help="Days to look back for updated merged PRs",
+        )
+        p_remind.add_argument(
+            "--pending-label-age-days",
+            type=int,
+            required=True,
+            help="Days between reminders for the same PR",
+        )
```
gbanasiak: Is there a strong reason to keep those required instead of providing reasonable defaults?

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+            "--repo",
+            help="Target repository in owner/repo form (overrides GITHUB_REPOSITORY env)",
+            required=False,
+        )
+        parser.add_argument(
+            "--trigger",
+            choices=["pull_request_target", "schedule", "workflow_dispatch"],
+            required=True,
+            help="Mode of operation",
+        )
```
gbanasiak: Nit: This option feels a bit awkward in the command-line use. There's no difference between `schedule` and `workflow_dispatch` so I think I would just add PR mode flag which is disabled by default.

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
```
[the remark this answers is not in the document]
NickDris: You're right, thanks. I prefer to keep this module portable (meaning we can port this in whichever repo we want), so I will add a shebang. This way we can execute the script directly in any repo, without needing to specify the Python interpreter on the command line.

.github/scripts/backport.py
```diff
@@ -0,0 +1,461 @@
...
+
+
+def is_dry_run() -> bool:
+    return bool(CONFIG.get("dry_run"))
+
+
+def require_mandatory_vars() -> None:
+    """Validate critical environment / CLI inputs using CONFIG."""
+    if not CONFIG.get("token") and not is_dry_run():
+        raise RuntimeError("Missing BACKPORT_TOKEN from environment.")
```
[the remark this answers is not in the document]
gbanasiak: > So, I believe this is making the tool more flexible for testing.

I see.

My view was without token we are not getting much value from `--dry-run`. For example:

```
% python .github/scripts/backport.py --dry-run --debug --trigger=schedule label --lookback-days=7
[2025-10-23 14:35:43] INFO [label (dry-run)]: Would list PRs with filter 'repo:elastic/rally-tracks is:pr is:merged' since 2025-10-16T12:35:43.011886+00:00
[2025-10-23 14:35:43] DEBUG [label (dry-run)]: Would GET https://api.github.com/search/issues?q=repo%3Aelastic%2Frally-tracks+is%3Apr+is%3Amerged+updated%3A%3E%3D2025-10-16&per_page=100&page=1 body={}
[2025-10-23 14:35:43] ERROR [label (dry-run)]: Runtime error: HTTP 401 Unauthorized {
  "message": "Bad credentials",
  "documentation_url": "https://docs.github.com/rest",
  "status": "401"
}
```

This tells me what the _first_ GET would be, but I still don't see the entire flow, i.e. iterating through all the PRs fulfilling conditions. So I need to use the token anyway.

.github/scripts/backport.py
NickDris: Seems legit. I added a --pr-mode with store_true instead of choosing between gh triggers.

NickDris:
> Thank you for modifications. I did another pass of the code and tested with `--debug --dry-run`. Check my comment under `needs_pending_label()`.
> 
> Edit: Do we need any automation to remove `backport pending` label if version label was applied but `backport pending` label not removed?

@gbanasiak note in my [reply](https://github.com/elastic/rally-tracks/pull/889#discussion_r2455439787) that I prefer moving the removal of `backport pending` label to user responsibility, in order to align with more consistent backporting status of each PR.

.github/scripts/backport.py
```diff
@@ -0,0 +1,479 @@
...
+
+    This centralizes setup so other entry points (tests, future subcommands)
+    can reuse consistent initialization semantics.
+    """
+    CONFIG["dry_run"] = bool(getattr(args, "dry_run", False))
+    CONFIG["log_level"] = "DEBUG" if getattr(args, "debug", False) else "INFO"
+    CONFIG["command"] = getattr(args, "command", None)
+    if getattr(args, "repo", None):
+        CONFIG["repo"] = args.repo
+    setup_logging(CONFIG["log_level"])
```
gbanasiak: `CONFIG["log_level"]` is `bool | str | None` but `str` expected in `setup_logging()`. You can use `assert` in the preceding line to address that.

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+import json
+import logging
+import os
+import re
+import sys
+import time
+import urllib.error
+import urllib.request
+from dataclasses import dataclass
+from typing import Any, Dict, Iterable, List
```
fressi-elastic: The last reccomendations of Python typing is using bare `list` instead of `typing.List` and `dict` instead of `typing.Dict`. `Iterable` should be better imported from `collections.abc`

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+REMINDER_BODY = (
+    "A backport is pending for this PR. Please add all required version labels (e.g. v9.2).\n\n"
+    "If it supports past released versions, add each released branch label (e.g. v9.1 and v9.2).\n"
+    "If it only targets a future version, wait until that version label exists.\n"
+    "After completing the backport(s), manually remove the 'backport pending' label.\n"
+    "We do that to ensure proper tracking of backport status.\n\n"
+    "Thank you!"
+)
+
+CONFIG: Dict[str, bool | str | None] = {
```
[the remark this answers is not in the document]
fressi-elastic:
```Python
@dataclasses.dataclass
class BackportConfig:
  token: str | None = os.environ.get("BACKPORT_TOKEN")
  repo: str | None = os.environ.get("GITHUB_REPOSITORY")
  dry_run: bool = False
  log_level: int = logging.INFO
  command: str | None = None
```

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+    "log_level": "INFO",  # INFO by default, set to DEBUG if --debug passed
+    "command": None,  # set post-parse for logging
+}
+
+
+# ----------------------------- Logging -----------------------------
+logger = logging.getLogger("backport")
+
+
+def setup_logging(level: str) -> None:
```
fressi-elastic: `def setup_logging(level: int | str) -> None:`

You can leave `logging` library making conversions from string to integers. I personally found integers more handy to use as they can be eventually compared.

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+    assert level.upper() in {"CRITICAL", "ERROR", "WARNING", "INFO", "DEBUG", "NOTSET"}, "Invalid log level"
```
fressi-elastic: There is no need for making conversion from string to integer or validating levels here as logging library already does it.

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+    labels: List[str]
+
+    def __init__(self, pr: Dict[str, Any]):
+        raw_number = pr.get("number")
+        if raw_number is None and pr.get("url"):
+            try:
+                raw_number = int(pr.get("url", "/").rstrip("/").split("/")[-1])
+            except TypeError:
+                raw_number = -1
+        self.number = raw_number if isinstance(raw_number, int) else int(raw_number)
```
fressi-elastic: `int(raw_number)` can be used on integers too. Therefore above code could be much simpler:

```Python
self.number = int(pr.get("number") or pr.get("url", "").rstrip("/").strip().rsplit("/", 1) or -1)
```

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+
+
+def remove_label(pr_number: int, label: str) -> None:
+    repo = CONFIG.get("repo")
+    issues_labels_api = f"/repos/{repo}/issues/{pr_number}/labels"
+    gh_request(issues_labels_api, method="DELETE", body={"labels": [label]})
+    logger.info(f"Removed label '{label}' from PR #{pr_number}")
+
+
+def run_label(prefetched_prs: List[Dict[str, Any]], remove: bool) -> int:
```
fressi-elastic: `list[dict[str, Any]]`

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+        )
+        items = result.get("items", [])
+        if not items:
+            break
+        for it in items:
+            yield it
+        if len(items) < 100:
+            break
+        page += 1
+        time.sleep(0.25)
```
fressi-elastic: Why are you sleeping here?

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+def get_issue_comments(number: int) -> List[Dict[str, Any]]:
+    comments: List[Dict[str, Any]] = []
+    page = 1
+    while True:
+        repo = CONFIG.get("repo")
+        data = gh_request(f"/repos/{repo}/issues/{number}/comments", params={"per_page": "100", "page": str(page)})
+        if not data:
+            break
+        comments.extend(data)
+        if len(data) < 100:
```
fressi-elastic: Can you add a comment explaining we are handling this condition?

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+    except RuntimeError as re:
+        logger.error(f"Runtime error: {re}")
+        return 1
+    except Exception as e:
+        logger.error(f"Unhandled error: {e}")
+        return 1
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
fressi-elastic: I typically prefer using `sys.exit(1)` instead of raising exception for errors, and to allow main returning None for exiting with 0 value (success). To me looks much cleaner and make simpler testing the `main` in a unit test.

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+    # In dry-run, skip mutating requests (anything not GET) and just log.
+    if is_dry_run():
+        logger.debug(f"Would {method} {url} body={json.dumps(body) if body else '{}'}")
+        if method.upper() != "GET":
+            return {}
+    req = urllib.request.Request(url, data=data, method=method)
+    req.add_header("Authorization", f"Bearer {token}")
+    req.add_header("Accept", "application/vnd.github+json")
+    try:
+        with urllib.request.urlopen(req) as resp:
```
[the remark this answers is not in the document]
NickDris: Thanks for the suggestion! I considered using requests, but I'd like to keep this script dependency-free since it's not business-critical and urllib covers our needs for the moment. My aim is to keep it lightweight and easy to run in any Python environment, because we may need to reuse it in other repos.

If we find these kind of optimizations necessary down the line, we'll definitely revisit this. For now simply sticking to stdlib seems like a reasonable tradeoff.

.github/scripts/backport.py
```diff
@@ -0,0 +1,480 @@
...
+        raise RuntimeError("Missing or invalid GITHUB_REPOSITORY. Either set it or pass --repo (owner/repo)")
+
+
+def configure(args: argparse.Namespace) -> None:
+    """Populate CONFIG, initialize logging, and validate required inputs.
+
+    This centralizes setup so other entry points (tests, future subcommands)
+    can reuse consistent initialization semantics.
+    """
+    CONFIG["dry_run"] = bool(getattr(args, "dry_run", False))
```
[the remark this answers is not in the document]
NickDris: You're right, I got overly defensive by forgetting argparse already does apply the defaults under the hood. I can also use the parser's default arguments in order to make the parsing more straightforward. Thanks!

.github/scripts/backport.py
NickDris: I've done some reading here and understood. Didn't know I would have problems during the unit tests in the former way. Thanks!

NickDris:
@elasticmachine update branch

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+            default=7,
+            help="Days to look back (default: 7). Ignored in --pr-mode",
+        )
+        p_remind.add_argument(
+            "--pending-reminder-age-days",
+            type=int,
+            required=False,
+            default=7,
+            help="Days between reminders for the same PR (default: 7). Adds initial reminder if none posted yet.",
+        )
```
gbanasiak: 7 feels too short for the ongoing reminder, can we default to 14?

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+            raise RuntimeError("PRInfo initialization failed: missing number and labels")
+
+
+# ----------------------------- PR Extraction (single or bulk) -----------------------------
+def load_event() -> dict:
+    """Load the GitHub event payload from GITHUB_EVENT_PATH for single PR mode.
+
+    Returns an empty dict if the path is missing to allow callers to decide on fallback behavior.
+    """
+    path = os.environ.get("GITHUB_EVENT_PATH")
```
fressi-elastic: In general I do prefer raising an exception instead of returning an empty object. It makes debugging corner cases much simpler. Example:

```Python
def load_event() -> dict:
  path = os.environ.get("GITHUB_EVENT_PATH", "").strip()
  if not path:
       raise FileNotFound("GITHUB_EVENT_PATH environment variable is empty")
   if os.path.isfile(path):
       raise FileNotFound(f"file not found: {path)")
  with open(path, "r", encoding="utf-8") as f:
        data = json.load(f)
  if not isinstance(data, dict):
       raise TypeError(f"Event data is a {type(data)}, want a dict.")
  return data
```

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+
+def repo_needs_pending_label(repo_labels: list[str]) -> bool:
+    return repo_labels.count(PENDING_LABEL) == 0
+
+
+def ensure_backport_pending_label() -> None:
+    """If the exact PENDING_LABEL string does not appear at least once, we create it."""
+    repo = CONFIG.repo
+    try:
+        existing = gh_request(path=f"/repos/{repo}/labels", params={"per_page": "100"})
```
fressi-elastic: It would be nice to have (for readability purpose) a method that list repository labels:

```
def list_repository_labels():
  return [
     l
     for e in gh_request(path=f"/repos/{repo}/labels", params={"per_page": 100}
     if l := e.get("name")
  ]
```

It would also be nice to add the support for `per_page` in the `gh_request` method itself.

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+    except Exception as e:
+        LOG.info(f"Could not list labels; attempting direct create: {e}")
+        existing = []
+    LOG.debug(f"Repo has labels: {existing};")
+    names = [lbl.get("name", "") for lbl in existing]
+    if not repo_needs_pending_label(names):
+        return
+    try:
+        gh_request(method="POST", path=f"/repos/{repo}/labels", body={"name": PENDING_LABEL, "color": PENDING_LABEL_COLOR})
```
fressi-elastic: I think this should be wrapped by a method/function that says what this request is doing.
Example:

```
def add_repository_label(repository: str, name: str, color: str):
   gh_request(method="POST",  path=f"/repos/{repository}/labels", body={"name": name, "color": color})
```

Probably checking if the repo has already a label is not necessary. Just capture that exception if produces one error and return with no error. In case of any other error please raise or re-raise an error. It is an anti pattern on my opinion silencing errors and leave the code proceed unnotified in low level functions. It should the high level workflow function in charge of it.

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+
+def pr_needs_pending_label(info: PRInfo) -> bool:
+    pr_labels = info.labels
+    has_version_label = any(VERSION_LABEL_RE.match(label) for label in pr_labels)
+    has_pending = any(label == PENDING_LABEL for label in pr_labels)
+    has_backport = any(label == BACKPORT_LABEL for label in pr_labels)
+    return not has_version_label and not has_pending and not has_backport
+
+
+def add_label(pr_number: int, label: str) -> None:
```
fressi-elastic: I see you should have two methods which adds labels:
 
- ```def add_pull_request_label(number: int, label: str):```
- ```def add_repository_label(repository: str, label: str, color: str):```

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+        comments.extend(data)
+        # We are using a page size of 100. If we get less, we are done.
+        if len(data) < 100:
+            break
+        page += 1
+    return comments
+
+
+def post_comment(number: int, body: str) -> None:
+    if is_dry_run():
```
fressi-elastic: I think `is_dry_run` should be called from `gh_request` function.

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+def last_reminder_time(comments: list[dict[str, Any]], marker: str) -> dt.datetime | None:
+    def comment_ts(c: dict[str, Any]) -> dt.datetime:
+        ts_raw = c.get("created_at") or c.get("updated_at")
+        if not ts_raw:
+            raise RuntimeError("Comment missing timestamp fields")
+        return dt.datetime.strptime(ts_raw, ISO_FORMAT).replace(tzinfo=dt.timezone.utc)
+
+    for c in sorted(comments, key=comment_ts, reverse=True):
```
fressi-elastic: Comments should be objects with attributes and key should be a lamda extraction the attributes. Example:

```Python
   comments: list[IssueComment] = sorted(
       c for c in list_comments(issue_number) if marker in comments.text
       key=lambda c: c.created_at or c.updated_at or SOME_VERY_FAR_AWAY_TIMESTAMP
   )
   if not comments:
     raise()
   return comments[0]
```

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+    if pr_needs_pending_label(info):
+        return False
+    comments = get_issue_comments(info.number)
+    prev_time = last_reminder_time(comments, COMMENT_MARKER_BASE)
+    if prev_time is None:
+        return True
+    return prev_time < threshold
+
+
+def run_remind(prefetched_prs: list[dict[str, Any]], pending_reminder_age_days: int, lookback_days: int) -> int:
```
fressi-elastic: It would be simpler to work on one issue/pr per time, instead of one list of PRs as the logic for every PR is independent. The code would be much simpler and also testing it.

.ci/scripts/backport.py
```diff
@@ -0,0 +1,523 @@
...
+        LOG.error(f"Missing configuration key: {ke}")
+        sys.exit(1)
+    except NotImplementedError as nie:
+        LOG.error(f"Not implemented error: {nie}")
+        sys.exit(2)
+    except RuntimeError as re:
+        LOG.error(f"Runtime error: {re}")
+        sys.exit(3)
+    except Exception as e:
+        LOG.error(f"Unhandled error: {e}")
```
fressi-elastic: Same as above. Please don't capture an exception just to hide its cause. The stack trace is very important for debugging. Use `LOG.exception` or just don't capture it. There is no requirement to produce a specific return value for an unknown error. Python will return a non zero error for an error that is not captured for you.

fressi-elastic (approved):
I have the filing that this PR got lot of unnecessary complexity and technical debt just because we didn't reused pre-existing 3rd party code [1]. It is nice to write code to learn and get code reviewed to be able to exchange ideas and experiences. But for something like this I would have preferred using a pre-existing library. It would have been much less work than writing (and testing) this code. Given the quality of the code of this PR is not very critical to me it LGTM but please read my comments so you can do different the next time (I hope I am not been rude, it has never been my intention).

[1] https://github.com/PyGithub/PyGithub

PR #895, opened 2025-10-29
Use default logsdb message template sort

jordan-powers:
Use the mapping parameter added in elastic/elasticsearch#136571 to sort on `message.template_id` instead of manually specifying with `index.sort.fields`.

Relates to #825.


martijnvg (approved):
LGTM

PR #896, opened 2025-10-30
Solves problems with expensive inlinestats benchmarks

ncordon:
Adds a smaller index `nyc_taxis_sample` with a 1000 rows of the original `nyc_taxis` and runs the benchmarks `stats_count_group_by_esql` and `inlinestats_count_group_by_esql` on it.

## Why

Because there are some benchmarks for `inline stats` that are triggering circuit breaker exceptions:

```
esrally.exceptions.RallyError: Cannot run task [stats_count_group_by_esql]: Request returned an error. Error type: api, Description: {"error":{"root_cause":[{"type":"circuit_breaking_exception","reason":"[request] 

Data too large, data for [<reused_arrays>] would be [5153974035/4.8gb], which is larger than the limit of [5153960755/4.7gb]; 
for more information, see https://www.elastic.co/docs/troubleshoot/elasticsearch/circuit-breaker-errors?version=master","bytes_wanted":5153974035,"bytes_limit":5153960755,"durability":"TRANSIENT","suppressed":[{"type":"circuit_breaking_exception","reason":"[request] 

Data too large, data for [<reused_arrays>] would be [5153974035/4.8gb], which is larger than the limit of [5153960755/4.7gb];
```

### DIFF
--- a/nyc_taxis/challenges/default.json
+++ b/nyc_taxis/challenges/default.json
@@ -926,11 +926,16 @@
           "tags": ["setup"]
         },
         {
+          "name": "refresh-after-index",
+          "operation": "refresh",
+          "tags": ["setup"]
+        },
+        {
           "operation": "create-nyc-taxis-sample-index",
           "tags": ["setup"]
         },
         {
-          "name": "refresh-after-index",
+          "name": "refresh-after-sample-index",
           "operation": "refresh",
           "tags": ["setup"]
         },


nyc_taxis/operations/default.json
```diff
@@ -1019,32 +1044,32 @@
     {
       "name": "stats_count_group_by_esql",
       "operation-type": "esql",
-      "query" : "FROM nyc_taxis METADATA _id | stats count(passenger_count) by _id | LIMIT 1000"
+      "query" : "FROM nyc_taxis_sample METADATA _id | stats count(passenger_count) by _id"
```
ncordon: Note the `LIMIT 1000` was moved after the queries in this other pr: https://github.com/elastic/rally-tracks/pull/873

nyc_taxis/operations/default.json
```diff
@@ -956,6 +956,31 @@
...
+    {
+      "name": "create-nyc-taxis-sample-index",
+      "operation-type": "raw-request",
+      "index": "nyc_taxis",
+      "method": "POST",
+      "path": "/_reindex",
+      "body": {
+        "source": {
+          "index": "nyc_taxis",
+          "query": { "match_all": {} }
```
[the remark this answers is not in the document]
craigtaverner: I assumed you had two ways of doing this, one with a query, and include a size inside that, and the other with the `max_docs` parameter. Since you used max_docs, I presumed the query was unnecessary.

nyc_taxis/operations/default.json
craigtaverner: Having the limit before or after the stats completely changes the behaviour of the query. The other PR moved it after, and this one effectively moves it before again. I'm assuming we will see performance differences with both changes?

craigtaverner: Having said that, I understand the purpose of these queries was to limit before, so the change in this current PR is moving back towards that original purpose.

nyc_taxis/challenges/default.json
```diff
@@ -926,6 +930,10 @@
           "operation": "refresh",
           "tags": ["setup"]
         },
+        {
+          "operation": "create-nyc-taxis-sample-index",
+          "tags": ["setup"]
+        },
```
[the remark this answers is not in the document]
gbanasiak: Or perhaps we need 2 refreshes to make sure the reindexing search works fine?

PR #898, opened 2025-11-03
ES|QL: add accept_pragma_risks to JOINS track

luigidellaquila:
Pragmas are only allowed in snapshot builds by default.
With this change we can run the track also on non-snapshot.


craigtaverner (approved):
LGTM

PR #900, opened 2025-11-05
Add new corpora for so_vector with vectors as base64 strings

iverase:
Elasticsearch has recently introduced support for sending vectors as base64 encoded strings. This commit introduces a new corpora (data files have already been added to the GCP bucket) which contains the same so_vector data set but with vectors encoded as base64 strings.

In addition this commit introduces a new parameter called corpora that allows to switch between the two existing datasets.

### DIFF
--- a/so_vector/files.txt
+++ b/so_vector/files.txt
@@ -1,4 +1,4 @@
 documents.json
 documents-1k.json
 documents_base64.json
-documents_base62-1k.json
+documents_base64-1k.json


so_vector/files.txt
```diff
@@ -1,2 +1,4 @@
 documents.json
 documents-1k.json
+documents_base64.json
+documents_base62-1k.json
```
pmpailis: typo? :)

iverase: doh!

PR #902, opened 2025-11-07
Add a benchmark for an ESQL agg on LENGTH

nik9000:
In https://github.com/elastic/elasticsearch/pull/137382 we're pushing functions into field loading and using LENGTH as an example. This adds a rally track to demonstrate the performance difference:
```
| 90th | esql-avg-message-length | 11078 | 6670 | -4407.95 | ms | -39.79% |
```

### DIFF
--- a/http_logs/challenges/default.json
+++ b/http_logs/challenges/default.json
@@ -390,4 +390,109 @@
           }
         }
       ]
+    },
+    {
+      "name": "esql",
+      "description": "Some ESQL commands",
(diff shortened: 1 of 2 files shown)


nik9000:
I'm not proposing we turn this one right now. Just have it so we can run it when needed.

nik9000:
Perf folks - should I backport this? I'm fine with it only being available when running against `main`.

PR #907, opened 2025-11-11
fix: use iterations for the wikipedia esql-full-text-functions challenge 

mromaios:
Switching to `iterations` from `time-period` in the wikipedia esql-full-text-functions-challenge. Rally can stall if there are too many metrics and `term` related queries in combination with running the task for a longer duration reach and exceed this limit causing intermittent failures.

### DIFF
--- a/wikipedia/challenges/common/esql-full-text-functions.json
+++ b/wikipedia/challenges/common/esql-full-text-functions.json
@@ -37,28 +37,28 @@
   "operation": "query-esql-match-search",
   "clients": {{ standalone_search_clients | default(20) | int }},
   "warmup-iterations": {{ warmup_iterations | default(100) }},
-  "iterations": {{ iterations | default(500) }}
+  "iterations": {{ iterations | default(1000) }}
 },
 {
   "name": "standalone-query-match-search",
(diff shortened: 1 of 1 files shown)


kderusso (commented):
Thanks for this change @mromaios ! 

For my education I'd like to understand a bit more: 

- If the issue is exceeding the established time period, why change to iterations vs. increasing the timeout to some huge number? 
- Are there any expected changes in existing reports that we need to be aware of? Would we expect current test output to significantly change?

mromaios:
Thanks @kderusso 
> If the issue is exceeding the established time period, why change to iterations vs. increasing the timeout to some huge number?

AFAIU it's not exceeding the established time period, but the metrics/events produced during this time period are too many to process, so it stalls. Changing this to iterations effectively reduces the events. 

> Are there any expected changes in existing reports that we need to be aware of? Would we expect current test output to significantly change?

That was my concern as well. We set the iterations to 500 which should give us a good representation (other benchmarks use less for example) and keep output on a similar level. Perhaps pushing iterations to 1000 (which is used by big5) might be even better. I'll make a run with that and compare to the current metrics so that we get an idea.

PR #909, opened 2025-11-13
fix: so_vector wait for merges after indexing

mromaios:
Adding a wait so that potential merges from indexing as well as merges after a forcemerge have finished


carlosdelest (approved):
Thanks @mromaios !

PR #920, opened 2025-11-17
Streams: Respect number_of_shards setting

flash1293:
When creating these, I forgot to make sure number_of_shards is respected. This PR is fixing this.

### DIFF
--- a/elastic/logs/templates/composable/logs.child1@stream.json
+++ b/elastic/logs/templates/composable/logs.child1@stream.json
@@ -5,9 +5,7 @@
     "template": {
       "settings": {
         "index": {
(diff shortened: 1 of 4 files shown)


inqueue (changes requested):
Hi @flash1293, component template `track-custom-shared-settings` is intended for this purpose:

```json
"composed_of": ["track-custom-shared-settings"]
```

Can it be utilized in the stream templates?

flash1293:
Thanks @inqueue , that's probably the right way I just didn't know of. So you are saying the `track-custom-shared-settings` component template is just always there and configured correctly?

Where can I see what else it's setting?

inqueue (approved):
LGTM, thanks!

PR #922, opened 2025-11-19
Exclude some challenges when testing with ES release builds

gbanasiak:
This unblocks ES release builds testing used in `9.2` branch and onwards.

Applies to:
* `elastic/logs` track, `logging-insist-chicken` challenge due to `INSIST_🐔` ESQL function
* `wikipedia` track, `esql-full-text-functions` challenge due to `TERM` ESQL function


NickDris (approved):
LGTM, I left a non-blocking comment

PR #935, opened 2025-11-24
so_vector - Use match with boost zero for filtering knn

carlosdelest:
so_vector uses ESQL term query for doing knn filtering.

However, the term query will be scored besides performing the filter, so it will both be less performant and impact the qiuery score.

This PR replaces the term query (which will be deleted soon) for a match query with a boost of zero to avoid the filter being scored.

### DIFF
--- a/so_vector/operations/default.json
+++ b/so_vector/operations/default.json
@@ -52,7 +52,7 @@
   "param-source": "esql-knn-param-source",
   "k": 10,
   "num_candidates": 50,
-  "filter": "term(tags, \"java\")"
+  "filter": "match(tags, \"java\", {\"boost\": 0})"
 },
 {
   "name": "knn-recall-10-50-java",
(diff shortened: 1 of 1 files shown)


so_vector/operations/default.json
```diff
@@ -521,7 +521,7 @@
   "param-source": "esql-knn-param-source",
   "exact": true,
   "k": 10,
-  "filter": "term(tags, \"java\")"
+  "filter": "match(tags, \"java\")"
```
mromaios: 💭 🔧  
```suggestion
  "filter": "match(tags, \"java\"), {\"boost\": 0})"
```
Do we also need it here?

carlosdelest: No, on exact search we're not scoring (there's no METADATA _score), so the filter is applied without scoring

PR #970, opened 2025-12-16
add data stream lifecycle put operation for elastic logs track

seanzatzdev:
This PR adds the data lifecycle management put operation to the elastic logs track. This will be used in the DLM Tiers project benchmarking.

Contributes to https://github.com/elastic/elasticsearch-team/issues/2131

### DIFF
--- a/elastic/shared/runners/dlm.py
+++ b/elastic/shared/runners/dlm.py
@@ -33,6 +33,7 @@ async def put_lifecycle(es, params):
     - data-stream: Name of the data stream (supports wildcards)
     - data_retention: Optional retention period (e.g., "7d", "30d")
     - enabled: Optional boolean to enable/disable lifecycle (default: true)
+    - downsampling: Optional dict for downsampling configuration
 
     Returns: Number of data streams configured
     """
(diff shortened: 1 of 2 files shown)


elastic/shared/runners/dlm.py
```diff
@@ -0,0 +1,75 @@
...
+
+    if "enabled" in params:
+        lifecycle_config["enabled"] = params["enabled"]
+
+    if "downsampling" in params:
+        lifecycle_config["downsampling"] = params["downsampling"]
+
+    logger.info(f"Configuring lifecycle on data stream [{data_stream}]: {lifecycle_config}")
+
+    # Get all data streams matching the pattern
```
nielsbauman: If we do add this API, we don't need to resolve the data streams ourselves as the API already supports comma-separated data stream names (including wildcards): https://github.com/elastic/elasticsearch/blob/4e91c46bbf8333e76051919b455dca25ba9ec408/server/src/main/java/org/elasticsearch/action/datastreams/lifecycle/PutDataStreamLifecycleAction.java#L40-L40

PR #984, opened 2026-01-09
Add wikipedia track ESQL profiling

carlosdelest:
Adds ES|QL profiling metrics to the ES|QL queries in the wikipedia track.

Related https://github.com/elastic/rally-tracks/pull/973, https://github.com/elastic/elasticsearch/pull/139540, https://github.com/elastic/elasticsearch/pull/138564

### DIFF
--- a/wikipedia/track.py
+++ b/wikipedia/track.py
@@ -338,45 +338,45 @@ class EsqlProfileRunner(runner.Runner):
                     if took_nanos > 0:
                         result[f"{phase_name}.took_ms"] = took_nanos / 1_000_000  # Convert to milliseconds
 
-        # Extract driver-level metrics
-        drivers = profile.get("drivers", [])
-        for driver in drivers:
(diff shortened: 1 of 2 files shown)


wikipedia/track.py
```diff
@@ -283,6 +284,106 @@ def params(self):
...
+        if not bool(headers):
+            headers = None
+
+        # Execute the ESQL query with profiling
+        response = await es.perform_request(method="POST", path="/_query", headers=headers, body=body, params=request_params)
+        profile = response["profile"]
+
+        # Build took_ms entries for each profiled phase
+        result = {}
+        if profile:
```
ioanatia: we have a check here to see that `profile` is present, but we don't have one after this if statement when we iterate over the drivers and plans.
I wonder if profile can ever be empty - and if it can - wouldn't it be better to just return early, e.g:
```
result = {}
if not profile:
  return result
```

PR #1108, opened 2026-04-01
Add knn recall measurement to random_vector track

ah89:
Adds recall computation for each partition tier (small, medium, large). Each recall phase runs an approximate kNN search and an exact script_score search against the same partition, then reports avg/min/max recall@k.

Relates to #1093


random_vector/operations/default.json
```diff
@@ -75,5 +75,47 @@
     {%- else %}
   "field": "emb"
     {%- endif %}
+},
+{
+  "name": "knn-recall-small-partition",
+  "operation-type": "knn-recall",
+  "param-source": "knn-recall-param-source",
+  "request-timeout": 600,
+  "include-in-reporting": false,
```
davidkyle: Surely  `"include-in-reporting": true`

random_vector/challenges/default.json
```diff
@@ -64,20 +64,41 @@
       "warmup-iterations": 500,
       "iterations": {{ search_iterations | default(10000) | int }},
       "clients": {{ search_clients | default(8) | int }}
+    },
+    {
+      "name": "small-partition-recall",
+      "operation": "knn-recall-small-partition",
+      "warmup-iterations": 0,
+      "iterations": 1,
+      "clients": 1
```
davidkyle: Are these settings used in the knn-recall task?

random_vector/track.py
```diff
@@ -192,6 +192,101 @@ def params(self):
...
+        self._tier_partition_count = count
+        self.infinite = True
+
+    def partition(self, partition_index, total_partitions):
+        return self
+
+    def params(self):
+        import numpy as np
+
+        query_vectors = [np.random.rand(self._dims).tolist() for _ in range(self._recall_iterations)]
```
davidkyle: Would a better name for `recall_iterations` be `number_of_queries`

random_vector/track.py
```diff
@@ -192,6 +192,101 @@ def params(self):
...
+            knn_body["_source"] = False
+            knn_result = await es.search(body=knn_body, index=index, size=k)
+            knn_hits = [hit["_id"] for hit in knn_result["hits"]["hits"]]
+
+            exact_body = {
+                "query": {
+                    "script_score": {
+                        "query": {"term": {"partition_id": partition_id}},
+                        "script": {
+                            "source": f"cosineSimilarity(params.query, '{field}') + 1.0",
```
davidkyle: I didn't know you can do a knn query and a exact query on the same field. Is this because the field is mapped as `bbq_flat`

benwtrent (commented):
I have been off and on thinking about this. And I know I said its something we might want to do...but, I just don't know.

It really seems like the main focus here should be performance with random vectors over flat & diskbbq with some settings. Now, which settings provide which recall, that honestly isn't that useful here as the vectors are random :/

What do you think @ah89 ? The more I think about it, the more I think recall is just the wrong kind of measurement here, and we should instead provide DiskBBQ perf with a given visit percentage (or percentages) provided by the caller?

ah89:
> I have been off and on thinking about this. And I know I said its something we might want to do...but, I just don't know.
> 
> It really seems like the main focus here should be performance with random vectors over flat & diskbbq with some settings. Now, which settings provide which recall, that honestly isn't that useful here as the vectors are random :/
> 
> What do you think @ah89 ? The more I think about it, the more I think recall is just the wrong kind of measurement here, and we should instead provide DiskBBQ perf with a given visit percentage (or percentages) provided by the caller?

@benwtrent I think you point is valid, I thought we could use this as a lower limit for other purposes otherwise with random vectors there's no semantic structure, so recall numbers are meaningless — they'd just reflect algorithmic mechanics, not real-world accuracy.

random_vector/challenges/default.json
ah89: Yes, those are standard Rally schedule parameters — Rally's task executor applies warmup-iterations, iterations, and clients to every task before invoking the runner. They're not specific to the recall operation. The values (0 warmup, 1 iteration, 1 client) make sense since recall is a correctness check, not a throughput benchmark. I might change the PR as @benwtrent [comment](https://github.com/elastic/rally-tracks/pull/1108#pullrequestreview-4084608644) so probably I send changed settings for visit percentage.

random_vector/operations/default.json
benwtrent: I wonder actually, reporting would indicate qps, latency, etc. which we don't care about. We only care about the recall value mean, which I am not sure will be provided in the output reporting.

random_vector/track.py
benwtrent: you should always be able to do a script query against a dense_vector field. It loads the float vectors via doc-values and executes the desired script.

benwtrent:
@ah89 for now, I think let's not do this. Sorry for waffling on it. The more I think, the more it seems that this just makes the benchmark very complex with very little benefit.

ah89:
This issue does not support our benchmarking goals.

PR #1136, opened 2026-05-04
[msmarco-v2-vector] Add recall ground truth support for scaling benchmark subsets

ChrisHegarty:
[DRAFT] I'm working on a scaling benchmark that demonstrates horizontal scale-out of vector search using the msmarco-v2-vector track. The benchmark doubles data and nodes at each tier: 18M/1 node, 36M/2 nodes, 72M/4 nodes, and full 138M/8 nodes.

To measure recall accurately at each tier, we need ground truth files generated against the exact document subset that's indexed - the true nearest neighbors change depending on which documents are in the index. This PR adds the tooling and track support needed for that.

## Changes
**Ground truth generation** (`_tools/generate_ground_truth.py`):
   - Computes brute-force exact nearest neighbors using numpy against the
      Hugging Face dataset, matching the ES script_score formula
   - Outputs JSONL in the same format as the existing `queries-recall.json.bz2`
   - Uses float64 for dot-product computation to match Java double precision
   - Validated against existing 10m and full ground truth files (99.97%+ overlap,
      differences only at score-tie boundaries)

**Ground truth comparison** (`_tools/compare_ground_truth.py`):
   - Utility to compare two ground truth files and report overlap, score diffs,
      and per-query mismatches for validation

**Track changes** (`track.py`, `operations/default.json`):
  - Added recall file selection for 18m, 36m, and 72m doc counts
  - The `recall-doc-set` parameter is automatically set based on
      `initial_indexing_ingest_doc_count`


john-wagster (approved):
this all lgtm; no major concerns.

PR #1147, opened 2026-05-22
[MS MARCO] Add explicit 10M docs corpora

davidkyle:
The 10M vector benchmarks use the `initial_indexing_ingest_doc_count` setting to limit the corpora to 10M. When evaluating recall on this dataset we found the score to be lower than expected, even increasing the visit percentage to 50% did not provide any improvement. 

The suspicion is that because rally uses multiple clients to index the documents there is a race between the clients and it might be different vectors uploaded depending on how each client progresses. This change adds a `msmarco-v2_base64-10-million` corpora consisting of the first 3 chunks of Cohere vector plus a new chunk I created containing 10M vectors. These vectors are fixed and will not differ making the recall measure more reliable.

I've tested with esbench against this PR branch using these params 

```
{
  "cloud.vendor": "aws",
  "cloud.region": "eu-west-1",
  "serverless.environment": "qa",
  "serverless.project_id": "XXXX",
  "track.name": "msmarco-v2-vector",
  "track.repository.url": "https://github.com/davidkyle/rally-tracks",    <--- my repo
  "track.revision": "10M-msmarco",      <-- This PR branch
  "track.challenge": "index-and-search",  
  "track.params": {
    "corpora": ["msmarco-v2_base64-10-million"],  <-- the important part
    "mapping_type": "vectors-only",
    "vector_index_type": "bbq_disk",
    "include_initial_indexing": true,
    "initial_ingest_clients": 8,
    "include_recall": true,
    "include_parallel_indexing": false,
    "search_ops": [[10, 100, 5, 3]],
    "standalone_search_warmup_iterations": 1000,
    "standalone_search_iterations": 1000  
  }
}
```

Closes https://github.com/elastic/vector-db-program/issues/124


john-wagster (approved):
baring validation from running the new queries file and seeing recall numbers; this lgtm

davidkyle:
Here's the results using the above config and the new data files
```
{
  "best_ndcg_10": 0.2340532429292017,
  "ndcg_10": 0.22874840819541162,
  "avg_recall": 0.9184210526315789,
  "min_recall": 4,
  "k": 10,
  "num_candidates": 100,
  "visit_percentage": 3,
  "avg_nodes_visited": 647303.4736842106,
  "99th_percentile_nodes_visited": 651712
}
```

PR #1170, opened 2026-06-01
msmarco-v2-vector: use base64 corpus for parallel updates

mayya-sharipova:
Ingestion uses base64 encoding, so use the base64 variant of the parallel-indexing corpus for compatibility.


mayya-sharipova:
Copilot is write the default for indexing is to use float not base64.
I was wrong, closing this PR

PR #1178, opened 2026-06-04
[MS MARCO] Fix autoscale schedule templates for multi-phase configurations

reugn:
All `as_*` parameters in the three autoscale schedule templates (search-autoscale, ingest-autoscale, ingest-search-autoscale) now use modulo indexing so shorter arrays repeat automatically across phases. A single-element array applies that value to every phase.

`warmup-time-period` values are clamped to a minimum of 1 to prevent Rally validation errors when users pass 0 to mean "no warmup".

`ingest-autoscale-schedule.json` loop driver was also aligned to `as_warmup_time_periods` for consistency with the other two templates.

Tests were added covering all three templates and README updated with shared autoscale parameter conventions.

### DIFF
--- a/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
+++ b/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
@@ -23,7 +23,13 @@
     "ingest-percentage": {{initial_ingest_percentage | default(100) | int}}
   }
 }
-{%- set p_as_phases = (as_phases | default(5))%}
+{%- if as_phases is defined %}
+{%- set p_as_phases = as_phases %}
+{%- elif as_warmup_time_periods is defined %}
+{%- set p_as_phases = as_warmup_time_periods|length %}
+{%- else %}
+{%- set p_as_phases = 5 %}
+{%- endif %}
 {%- set p_as_warmup_time_periods = (as_warmup_time_periods | default([600]))%}
 {%- set p_as_ingest_target_throughputs = (as_ingest_target_throughputs | default([-1]))%}
 {%- set p_as_time_periods = (as_time_periods | default([1800]))%}
--- a/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
+++ b/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
@@ -23,12 +23,13 @@
     "ingest-percentage": {{initial_ingest_percentage | default(100) | int}}
   }
 }
-{%- set p_as_warmup_time_periods = (as_warmup_time_periods | default([600,600,600,600,600]))%}
-{%- set p_as_ingest_target_throughputs = (as_ingest_target_throughputs | default([-1,-1,-1,-1,-1]))%}
-{%- set p_as_time_periods = (as_time_periods | default([1800,1800,1800,1800,1800]))%}
+{%- set p_as_phases = (as_phases | default(5))%}
+{%- set p_as_warmup_time_periods = (as_warmup_time_periods | default([600]))%}
+{%- set p_as_ingest_target_throughputs = (as_ingest_target_throughputs | default([-1]))%}
+{%- set p_as_time_periods = (as_time_periods | default([1800]))%}
 {%- set p_as_ingest_clients = (as_ingest_clients | default([1,2,4,8,16]))%}
 {%- set p_ingest_bulk_size = (ingest_bulk_size | default(100))%}
-{%- for i in range(p_as_warmup_time_periods|length) %},
+{%- for i in range(p_as_phases) %},
   {%- if p_as_ingest_target_throughputs[i % (p_as_ingest_target_throughputs|length)] < 0 %}
 {
   "name": "ingest-{{loop.index}}-c{{p_as_ingest_clients[i % (p_as_ingest_clients|length)]}}-b{{p_ingest_bulk_size}}",
--- a/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
+++ b/msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
@@ -28,31 +28,31 @@
 {%- set p_as_time_periods = (as_time_periods | default([1800,1800,1800,1800,1800]))%}
 {%- set p_as_ingest_clients = (as_ingest_clients | default([1,2,4,8,16]))%}
 {%- set p_ingest_bulk_size = (ingest_bulk_size | default(100))%}
-{%- for i in range(p_as_ingest_clients|length) %},
-  {%- if p_as_ingest_target_throughputs[i] < 0 %}
+{%- for i in range(p_as_warmup_time_periods|length) %},
+  {%- if p_as_ingest_target_throughputs[i % (p_as_ingest_target_throughputs|length)] < 0 %}
 {
-  "name": "ingest-{{loop.index}}-c{{p_as_ingest_clients[i]}}-b{{p_ingest_bulk_size}}",
-  "clients": {{p_as_ingest_clients[i]}},
+  "name": "ingest-{{loop.index}}-c{{p_as_ingest_clients[i % (p_as_ingest_clients|length)]}}-b{{p_ingest_bulk_size}}",
+  "clients": {{p_as_ingest_clients[i % (p_as_ingest_clients|length)]}},
   "operation": {
     "operation-type": "bulk",
     "bulk-size": {{p_ingest_bulk_size}},
     "looped": true
   },
-  "warmup-time-period": {{p_as_warmup_time_periods[i]}},
-  "time-period": {{p_as_time_periods[i]}}
+  "warmup-time-period": {{[p_as_warmup_time_periods[i % (p_as_warmup_time_periods|length)], 1]|max}},
+  "time-period": {{p_as_time_periods[i % (p_as_time_periods|length)]}}
 },
   {%- else %}
 {
-  "name": "ingest-{{loop.index}}-c{{p_as_ingest_clients[i]}}-b{{p_ingest_bulk_size}}",
-  "clients": {{p_as_ingest_clients[i]}},
+  "name": "ingest-{{loop.index}}-c{{p_as_ingest_clients[i % (p_as_ingest_clients|length)]}}-b{{p_ingest_bulk_size}}",
+  "clients": {{p_as_ingest_clients[i % (p_as_ingest_clients|length)]}},
   "operation": {
     "operation-type": "bulk",
     "bulk-size": {{p_ingest_bulk_size}},
     "looped": true
   },
-  "warmup-time-period": {{p_as_warmup_time_periods[i]}},
-  "time-period": {{p_as_time_periods[i]}},
-  "target-throughput": {{p_as_ingest_target_throughputs[i]}}
+  "warmup-time-period": {{[p_as_warmup_time_periods[i % (p_as_warmup_time_periods|length)], 1]|max}},
+  "time-period": {{p_as_time_periods[i % (p_as_time_periods|length)]}},
+  "target-throughput": {{p_as_ingest_target_throughputs[i % (p_as_ingest_target_throughputs|length)]}}
 },
   {%- endif %}
 {
--- a/msmarco-v2-vector/tests/test_autoscale_schedule.py
+++ b/msmarco-v2-vector/tests/test_autoscale_schedule.py
@@ -19,7 +19,6 @@ import json
 import pathlib
 
 import jinja2
-import pytest
 
 COMMON_DIR = pathlib.Path(__file__).parents[1] / "challenges" / "common"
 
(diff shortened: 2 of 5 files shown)


msmarco-v2-vector/tests/test_autoscale_schedule.py
```diff
@@ -0,0 +1,289 @@
...
+import json
+import pathlib
+
+import jinja2
+import pytest
+
+COMMON_DIR = pathlib.Path(__file__).parents[1] / "challenges" / "common"
+
+
+def render(template_name, params):
```
davidkyle, line 26: Nice, TIL this is testable

davidkyle (commented):
Thanks @reugn I see that this change makes the configuration less brittle. It has be problematic in the past becuase if you mis size one of the arrays you get an obscure error late in process. 

I'm slightly concerned that this change would silently swallow the configuration errors where 1 array is the wrong size and that may not be what the user intended. I'm not saying today's situation is any better, please could you give an example of the kind of problem you want to solve.

reugn:
Hi @davidkyle. To run a simple 4-phase search benchmark I had to write the following [override](https://github.com/elastic/elasticsearch-benchmarks/pull/3236):
```json
{
  "include_initial_indexing": false,
  "as_time_periods": [3600, 3600, 3600, 3600],
  "as_warmup_time_periods": [30, 1, 1, 1],
  "as_search_clients": [100, 100, 100, 100],
  "as_search_target_throughputs": [500, 2500, 500, 2500],
  "as_ingest_clients": [1, 1, 1, 1],
  "as_ingest_target_throughputs": [1, 1, 1, 1]
}
```
In the majority of cases the same value repeated multiple times. Another advantage is that if the default value suits all phases, the parameter doesn't need to be specified at all - the single default repeats automatically.

Also, passing 0 for no warmup caused a validation error, which wasn't obvious.

davidkyle:
@john-wagster you've suffered making changes to this benchmark before. WDTY of the proposal?

john-wagster:
@reugn I also think that a user of this might accidentally leave a value off of one of the "as*" variables and accidentally reuse the first value in an array instead of getting an error with this approach.  

However, I definitely agree there's pain here and this fix is in the right direction.   

I struggle with aligning around `as_warmup_time_periods` maybe we should introduce a variable that's explicit about the number of iterations something like `as_iterations` or `as_trials` to run and then it avoids the ambiguity.  For backwards compatibility make the default the same length as the `as_warmup_time_periods`, still aligning around this as the default.  Thoughts on this?  None of the "as*" variables seem like explicit drivers that are required to me.

Instead of supporting repeating all values then is it reasonable also to be a little less flexible and support 1 or `as_trials` values.  Where 1 value implies repeated `as_trials` times.  I'm not sure I know a use case for module indexing and repeating values through the array itself but could be convinced otherwise.

reugn:
@john-wagster I see your concern about accidental value skipping when array lengths don't match the phase count. There is currently no clean way to validate track parameters in Rally before a run starts, so I opened a proposal to add a proper validation hook: https://github.com/elastic/rally/issues/2143. In the meantime I applied the following changes to the PR:

- Added `as_phases` (default: 5) as an explicit parameter to control the number of phases, replacing the implicit derivation from `as_warmup_time_periods length` which was non-obvious

- Simplified all repeating-value array defaults to a single element, for example `as_time_periods: [1800]` instead of `[1800, 1800, 1800, 1800, 1800]`, since a single-element array already repeats across all phases via modulo

- `as_warmup_time_periods` default is now `[600]` and behaves like any other array rather than being the special driver for phase count

davidkyle:
@reugn https://github.com/elastic/rally/issues/2143 is a great idea I suggest making that change rather than iterating here. The validation is what will be the most beneficial to users. I also have a concern that changing the params in this PR might break existing benchmarks (public or private).

reugn:
@davidkyle I already opened a PR on the Rally side to implement the validation hook. I'd like to keep the parameter simplifications in this PR since they directly address repeating the same value across all array positions. To address the breaking change concern I'll make `as_phases` fall back to the length of `as_warmup_time_periods` when not explicitly set, to avoid breaking any existing config. Also, I plan to add support for per-phase ES settings in a follow-up PR.

msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
```diff
@@ -23,36 +23,37 @@
     "ingest-percentage": {{initial_ingest_percentage | default(100) | int}}
   }
 }
-{%- set p_as_warmup_time_periods = (as_warmup_time_periods | default([600,600,600,600,600]))%}
-{%- set p_as_ingest_target_throughputs = (as_ingest_target_throughputs | default([-1,-1,-1,-1,-1]))%}
-{%- set p_as_time_periods = (as_time_periods | default([1800,1800,1800,1800,1800]))%}
+{%- set p_as_phases = (as_phases | default(5))%}
```
john-wagster: This is on the track imo.  But I think defaulting to 5 here breaks backwards compatibility for instance with an existing configuration like this where the expectationg was 1 'phase' and not 5: 

```
    track.name: msmarco-v2-vector
    track.challenge: search-autoscale
    track.params:
      corpora: [msmarco-v2_base64-initial-indexing-1]
      mapping_type: vectors-only
      vector_index_type: bbq_disk
      initial_ingest_clients: 16
      initial_indexing_ingest_percentage: 55.55
      as_warmup_time_periods: [1200]
      as_time_periods: [1800]
      as_search_clients: [1000]
      as_search_target_throughputs: [1000]
      as_ingest_clients: [1]
```

I think defaulting to `as_warmup_time_periods | length` would work here to fix that since it was what we relied on previously to dictate.  

And I think it's reasonable still to assume that `as_ingest_clients` and `as_warmup_time_periods` are of the same length in any existing configurations.

davidkyle: > I think defaulting to as_warmup_time_periods | length would work here to fix that since it was what we relied on previously to dictate.

If this is possible then LGTM. I'll note that you could set `as_phases` to something really high like 20 and end up running the same benchmarks again and again but that's no worse than we have today and this change feels like progress in the right direction.

davidkyle:
@reugn I'll wait for perf team to comment on your rally PR before reviewing https://github.com/elastic/rally/pull/2144

msmarco-v2-vector/challenges/common/ingest-autoscale-schedule.json
reugn: > defaulting to 5 here breaks backwards compatibility

Made `as_phases` fall back to `len(as_warmup_time_periods)` when only `as_warmup_time_periods` is provided, so existing configurations are unaffected. The default of 5 is consistent with the original behavior (`as_warmup_time_periods` previously defaulted to 5 elements). Added backward compatibility tests.

john-wagster: I like it

john-wagster:
@reugn i think this can go in independent of the parameter validation change to rally itself which also seems very valuable to me if you remove the modulo operations from the arrays, effectively only supporting 1 or exact match.  So something like this:


```
{%- for i in range(p_as_phases) %}
  "warmup-time-period": {{[p_as_warmup_time_periods[0] if p_as_warmup_time_periods|length == 1 else p_as_warmup_time_periods[i], 1]|max}},
  "time-period": {{p_as_time_periods[0] if p_as_time_periods|length == 1 else p_as_time_periods[i]}},
  "clients": {{p_as_ingest_clients[0] if p_as_ingest_clients|length == 1 else p_as_ingest_clients[i]}},
  "target-throughput": {{p_as_ingest_target_throughputs[0] if p_as_ingest_target_throughputs|length == 1 else p_as_ingest_target_throughputs[i]}}
{%- endfor %}
```

thoughts?  could always come back in a subsequent PR and argue for supporting the module-based operations.

reugn:
> only supporting 1 or exact match

Rally's out-of-bounds list access silently returns an empty string. `p_as_warmup_time_periods[i]` with a mismatched array produces invalid JSON and surfaces as a JSONDecodeError. It also doesn't handle the case where an array is longer than `as_phases` - extra elements are silently ignored. Modulo avoids this by always producing valid output. Proper validation will be possible once the Rally PR is merged.

davidkyle:
> Rally's out-of-bounds list access silently returns an empty string.

😞

davidkyle (approved):
LGTM thanks for iterating on this

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

esbenchmachine:
@reugn
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

PR #1186, opened 2026-06-12
Add msmarco-jina-10m-vector track for Jina v5 10M vector benchmarking

wangch079:
## Summary

Adds a new Rally track for benchmarking **10M MSMARCO v2 passage embeddings** (Jina v5-small, 1024-dim, L2-normalized) with pure kNN search and vector recall.

- **Dataset**: 10M docs (`docid` + `emb` only)
- **Index types**: configurable `vector_index_type` (`hnsw`, `int8_hnsw`, `int4_hnsw`, `bbq_hnsw`, `bbq_disk`, …) and `vector_index_element_type` (`float`, `bfloat16`, …) via `--track-params` (defaults: `bbq_hnsw` / `float`)
- **Filter-free**: kNN-only search — no metadata filters, hybrid, or BM25
- **Pre/post force-merge**: `index-and-search` runs the same search + recall matrix before and after force-merge (multi-segment vs single-segment comparison on hosted Elasticsearch)
- **Configurable kNN params**: `search_ops` supports explicit `[k, num_candidates]` pairs and JSON `null` to omit either value so Elasticsearch applies defaults (`k = size`, `num_candidates = 1.5 × k`); recall reports `effective_k`
- **Recall**: vector top-100 ground truth per query (no human qrels / NDCG)
- **Serverless**: force-merge skipped by default; ingest/search/recall still supported with documented limitations


msmarco-jina-10m-vector/track.py
```diff
@@ -0,0 +1,192 @@
...
+        self._seed = params.get("seed")
+
+        cwd = os.path.dirname(__file__)
+        with bz2.open(os.path.join(cwd, QUERIES_FILENAME), "r") as queries_file:
+            for vector_query in queries_file:
+                self._queries.append(json.loads(vector_query))
+        random.Random(self._seed).shuffle(self._queries)
+
+        self._iters = 0
+        self._maxIters = len(self._queries)
```
davidkyle:
```suggestion
        self._max_iters = len(self._queries)
```

msmarco-jina-10m-vector/track.py
```diff
@@ -0,0 +1,192 @@
...
+        return {
+            "index": self._index_name,
+            "cache": self._params.get("cache", False),
+            "k": self._params.get("k"),
+            "num_candidates": self._params.get("num-candidates"),
+            "oversample_rescore": self._params.get("oversample-rescore", -1),
+        }
+
+
+class KnnRecallRunner:
```
davidkyle, line 124: Not related to this PR, just thinking out loud. I wonder if this class could be moved the rally code base as there are 5 copies of it in this repo

davidkyle (approved):
LGTM

I recently learned there is a way to test the parameter replacement in the rendered track files see https://github.com/elastic/rally-tracks/blob/master/msmarco-v2-vector/tests/test_autoscale_schedule.py. It might be worth added a test for this PR

msmarco-jina-10m-vector/track.py
wangch079: Good catch, will rename to `_max_iters`

wangch079:
> I recently learned there is a way to test the parameter replacement in the rendered track files see https://github.com/elastic/rally-tracks/blob/master/msmarco-v2-vector/tests/test_autoscale_schedule.py. It might be worth added a test for this PR

Good call — I’ll add a small unit test for this new track

msmarco-jina-10m-vector/track.py
wangch079: Fair point — there are several near-copies of `KnnRecallRunner` in rally-tracks.

Today each track is mostly self-contained. The one shared Python pattern in this repo is under `elastic/` (e.g. `elastic/shared/` used by `elastic/logs`, `elastic/security`, etc.).

If we want to dedupe, I’d probably start by grouping vector tracks under something like a `vector/` tree (similar to `elastic/`), then extract a shared helper there — rather than moving this into Rally core in one step. That’s a larger refactor than this PR, though.

Happy to open a follow-up issue if the team wants to explore that.

davidkyle: I created an issue in the Rally repo as I thought that would be a good place for the KnnRecallRunner but equally a shared class in this repo is valid. 

https://github.com/elastic/rally/issues/2151

davidkyle (approved):
Thanks for adding the tests

msmarco-jina-10m-vector/_tests/test_rendered_track.py
```diff
@@ -0,0 +1,106 @@
+# Licensed to Elasticsearch B.V. under one or more contributor
```
davidkyle, line 1: I've just found out that the `tests` directory should be named `_tests` so Rally will ignore it when loading the track. 

I discovered this because this line brings in the pytest dependency which has broken the track

wangch079: Good catch. Will update

esbenchmachine:
@wangch079
A backport is pending for this PR.
Apply all the labels that correspond to Elasticsearch minor versions expected to work with this PR, but select only from the available ones.
If intended for future releases, apply label for next minor

If the PR is only applicable for serverless, add the `vServerless` label.

When a `vX.Y` label is added, a new pull request will be automatically created, unless merge conflicts are detected or if the label supplied points to the next Elasticsearch minor version. If successful, a link to the newly opened backport PR will be provided in a comment.

In case of merge conflicts during backporting, create the backport PR manually following the steps from [README](https://github.com/elastic/rally-tracks?tab=readme-ov-file#merge-conflicts):
**Final steps to complete the backporting process:**
   2. Ensure each backport pull request is labeled with `backport`.
   3. Review and merge each backport pull request into the appropriate version branch.
   4. Remove `backport pending` label from this PR once all backport PRs are merged. Note: a version label must be added, otherwise the bot will add back the ´backport pending´ label. 

Thank you!

PR #1181, opened 2026-06-04
Add `index.mode` and `index.mapping.source.mode` configuration to big5

dakrone:
This allows configuring the index and mapping source mode in the big5 configuration. These default to "standard" and unset respectively.

PR #1193, opened 2026-06-26
Backport #1181 to 9.4: Add `index.mode` and `index.mapping.source.mode` configuration to big5

parkertimmins:
Backport of #1181 to the `9.4` branch.

## Summary
- Adds `index_mode` and `mapping_source_mode` track parameters to the big5 track for ES 9.x benchmarks
- Without this, passing `index_mode` as a track param against a 9.x cluster fails with `TrackConfigError: Unused track parameters ['index_mode']` because Rally selects the `9.4` branch (not `master`) based on the ES version

## Conflict resolution
The original commit added lines after `query_clients` in README.md, but `query_clients` wasn't yet on `9.4`. Resolved by including the `query_clients` line as well (it's on master but missing from this branch).


parkertimmins:
I'm not actually sure yet if this is needed :thinking:

parkertimmins:
This is not needed. Nightlies already force usage of master rather than 9.4

PR #1246, opened 2026-08-18
fix(geonames): use scroll-search for scroll operation

### LINKED ISSUE #624: Update geonames to use scroll-search instead of 'search'
`2024-06-27 12:16:26,998 ActorAddr-(T|:36351)/PID:8673 esrally.driver.runner WARNING Invoking a scroll search with the 'search' operation is deprecated and will be removed in a future release. Use 'scroll-search' instead.`

We get the above message when running geonames,  we should fix the track to use the appropriate operation

amarkdotdev:
## Summary
- Update the geonames `scroll` operation to use `operation-type: scroll-search` instead of deprecated `search` with pages.

Fixes #624

## Test plan
- [ ] Confirm `geonames/operations/default.json` scroll op uses `scroll-search`
- [ ] Run geonames track (or dry-check) and verify the deprecation warning is gone

### DIFF
--- a/geonames/operations/default.json
+++ b/geonames/operations/default.json
@@ -90,7 +90,7 @@
     },
     {
       "name": "scroll",
-      "operation-type": "search",
+      "operation-type": "scroll-search",
       "pages": 25,
       "results-per-page": 1000,
       "body": {


amarkdotdev:
Quick bump — still green on my side and happy to adjust if anything's needed. No rush, just checking if this is still on the radar.

PR #1253, opened 2026-09-09
feat(logs): make ingest_percentage effective in elastic/logs

salvatore-campagna:
## TL;DR

`elastic/logs` documents an `ingest_percentage` track parameter, but it has no runtime effect: setting it changes nothing about the benchmark. This implements the parameter in `ProcessedCorpusParamSource` so the documented behaviour becomes real.

## Summary

The parameter was inert for two independent reasons. `track.json` set a Jinja variable (`p_ingest_percentage`) that nothing referenced, and `logging-indexing-querying.json` emitted `ingest-percentage` at task level, where Rally passes it to schedulers rather than to the parameter source. Even at operation level it would have been ignored, because the indexing operations use the custom `processed-source` parameter source, which bypasses Rally's built-in `BulkIndexParamSource` (the component that implements `ingest-percentage` everywhere else). Because the dead Jinja variable referenced the parameter, Rally's unused-track-param check counted it as used, so a user setting it did not even get a warning.

The implementation scales `total_docs_per_day` (derived from `raw_data_volume_per_day`) before client partitioning and before the timestamp generator is built. Scaling the daily document count rather than truncating the document stream keeps generated timestamps covering the entire `start_date`..`end_date` range, so the same period is indexed with proportionally fewer documents and query workflows keep finding data in every window; it is equivalent to reducing `raw_data_volume_per_day` by the same factor. Validation matches Rally's built-in semantics: a number in `(0, 100]`, default 100, so existing invocations are unaffected. The operation-level parameter is wired into the `logging-indexing` and `logging-indexing-querying` challenges; the stale task-level occurrence and the dead Jinja variable are removed.

> **IMPORTANT**
> Anyone who previously set `ingest_percentage` on this track was silently getting a full-volume run. After this change the parameter takes effect, so such runs will ingest less data than before. Values other than 100 also shorten the throttled indexing phase proportionally, and in `logging-indexing-querying` the `completed-by: bulk-index` parallel block ends earlier, shortening the query phase with it.

## Testing

```
make test
make lint
```

### DIFF
--- a/elastic/logs/README.md
+++ b/elastic/logs/README.md
@@ -257,7 +257,7 @@ The following parameters are available:
 * `throttle_indexing` (default: `false`) - Whether indexing should be throttled to the rate determined by `raw_data_volume_per_day`, assuming a uniform distribution of data, or whether indexing should go as fast as possible. 
 * `disable_pipelines` (default: `false`) - Prevent installing ingest node pipelines. This parameter is experimental and is to be used with indexing-only challenges.
 * `initial_indices_count` (default: 0) - Number of initial indices to create, each containing `100` auditbeat style documents. Parameter is applicable in [many-shards-quantitative challenge](#many-shards-quantitative-many-shards-quantitative) and in [many-shards-snapshots challenge](#many-shards-snapshots-many-shards-snapshots).
-* `ingest_percentage` (default: 100) - The percentage of data to be ingested.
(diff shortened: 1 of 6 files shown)


jordan-powers:
The change makes sense to me. But I think we should evaluate which benchmarks are currently running with this parameter configured but non-functional. We'll want to either set that parameter back to 100% on those benchmarks, or add an annotation when this PR is merged.

pquentin (approved):
Thanks! LGTM

Tested with

```
esrally race \
  --track-path=../rally-tracks/elastic/logs \
  --challenge=logging-indexing \
  --track-params="ingest_percentage:50,raw_data_volume_per_day:0.001GB,number_of_replicas:0"
```

PR #1255, opened 2026-09-18
Update geopoint and geoshape tracks to reflect improved performance from ES PR159459

craigtaverner:
In Elasticsearch we have a [performance improvement PR](https://github.com/elastic/elasticsearch/pull/159459) that improves a number of geo-grid queries on both geo_point and geo_shape tracks. Here we update the rally tracks for the nightly benchmarks to reflect these improvements, adding tasks for queries that were previously too slow to include in benchmarks:
* Sorting queries in the geo_point track
* geogrid tasks in the geo_shape track

If we merge this PR before the elasticsearch one, all tasks should run, but some will take a lot longer. On my M4 Mac running the full suites on main took 2x - 3x longer than running on the PR branch. So ideally we merge this after merging the ES PR. See that PR description for some of the benchmark results from local runs.


geopoint/operations/default.json
```diff
@@ -762,6 +762,186 @@
...
+            }
+          }
+        ],
+        "fields": ["grid", "location"]
+      }
+    },
+    {
+      "name": "geoGrid_sort_geohash-esql",
+      "operation-type": "esql",
+      "query": "FROM osmgeopoints | EVAL grid = ST_GEOHASH(location, 5)::long | SORT grid ASC | KEEP grid, location | LIMIT 10"
```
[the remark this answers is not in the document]
craigtaverner: ES|QL does not allow sorting on grid ids, but query DSL does, so this cast is required. It is also extremely cheap, since the underlying type is already a long, so this cast really only changes the column type, not the actual values.

geoshape/challenges/default.json
```diff
@@ -445,35 +445,71 @@
...
         {
           "operation": "geoGrid_aggs_geohex",
           "warmup-iterations": 100,
           "iterations": 50,
           "tags": ["geo_grid"]
+        },
+        {
+          "operation": "geoGrid_aggs_geohex-esql",
+          "warmup-iterations": 10,
+          "iterations": 10,
```
ncordon, line 511: This one has 10 iterations and warmup iterations vs 50 for the rest of the operations. Why?

geopoint/operations/default.json
```diff
@@ -762,6 +762,186 @@
...
+      "operation-type": "search",
+      "body": {
+        "size": 10,
+        "_source": false,
+        "runtime_mappings": {
+          "grid": {
+            "type": "keyword",
+            "script": {
+              "lang": "painless",
+              "source": "double lon = doc['location'].lon; double lat = doc['location'].lat; int res = 5; double size = 180.0 / Math.pow(Math.sqrt(7), res); double q = (2.0 / 3.0 * lon) / size; double r = (-1.0 / 3.0 * lon + Math.sqrt(3.0) / 3.0 * lat) / size; double x = q; double zc = r; double y = -x - zc; int rx = (int) Math.round(x); int ry = (int) Math.round(y); int rz = (int) Math.round(zc); double xd = Math.abs(rx - x); double yd = Math.abs(ry - y); double zd = Math.abs(rz - zc); if (xd > yd && xd > zd) { rx = -ry - rz; } else if (yd > zd) { ry = -rx - rz; } else { rz = -rx - ry; } emit(rx + '/' + rz);"
```
ncordon, line 926: What does this painless script do? It's gibberish to me

ncordon (approved):

